PR #1, opened 2010-09-01
processTV crashes when parsing filenames that contain roman numerals

f1sherman:
Looks like you added the miniseriesnaming stuff, thanks!  I noticed that processTV is crashing when processing the filenames 'The.Pillars.of.the.Earth.Pt.I.PROPER.720p.HDTV.X264-DIMENSION.mkv', 'The.Pillars.of.the.Earth.Pt.VII.and.Pt.VIII.720p.HDTV.X264-DIMENSION.mkv' and 'The.Pillars.of.the.Earth.Pt.VII.&.Pt.VIII.720p.HDTV.X264-DIMENSION.mkv'.  Looks like it's choking on a TypeError raised by roman.py.  Apparently 'VII' isn't of type string???  In any case, I added some logging to roman.py and modified processTV.py to gracefully handle any exceptions thrown from tvnamers parser.

I'm not sure if parsing roman numerals is a reality at this point (at least not without some changes on the tvnamer side).  Here is the error message that is generated after my changes:
Unable to parse the filename /data/sabnzbd_downloads/complete/TV/The.Pillars.of.the.Earth.Pt.VII.and.Pt.VIII.720p.HDTV.X264-DIMENSION.1/The.Pillars.of.the.Earth.Pt.VII.and.Pt.VIII.720p.HDTV.X264-DIMENSION.mkv into a valid episode, error message: expected string, got for input VII

### DIFF
--- a/lib/tvnamer/roman.py
+++ b/lib/tvnamer/roman.py
@@ -44,7 +44,7 @@ def int_to_roman(input):
    MCMXCIX
    """
    if type(input) != type(1):
-      raise TypeError, "expected integer, got %s" % type(input)
+      raise TypeError, "expected integer, got {0} for input {1}".format(type(input), input)
    if not 0 < input < 4000:
       raise ValueError, "Argument must be between 1 and 3999"   
    ints = (1000, 900,  500, 400, 100,  90, 50,  40, 10,  9,   5,  4,   1)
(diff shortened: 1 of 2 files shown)


midgetspy:
I just fixed the TypeError instead of catching it :0)

midgetspy:
Although I just noticed - those double-episodes are not going to work anyway. Single episode roman numerals will work fine but I didn't make any double-episode regexes for them.

f1sherman:
Cool, that is obviously better.  I did notice that double-episodes are just treated as one episode.

Thanks!

midgetspy:
BTW I added a double-episode regex for the miniseries stuff last night as well, so those files should work properly - let me know if that's not the case.

PR #2, opened 2010-10-05
Updates to include timezone offset in coming episodes and related configuration.

chaosloth:
Please note that that the this version of SB now depends on pytz

Refer http://www.sickbeard.com/forums/viewtopic.php?f=3&t=127

For screen shots, bugfixes 

Cheers,
C

### DIFF
--- a/data/interfaces/default/comingEpisodes.tmpl
+++ b/data/interfaces/default/comingEpisodes.tmpl
@@ -1,6 +1,8 @@
 #import sickbeard
 #import datetime
 #from sickbeard.common import *
+#import pytz
+#from pytz import timezone
 #set $oneWeek = datetime.timedelta(days=7)
 #set global $title="Coming Episodes"
 
@@ -57,8 +59,27 @@
           </p>
           <p class="noMargin">
             <span class="title">Status:</span> <span class="info ">$curEp.show.status</span><br />
-            <span class="title">Airs:</span> <span class="info ">$curEp.show.airs on $curEp.show.network</span><br />
-            <span class="title">Format:</span> <span class="info ">
+            <span class="title">Airs:</span> <span class="info ">$curEp.show.airs $sickbeard.TZ_EPS on $curEp.show.network</span><br />
+            
+            #set t1 = $curEp.airdate.__str__() + " " + $curEp.show.airs
+            #if t1.count(":") > 0:
+            #set fmt1 = "%Y-%m-%d %A %I:%M %p"
+            #else
+            #set fmt1 = "%Y-%m-%d %A %I %p"
+            #end if
+            
+            #set raw_dt = datetime.datetime(*time.strptime($t1, fmt1)[0:5])
+            #set fmt2 = "%A %I:%M %p %d/%m/%Y"            
+            #set ep_tz = timezone(sickbeard.TZ_EPS)
+            #set ep_dt1 = ep_tz.localize(raw_dt)
+            #set usr_tz = timezone($sickbeard.TZ_USER)
+            #set usr_dt1 = usr_tz.normalize(ep_dt1.astimezone(usr_tz))
+            #set usr_dt2 = usr_dt1 + datetime.timedelta(hours=+2)
+            
+            <span class="title">$usr_tz</span> <span class="info ">$usr_dt1.strftime(fmt2)</span><br />
+            <span class="title">Approx download</span> <span class="info ">$usr_dt2.strftime(fmt2)</span><br />
+            
+            <span class="title">Format:</span> <span class="info ">       
 #if $curEp.show.quality in $qualityPresets:
 $qualityPresetStrings[$curEp.show.quality]
 #else:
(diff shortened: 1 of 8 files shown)


midgetspy:
OK I will get around to this soon, promise :0)

ghost:
i would love time-zone

thezoggy:
timezones is on hold until the new tvdb site is launched. or until tvrage api supports it.

ghost:
in other words: never

:P

midgetspy:
Actually SB will have timezone support sometime, it's available on tvrage... probably before the new TVDB site ;-P

ghost:
Nice, because the new TVDB site is never gonna happen, lol

thezoggy:
this pull request is pretty much dead in the water. needs to be redone..

bjeanes:
+1

dunpealhunter:
To bring new life into this pull request: can the website trakt.tv be used for this? On their website they managed to make it work, for your personal calendar all the air times are set to your own time zone. They have a usable API for the calendar, can that somehow be used to work for this pull request?

cagnulein:
+1

kierr:
+1

simonk83:
+1

EDIT:  As Zoggy so succinctly put it, as only he can, "this isn't going to happen", so how about you close this pull request if you have no intention of actioning it,  to at least stop other people getting their hopes up?

It appears to be the 4th most requested feature though.   Priorities can be a little strange on this project, as much as I love it.

thezoggy:
there is a lot more to doing this properly than some arbitrary offset.you have to factor in changing the actual date as well at times. on top of this.. the source data from tvdb sometimes is not even there or just all sorts of incorrect. that field is just a text field with nothing being validated or sanitized.. its been talked about time and time again on googlecode and the forums. there was a pull request 2 years ago for this but it never went anywhere. it was actually part of sb (the config > timezone page) but the implementation was less than desirable. 'we' have been hoping that the new tvdb site (which has been in dev forever now..) to take care of this issue. if the data was in UTC/GMT then everything would really simple. also.. tvdb uses a timeslot for the show.. not the actual ep. which is what matters honestly. because when a show changes timeslots.. or airs an ep out of their normal time.. sb has no clue.. anyways regardless of all of this. this pull request is so old.. so many merge conflicts would happen. along with the fact that there are multiple features in this one pull request.. which just makes things even more of a pita to pull and test. its just honestly not worth the time even mess with it.

simonk83:
Fair enough.

I suppose what I don't understand is that right now, SB either gets, or has a guess at the day (I don't really care about the time too much) that the episode is airing, and for the most part it's correct.    From my point of view (which I understand is coming from somewhere that doesn't have all the facts), I would have thought it possible to just change the "display" date of the show, but not touch the "actual" date of the show that sickbeard is going off.

Can we not have two variables (this probably isn't how it's done, but I'm sure you'll get what I mean), so we have for example:

airdate - which is the date that sickbeard is assuming the episode is airing on (however it gets that info), and this is the way that it currently works (you know what I mean)

and then

offsetdate - which is simply airdate +10 (or whatever)

So for all intents and purposes, nothing changes for sickbeard in the inner workings as it always uses "airdate" for it's calculations.  It's just that it now uses "offsetdate" in a superficial way only, to display to the user.

Does that make sense?

So when doing calculations, use airdate.   When pushing that calc to the GUI, use offsetdate.

It's hacky, but it achieves it's purpose in that for people outside of the US (or in the wrong timezone in the US), the coming episodes show on the actual date rather than a day ahead (which is what happens here in Australia).

Right now, it's the 28th of March.  In the US, it's still the 27th, so all the coming episodes go into "missing" a day early.

Aaaanyway, I hope all that makes sense.   We don't really need sickbeard to change the way it works, we just need it to display something different to the actual dates it's using.

pdf:
So, tell us what we need to do to make this happen - for all of us that are affected by this, a best-effort approach would be more than acceptable: data doesn't exist on tvdb, fine, that show gets scheduled wrong, rather than _every_ show being scheduled wrong.  Waiting for thetvdb's new code is obviously not a viable option.

A perfect solution is not necessary here - an incremental improvement would be greatly appreciated.  If devs can agree that best-effort code will be accepted, I'll spend some time to make this happen.  There's obviously no point wasting time on code that will not be merged.

ghost:
Do not see this as an attempt to bump an old thread, but since I'm from Norway, I would personally prefer having some sort of time zone corrections applied :) Would benefit quite a bit, to see when they air, and not get the "missed" several hours before the episode even ran.

Good luck!

alexgrist:
Agree, this deserves a bump actually. It'd be great to have time zone offsets to reflect the Air Time in the current time zone.

thezoggy:
while yes i do agree it would be nice to have timezone offset in sb. we really cant provide this until the data being served up from tvdb is reliable and sanitized. and honestly.. we need ep info.. not show info..

chaosloth:
I guess after 10 years this should really be closed aye?!

PR #4, opened 2010-10-08
Added show exceptions

bekriebel:
I have added "Grey's Anatomy" and "$#*! My Dad Says" to the common.py exceptions. They both are released under different names ("Greys Anatomy" and "Shit My Dad Says").

Thanks for the great work midgetspy!

-Brint

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -131,6 +131,8 @@ def searchForNeededEpisodes():
 			logger.log(traceback.format_exc(), logger.DEBUG)
 			continue
 
+		curFoundResults = filter(lambda x: sceneHelpers.filterBadReleases(x.name) and isGoodResult(x, episode.show), curFoundResults)
+
 		didSearch = True
 		
 		# pick a single result for each episode, respecting existing results


midgetspy:
Apostrophes are automatically filtered out, no exception is necessary for Grey's Anatomy. I'll add the other one, thx.

bekriebel:
Hrmm... Grey's didn't scrape for me automatically last night. I'll do some testing; maybe it was just a problem with the feeds.

bekriebel:
Ahh... I think I see now. The name sanitizer and exceptions list only works for searches, it doesn't seem to work on the RSS feeds. When you are pulling the episodes into cache, it isn't matching against sanitized names, so it never gives the cached version a TVDBID. Since there is no TVDBID, it never matches the episode when it checks the airing episodes against the cache.

I may try to hack at this tonight and try to come up with a way to handle sanitized and exception based names in the RSS scans.

Log:
Oct-08 16:18:44 DEBUG    SEARCH :: Adding item from RSS to cache: Greys.Anatomy.S07E03.HDTV.XviD-2HD
Oct-08 16:18:44 DEBUG    SEARCH :: sickbeard.db: SELECT \* FROM tv_shows WHERE show_name LIKE ? OR tvr_name LIKE ? with args ['Greys_Anatomy%', 'Greys_Anatomy%']
Oct-08 16:18:44 DEBUG    SEARCH :: Unable to match a record in the DB for Greys_Anatomy%
Oct-08 16:18:44 DEBUG    SEARCH :: sickbeard.db: SELECT \* FROM tv_shows WHERE show_name LIKE ? OR tvr_name LIKE ? with args ['Greys Anatomy%', 'Greys Anatomy%']
Oct-08 16:18:44 DEBUG    SEARCH :: Unable to match a record in the DB for Greys Anatomy%
Oct-08 16:18:44 DEBUG    SEARCH :: cache.db: INSERT INTO nzb_su (name, season, episodes, tvrid, tvdbid, url, time, quality) VALUES (?,?,?,?,?,?,?,?) with args ['Greys.Anatomy.S07E03.HDTV.XviD-2HD', 7, '|3|', 0, 0, 'http://SOME_RSS_SITE/rss/REMOVED_FOR_COMMENT', 1286576324, 1]

midgetspy:
My Grey's downloaded off RSS just fine last night and has been for many seasons so I don't think that is the problem. But looking at your log snippet what you're saying does appear to be true... I will check my logs and see what's going on.

midgetspy:
So it looks like mine actually got snatched on the recent backlog search rather than the RSS. I'm pretty sure the fix for this is to reverse the lookup for shows that fail.  For this example, we would:

1) Take the scene name Greys.Anatomy and look it up in the DB (which will fail)
2) Since it failed, for each DB entry generate all scene names and compare them to the scene release name. These "scene names" are what I use to search backlog stuff and they are pretty solid. I probably won't have time to do this until next week.

bekriebel:
I may have some time this weekend; if so, I may try to implement a fix that you are more than welcome to pull (if you agree with the solution, of course).

I was thinking of a few options for this. One is how you mentioned  (starting as is, then failing over to the reversed lookup). Another option is to just tie it into the normal backlog searching and completely forgo the initial lookup that is failing. And finally, I was considering adding a column to the shows table that includes a sanitized copy of the name and adding that column to the search.

Do you have any particular preference? Would there be any other benefit to having the additional sanitized column so that the release could be properly matched to the proper tvdbid?

-Brint

midgetspy:
We definitely want to do everything possible on the client side to get a match without making an additional request to the server (aka use RSS instead of backlog if at all possible). SB puts a very heavy load on the search providers already.

The only problem with storing a sanitized scene name in the database is that some shows have more than one possible scene name because of country and year suffixes (eg "Some Show (2003)" could be just posted as "Some.Show". see sceneHelpers.allPossibleShowNames() for more info). Because of this we would have to store multiple scene names in the db and pull them all out of the database and iterate through them anyway in which case there's basically no difference from just generating them ourselves each time. One other possible option is if there is a way to ignore punctuation in the DB value when doing the query - MYSQL allows some limited regex support but I kinda doubt SQLite does, but it's worth looking into at least.

Implementing this would be fairly easy. To limit performance impact I'd do it like this:

After populating a whole cache using only db lookups, call a new method (cache.attemptShowSynch() or something) that looks something like:

```
for each show in the show list:
    for each scene name in the cache with no tvdb_id:
        if search.isGoodResult(curShow, curSceneName):
            update cache with curShow.tvdb_id
```

I think that's the simplest way at least, since we can leverage the regex matching in search.isGoodResult().

Nic

PR #5, opened 2010-10-10
Untitled

adam-bielicki:
I suspect this was intended for addition. I was having problems with germen releases being downloaded on accident.


midgetspy:
I haven't had a chance to look into it but this code doesn't seem to work. This is a valid bug that needs to get fixed though.

adam-bielicki:
odd. I've been running that change for a while and haven't been getting the same reported problems. whether or not it is working is another story. :p

midgetspy:
Err, yeah I meant to post this on a different request.

PR #6, opened 2010-10-10
Untitled

adam-bielicki:
it seems all (I couldn't find an exception amongst my shows) episodes appearing with airdate 'never' are airing after episodes with a defined airdate in the future. They should be excluded from the progress bar on home as they are unavailable.


adam-bielicki:
see the current commits. This should work properly.

adam-bielicki:
any word on these changes?

midgetspy:
I didn't notice you updated this... I'll pull this tonight.

midgetspy:
Should be in now.

PR #7, opened 2010-10-11
SB changes ignored eps to skipped during refresh

Smenus:
When an ep has an airdate of 'never', which, with quite a few shows, happens a lot, and the episode is set to ignore, when the show gets refreshed (`loadFromTVDB` is called on the episode), the episode is changed to skipped with no regard to what the previous status was.
This was my way around the fact that episodes with airdates of never are being counted in the downloaded episodes bar, so I ignored them all. Obviously, this changed them back to skipped every refresh, which made my method pointless. I think I'll just merge one of the other methods, which seem cleaner and easier.


Smenus:
From the look of it, this was included in one of the last commits. Closing...

PR #8, opened 2010-10-12
Revert e62e19095

bekriebel:
This causes all RSS feed results to error with "Unable to parse the filename <episode name> into a valid episode".

It looks like the curFoundResults.name value being passed is not the proper input for filterBadRelease.


midgetspy:
Guess I should have tested this before pulling, heh.

bekriebel:
No worries. It threw me for a bit since I was testing the new RSS code for scraping with other possible names. I thought for a while that I had blown it up until I reverted that one.

PR #11, opened 2010-10-14
Twitter Integration is ready to be pulled

adamrb:
Twitter notifications have been tested and have worked without issues for the last month or so using OAUTH.

I recommend that Midgetspy creates a twitter "app" at http://dev.twitter.com/apps

You will then need to change the following strings in "sickbeard/notifiers/tweet.py", so that notifications appear to come from SickBeard in the user's stream.

consumer_key = "sBD0atiSA1TY5tv4ru3Cyw"
consumer_secret = "eWeyr7uyVJZpr2AnBW4s4NKRm2WjyHDV2omyq88CeFc"


midgetspy:
OK so I have no idea how twitter works, but is this going to send messages to my account as though I posted them myself? Or is it some kind of private message sent only to me? (I haven't tried it yet, just curious)

adamrb:
Correct.  It's going to post a message as though you've posted it yourself.

Most people would setup a second account they can follow from their first account so their normal twitter stream isn't filled with downloading posts.

Here's an example of my SickBeard twitter account: http://twitter.com/#!/abdvr

PR #15, opened 2010-10-27
unnecessary whitespace

michaelficarra:
The codebase was absolutely littered with trailing whitespace and dos-format line endings. Cleaned all of this up for a decent reduction in size and to suppress the god-awful syntax highlighting of trailing whitespace.


midgetspy:
The reason I've never fixed this is cause changing line endings means complete loss of history and tracking between branches... but I really do need to do it sometime. I will try to get the existing branches merged sometime soon so I can change it more easily...

midgetspy:
Changing line endings now would be a nightmare to existing branches/forks so I'm just going to leave them for now - the python interpreter doesn't care and it shouldn't make a difference to any decent editor. Since your change was a little old by the time I got to it I ended up just killing the whitespace myself, thanks. The time thing I will either do separately or wait for a new commit from you since as per my comment part of your change is unnecessary (we only need to change the renaming stuff, not the search text). Thanks for the submissions.

PR #17, opened 2010-10-27
Exception zldme

bekriebel:
Can you please add the following commit to include an exclusion for Zane Lamprey's Drinking Made Easy?

Thanks!
-Brint


midgetspy:
Pulled.

PR #18, opened 2010-10-27
Added exception for Agatha Christie's Poirot

tolecnal:
Cheers mate!


midgetspy:
I did this manually because it conflicted with another pull request. Also I changed the second exception to "Agatha Cristie's Poirot" because all eps except one are named that way - no point making an exception for one mistake.

PR #21, opened 2010-11-05
Issue 570: "Select all" in displayShow

ghost:
Added links (Select Filtered Episodes; Clear All) to the displayShow page, just beneath the existing filters.  If you just wanted to select _all_ 'Wanted' episodes, you could filter based on that, then hit 'Select Filtered Episodes' and voila.  If you do, it'll also check the 'seasonCheck' checkboxes so it's easy to select a series, then unselect a couple of seasons.


ghost:
No, they're just unstyled links-- I gave the span tags classes to make it easier to style in the future.

midgetspy:
Pulled finally, thanks.

PR #22, opened 2010-11-08
Untitled

mannkind:
I don't know if anyone else has complained about this, but I find it annoying that my episode download count drops when propers/higher quality episodes have been snatched.


midgetspy:
Pulled, thanks.

PR #24, opened 2010-11-13
Add IPv6 Listening Support for Web Interface

jerub:
There are two commits here, one is to fix the line endings in sickbeard/**init**.py which I needed to do in order to make changes to that file, the second commit adds a checkbox under 'General' to enable listening on an ipv6 address ('::') instead of an ipv4 address ('0.0.0.0').

This may not work on ipv4 only hosts, so it's optional and off by default. This will work with all ipv6 dual stack setups, and hosts with ipv4 configured, and ipv6 unconfigured.


midgetspy:
Merged and closed.

PR #25, opened 2010-11-15
Made SickBeard work with Python 2.5

fernandotakai:
SickBeard was using with keyword without using  from **future** import with_statement. I added them so it could work with 2.5 out-of-the-box.


midgetspy:
Whoops, forgot about that. Thanks, pulled.

PR #26, opened 2010-11-16
Added babylon5 to sceneExceptions

jerub:
This configuration should probably be easier to update, a metadata file that can be retrieved from somewhere central-ish, so it doesn't have to require a sickbeard upgrade every time a new show is being stupid. Talk to me tomorrow about it on IRC if you want.


jerub:
I don't fully understand git. This seems to be pushing commits at you that you've already accepted. If there's some commands I can run that will stop that, let me know too :)

midgetspy:
This exception should still have the normal case "Babylon 5" - it overrides the show name rather than adds to it. I'll catch you on IRC.

midgetspy:
I just committed the change manually, closing this.

PR #28, opened 2010-11-23
Launch browser

generica:
launchBrowser is spawned on default port, which doesn't work if you used --port on the command line.  If forced port, use that instead.


midgetspy:
Pulled.

PR #30, opened 2010-11-23
PS3 metadata provider

swc:
Hello Nic,

I added a provider for PS3 to allow the episode thumbnails to show up properly. It also generates season, fan art, and series thumbnails if the user requests them in the config... although the PS3 will ignore those. This metadata provider is working very well for me, and I expect it would be appreciated by anyone using the PS3 as their media server.

Currently, I'm not generating NFO files because the PS3 doesn't recognize them. Should I produce XBMC NFOs as a default if the user wants them?

One last thing... On the PS3, 'folder.jpg' is a special file name. The PS3 will use this file as the thumbnail for any video files in the same directory as 'folder.jpg'. Unfortunately, this will overrides any episode thumbnails that may also be present. The good news is that sorting a show into season folders avoids this issue.

I believe the best solution for this would be to have the name of the series thumbnail specified by the metadata provider instead of having it hardcoded to 'folder.jpg'... The web interface could load the images according to whatever name is specified in the metadata provider. I'd be happy to make this change if you're in agreement.

Sick Beard is amazing! Makes my life a lot simpler, so thanks!

### DIFF
--- a/sickbeard/metadata/__init__.py
+++ b/sickbeard/metadata/__init__.py
@@ -1,7 +1,7 @@
-__all__ = ['generic', 'helpers', 'xbmc', 'mediabrowser']
+__all__ = ['generic', 'helpers', 'xbmc', 'mediabrowser', 'ps3']
 
 import sys
-import xbmc, mediabrowser
+import xbmc, mediabrowser, ps3
 
 def available_generators():
     return filter(lambda x: x not in ('generic', 'helpers'), __all__)
--- a/sickbeard/metadata/ps3.py
+++ b/sickbeard/metadata/ps3.py
@@ -1,0 +1,82 @@
+# Author: Nic Wolfe <nic@wolfeden.ca>
+# URL: http://code.google.com/p/sickbeard/
+#
+# This file is part of Sick Beard.
+#
+# Sick Beard is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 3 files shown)


midgetspy:
IMO you shouldn't generate anything here that isn't used by the PS3... I will eventually make more than one metadata provider enabled at once if people want to combine them.

You can already change the name of the show poster, just set self.poster_name in your constructor.

Thanks for contributing!

swc:
Ok, I have made the above changes.

midgetspy:
OK, pulled.

midgetspy:
I was trying to figure out a ticket (http://code.google.com/p/sickbeard/issues/detail?id=1056) and ended up reviewing the PS3 metadata and I think there are unnecessary things in here, I just wanted to confirm.

From your comments it looks like the only file that is needed for the PS3 provider is episode thumbnails as filename.avi.cover.jpg, and IF sorted into season folders then a show poster could be added as folder.jpg in the show root folder. Is this correct?

I'm not sure why I have cover.jpg in the show root but cover.jpg & fanart.jpg are both not supported by the PS3, right?

swc:
> From your comments it looks like the only file that is needed for the PS3 provider is episode thumbnails as filename.avi.cover.jpg, and IF sorted into season folders then a show poster could be added as folder.jpg in the show root folder. Is this correct?

Absolutely correct... folder.jpg will override any filename.avi.cover.jpg siblings.

fanart.jpg is not used by the PS3, so it can be safely removed. If I recall correctly, I left cover.jpg so that the sickbeard web interface would have thumbnails to display for upcoming shows. Is this not necessary?

PR #33, opened 2010-11-24
XMPP/Jabber Notifications

wdudokvanheel:
# Motivation

Currently I, as a Sick Beard user, either need to create a twitter account or get Growl working to get updates on new episodes found/downloaded by Sick Beard. The addition of XMPP support will enable users to get instant notifications in their IM application. Users of Google's Android will be able to receive messages to their phone by using the standard gtalk application.
# Implementation

Implementation of the XMPP notifier is done using the already existing notfication system and is implemented roughly the same as the Growl notifier.
# Dependencies

The communication with the XMPP server is done with the xmpppy, an XMPP python library. The xmppy library has been patched (to prevent some deprecation warnings) and added to _lib_ directory.
# Usage

The use of this notifier is quite simple. The use case scenario;
- The user goes to Config -> Notifications
- The user check the _Enable Jabber Notifications_ box
- The user enters his account details, server and the recipient of the message. (Server might prevent users from messaging themselves) 
- The user can choose to test the given details by pressing the _Test Jabber_ button.
- The user either receives an error message or successful message on the page
- The user receives a test notification in their IM client
- The user can save the entered values by pressing the _Save Changes_ button
- Next time a download is snachted or finished, it will notify the user 
# Resources

[Xmppy](http://xmpppy.sourceforge.net/) - The Python XMPP library

[My blog](http://blog.bitechular.com/2010/11/24/sick-beard-xmpp-intergration/) - For a screenshot

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -49,6 +49,15 @@
 
   });
 
+  \$('#testJabber').click(function(){
+    var jabber_username = \$("#jabber_username").val();
+    var jabber_password = \$("#jabber_password").val();
+    var jabber_server = \$("#jabber_server").val();
+    var jabber_port = \$("#jabber_port").val();
+    var jabber_recipient = \$("#jabber_recipient").val();
+
+    \$.get("$sbRoot/home/testJabber", {'username': jabber_username, 'password': jabber_password, 'server': jabber_server, 'port': jabber_port, 'recipient': jabber_recipient}, function (data){alert(data)});
+  });
 });  
       
 </script>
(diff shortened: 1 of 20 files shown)


wdudokvanheel:
Strange, the diffs show that I removed large parts of (unchanged) code and added them back in? Probably something to do with my IDE (maybe changed the tabs to spaces or something?) Anyway, I shall fix it!

midgetspy:
Hm, it should be all spaces and windows \r\n line endings. If it messes with the diffs to much just make a new branch and pull your changes into it so you have a clean diff set for me to pull. I will review this feature this weekend hopefully, thanks for the effort!

wdudokvanheel:
Could be the newlines, as I use Linux and very crappy (imho) IDE. I shall try to fix it today, otherwise it might be next week. 
As I didn't mention testing; I've been running this on my fileserver for the past couple of days and it works like a charm.

midgetspy:
I still haven't had a chance to check this out, but I have two questions:
1) Is there a reason we can't just pick a public server and hard-code it? The less configuration required the better IMO
2) If we pick a server that allows you to send messages to yourself couldn't we just have a single email field and use it for both? Having two accounts so that one can send messages to the other seems like a terrible hack IMO.

I did look into making a SB XMPP bot so that SB could send a message to the bot from a user and the bot would just send it back, which would allow simple configuration with a single account. I never got it to work though on the google apps framework.

mtehonica:
I would love to see something like this implemented!

travisghansen:
I have a bit of experience with xmpppy as well.  To answer your question midgetspy you really need the configuration options he has.  In order to send a message both a sender and recipient are necessary along with the other values asked for.  I can provide some assistance here if desired.

depassp:
I have added XMPP functionality to a more recent master:

https://github.com/midgetspy/Sick-Beard/pull/212

PR #35, opened 2010-12-04
Link to each ep in displayShow.tmpl

DaveWhite:
Added a hyperlink to each episode in the show list for launching that episode straight from the browser.
Changes in lines 124 and 126.


midgetspy:
Thanks, but this won't get pulled for a few reasons. First, it only works if you're browsing from the same PC running Sick Beard - this is not the case for many SB users. Second, this is what HTPC software is for, SB intentionally doesn't attempt to do that stuff.

Also FYI in the future you've changed the line endings in that file so every line is changed so I can't tell what's actually changed - all line endings in SB source should be \r\n. You can probably fix this by just changing you autocrlf setting in git.

PR #37, opened 2010-12-05
Changes to fix issue 737

LethalCorpse:
This should take care of it until you can rewrite the PP code.


midgetspy:
Hrm your line endings are messed up, I can't tell what the changes are... try reverting and committing again with windows line endings? (Git can take care of it for you with autocrlf)

PR #44, opened 2010-12-14
Simple language support for TVDB and nzbmatrix

ghost:
I have added language support by using three parameters in the config.ini:
language_long (can be empty if you want english=
language_short (default "en")
result_filters (this was used in sceneHelpers.py to filter bad releases. now it is configurable)

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -166,10 +166,6 @@ def main():
 	logger.log(u"Loading initial show list")
 	loadShowsFromDB()
 
-	# set up the lists
-	sickbeard.updateAiringList()
-	sickbeard.updateComingList()
-
 	# fire up all our threads
 	sickbeard.start()
 
--- a/data/interfaces/default/comingEpisodes.tmpl
+++ b/data/interfaces/default/comingEpisodes.tmpl
@@ -1,7 +1,6 @@
 #import sickbeard
 #import datetime
 #from sickbeard.common import *
-#set $oneWeek = datetime.timedelta(days=7)
 #set global $title="Coming Episodes"
 
 #set global $sbPath=".."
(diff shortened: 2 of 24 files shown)


SkillPhil:
Great news! That's what i have been looking for. Have you considered adding
language support for Newzbin as well?

ghost:
As I do not have a newzbin account this is "experimental". The language can be set using an URL parameter. I have added a new config-property called LANGUAGE_MED (as medium :-)). Set it to "Ger" and you get only german results.

SkillPhil:
Amazing! Can't wait to check this out. Thanks a lot!

ghost:
Sorry, but you have to synch again (if already done). I have fixed a bug (see above). If you find any error please report it here. I am not able to download NZBs from newzbin because I do not have a premium account. I tested my code until SickBeard needs to download the NZB, which fails because of not having premium status.

SkillPhil:
Just tried it out. I couldn't find the LANGUAGE_MED parameter in the config.ini. I added it manually. Sickbeard found a file but ignored the LANGUAGE_MED = German. As i can see in the Debug Log the Newzbin search URL has still "Attr%3ALang%7EEng+AND+NOT+Attr%3AVideoF" in it.

ghost:
Hi!
The parameter is only added if there is no config.ini at all. After you have added the paramter "LANGUAGE_MED = Ger" you have to stop SickBeard and restart it. SickBeard is reading the config only at startup.

SkillPhil:
Hm, strange behavior here. i had it up and working several times but after a restart sickbeard seems to reset the config file. The LANGUAGE_MED parameter is gone again. But if it's up it doesn't seem to work the way it's supposed to. Not sure if this is a general issue with newzbin but not all files are found. E.g. a file with german AND english language (attr:~Ger) should match. Although the file does exisit as i can find it on the page, sickbeard only picks up the german only version (attr:=Ger, if one exists). In some cases it even doesn't pick up that one, as the Newzbin search URL (RSS) doesn't bring up results for attr:~Ger by mistake.

ghost:
Interresting. As I said, I do not owe a newzbin account. Maybe you could get in contact with me by email: ssppaamm@gmx.de So you could help me testing. Take care!

PR #45, opened 2010-12-16
Added option to use season thumbnails for Coming Episodes

bardelot:
For those that prefer season thumbnails instead of show thumbnail. Also very useful if TV show banners are used instead of covers.


bardelot:
Withdrawn in favor of the changes made by thezoggy in zog_config-coming_ep.

PR #46, opened 2010-12-17
perform episode search via ajax

aforty:
Hi there,
Love Sick-Beard, wanted to help. Hopefully I did just that.

Changes episode search from show page to use an ajax call. Updated webserve.py-searchEpisode but it still works the same way if needed (wasn't sure if you referenced it from some other place) and cleaned it up a bit while I was there.

Hope this is helpful.

Thanks!

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -166,10 +166,6 @@ def main():
 	logger.log(u"Loading initial show list")
 	loadShowsFromDB()
 
-	# set up the lists
-	sickbeard.updateAiringList()
-	sickbeard.updateComingList()
-
 	# fire up all our threads
 	sickbeard.start()
 
--- a/data/interfaces/default/comingEpisodes.tmpl
+++ b/data/interfaces/default/comingEpisodes.tmpl
@@ -12,14 +12,13 @@
 <div id="outerWrapper">
   <div id="listingWrapper">
 
-
 	#set $cur_segment = None
 	#set $too_late_header = False
 	#set $missed_header = False
 	#set $show_div = ""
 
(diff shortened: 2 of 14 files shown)


midgetspy:
I have a feeling this won't work with the new notification system - can you rebase onto the new_config branch? Thx for the contribution.

aforty:
rebased to new_config with new pull request

midgetspy:
OK sorry to jerk you around, new_config is having issues so I am going to use your original one and pull to master. I've pulled to an ajax_ep_search branch in my repo for now because there's a problem with multiple searches. Because the searches aren't being queued in the back end if you click many search links quickly you are searching your providers simultaneously which can violate their API rules. There are a few solutions to this.

The easiest one is to allow only one search at a time on the client side - if a search is ongoing just disable the other links or reject them when they're clicked. This makes the feature less useful though, heh.

The more difficult solution is to buffer the requests. This could be done in ajax on the client side for now, if only because it's easier than doing it on the server side. Queueing requests like this is something that needs to be done server side eventually but it's a big job so I haven't gotten around to it yet. Doing it client side is counter-intuitive though, since people will probably assume if they clicked the link it's being done server side.

aforty:
No worries. I actually like the implementation in new_config better as it provides feedback using the new notification dialogs (only difference is in .js) instead of only displaying result in the search field area (where they clicked and are unlikely to return to see the result).

I realize now that there are issues with this, I wasn't aware that API rules might be violated with this. Maybe that's something I can take a look at. I definitely think that buffering the requests server-side is the only real way to fix this.

For now though I do think the new_config implementation is better if you can get that to work.

PR #47, opened 2010-12-17
perform episode search via ajax

aforty:
Hi there,
Love Sick-Beard, wanted to help. Hopefully I did just that.

Changes episode search from show page to use an ajax call. Updated webserve.py-searchEpisode but it still works the same way if needed (wasn't sure if you referenced it from some other place) and cleaned it up a bit while I was there.

Hopefully this is helpful.

Thanks!


aforty:
This is a duplicate, sorry, still learning github...

PR #50, opened 2010-12-20
Daemonize

sgranade:
Added ability to run Sick Beard as a well-behaved Unix daemon. The script can be turned into a Unix daemon by running it with the --daemon (or -d) option.


midgetspy:
Your SickBeard.py line endings changed so your diff is broken - can you undo it and commit again with the windows line endings?

sgranade:
Done. Would you be willing to switch SickBeard.py to use LF-only line endings in the master? Python doesn't care, and having LF endings allows Unix users to run it as "SickBeard.py" instead of "python SickBeard.py"

midgetspy:
Yeah we can probably do that once it's ready to do pulled.

I see you set sickbeard.DAEMON but it never seems to be used - what's it for? Why have it as well as "fork" in SickBeard.py?

sgranade:
I started out using a local variable, then realized that it's information that should be stored in the sickbeard object for future expansion. Then I forgot to get rid of "fork". I've just pushed a version that uses sickbeard.DAEMON instead of the local fork variable.

midgetspy:
Merged.

PR #51, opened 2010-12-20
NFO file secondary check.

wyze:
If Sick-Beard is unable to find a matching .nfo file with the same name as the video file, it will try and find one that matches the name of the parent directory where the file is located.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -922,7 +922,11 @@ class TVEpisode:
             if ek.ek(os.path.isfile, sickbeard.metadata_generator.get_episode_file_path(self)):
                 self.hasnfo = True
             else:
-                self.hasnfo = False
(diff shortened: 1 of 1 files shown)


midgetspy:
What is the use case where this is desired?

wyze:
I keep my collection in archived format the way it is released and here is one example:

The.Big.Bang.Theory.S01E01.720p.HDTV.x264-CTU
ctu-x264-the.big.bang.theory.101.rar
the.big.bang.theory.s01e01.720p.hdtv.x264-ctu.nfo

midgetspy:
What do you make the nfos with? I don't understand the reason for such a layout?

wyze:
I don't make the nfos. That is how they are named when they get downloaded. I never create my own nfo files.

midgetspy:
The NFO checkbox is meant for the NFOs SB generates with the metadata settings, not scene NFOs. This logic is also changed in the new_config_v2 branch since the metadata stuff was changed there too.

PR #57, opened 2011-01-06
Configurable Season Thumbnails

mattaltus:
I noticed that the season thumbnails are being written to seasonXX.tbn despite what the season folder format is set to.  So I patched it so that the season thumbnails get written as the same format.


mattaltus:
So I made the changes to the MediaBrowser metadata to get to write to the correct folder, could most likely be tidied up though.

PR #59, opened 2011-01-10
Added twitter_prefix config option to web config-notifications page

twonk:
Hi,
I've added the twitter_prefix config option to the web config-notifications page.  Only a small addition to make the option available via the website rather than just the config file.

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -254,6 +254,16 @@
 
                         <div class="field-pair">
                             <label class="clearfix">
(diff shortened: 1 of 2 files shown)


midgetspy:
It's intentionally not on the config page, it is an option that very few people need so it just contributes to clutter on the config page. Feel free to maintain a fork with it in config though :0)

twonk:
I did wonder why it was missing. It is quite a useful option because it can be used to @yourself, and a lot of twitter clients will notify you on a mention rather than missing it on the timeline. 
No worries though, I'll stick with just using the config file. :-)

PR #60, opened 2011-01-12
Viraptor/endings

viraptor:
Migrate dos -> unix endings. Running SickBeard.py doesn't really work otherwise on linux.


colstrom:
Alternatively, invoking it with:

$ python SickBeard.py

works fine under Linux.

viraptor:
Yeah, but that's not nice really - if there's a shbang line on top of the file normally you'd expect it to "just work"...

colstrom:
Have you tested this fix on Windows? Not a user myself, but intuitively it seems like switching the line endings would resolve the problem on Linux, but create it on Windows... unless I'm not clear on something (possible).

viraptor:
"In source files, any of the standard platform line termination sequences can be used - [...] All of these forms can be used equally, regardless of platform."
http://docs.python.org/reference/lexical_analysis.html#physical-lines

colstrom:
That's all well and good when invoked through the Python interpreter. However, it's the shebang that tells the host OS which interpreter to use.

The shebang is not interpreted by Python, negating the validity of that documentation.

viraptor:
Ok - from the beginning: there's an issue with the main file, because the shbang line ends with "\r\n". When you try to execute the file, Linux tries to look for an interpreter called "python\r" instead of "python". Line endings don't matter for the script execution once the interpreter is loaded. shbang line doesn't matter to windows users, since Windows does not use it and for the interpreter it's just a comment.

Therefore -> the change doesn't make any difference for Windows setups but does actually fix a broken shbang on Linux.

colstrom:
Windows is unable to use shebangs? It _has_ to be invoked as 'python SickBeard.py', even when using platform-supported newlines? Weird...

(Sorry, my query was speculative, as a non-Windows user.)

If the newlines beyond the shebang can be either form, why not simply patch line 1 to use \n only, and add a \r\n on a line between that and what (the current) line 2? This would preserve the visual appearance for folks coding on a platform that uses \r\n, while resolving the Linux-specific issue with the fewest changes possible. (1-2 lines vs 230 lines)

I think this may be more suitable. I'm assuming those \r\n newlines got there as a result of a developer's platform choice, and future additions may introduce more of them, unless their environment is modified accordingly.

viraptor:
There is no /usr/bin/... on windows, so it cannot "use" shebangs in any way. On windows machines, the standard is to map the extensions, so it's more likely that the .py itself will be bound to the interpreter.

You can of course switch only the first line to a different line ending, however many decent editors will make the line-endings the same type if you mix them up - which is arguably a good thing... Then again almost any decent editor can also preserve the already-used line ending style.

Github's page seems to suggests to just go with CRLF endings everywhere http://help.github.com/dealing-with-lineendings/ - what's the best way here depends on the project really, but I'd advise against mixing the endings - at least in a single file.

colstrom:
_"To be safe, you should set your config to convert line endings on commit so they are always LF in the repo [...]"_ (taken from your link) does not advocate "CRLF endings everywhere" at all.

Though that article does indicate that Windows systems may mangle line endings automagically (see _"I just cloned and git says files have changed!"_). **How do we resolve this in a manner that prevents that from recurring?**

If switching the first line will resolve this problem, and editors may or may not automatically alter line endings (this raises the question of _which editors are used most widely with this project_...), perhaps making the smallest change possible to resolve the issue is the cleanest solution?

DisasteR:
For me Switching all the project to LF it's the best things to do.
Windows doesn't car about line endings and any good windows editor can manage LF line ending.

midgetspy:
The problem with switching is you change every line in the whole project so every pull request, every branch, every fork are all broken and unmergable. At some point where SB is at an obvious turning point (beta?) maybe I will do this but for now I am not going to bother. This repo started by getting pulled in from an svn repo which is why we have this problem in the first place.

colstrom:
_nods_ Changing the endings project-wide is a very large hammer for a relatively small problem. However, altering them as proposed in my last comment on Mar 04 would resolve the alleged "problem" without mucking with the rest of the code, and thus branches/forks/pulls/etc.

Alternatively, my comment on Feb 23 outlined a simple workaround that requires zero code changes, though it is every-so-slightly "less pretty".

viraptor:
Woah - aren't you all overthinking it a bit? The entry file's shbang is broken. You can fix it by chaning endings on a single line (and risk IDEs reverting it), or single file (as in pull request) - not the whole project (line endings don't matter for the not-executable files anyways). How often do you need to change this file? Why did we get to 12 comments over 5 months?
Either delete the broken shbang (and make 'python' required), or convert the file and just be done with that :D Other pull requests are more worthy of your attention.

colstrom:
I'd wager the period these comments span is indicative of precisely that point: This is a very minor issue, and other requests have correctly received higher priority to date.

As for the resolution, yes. A single-line fix would be ideal, if a fix is actually needed. Assuming nothing else breaks as a result of it.

ghost:
A single-line fix have been applied already in 39c4fefe93e27a65cce6673439c13f002d20d7ee

bungiman:
Should this pull request be officially closed now?

PR #61, opened 2011-01-14
Small change to coming episodes template

garlandkr:
Hey,

Thanks for a great app, I was able to throw away a bunch of random scripts I've made over this nice interface.


midgetspy:
So if it's thursday and an ep airs on monday it would say "Next Monday"?

garlandkr:
Correct, as it is based on the isocalendar week of the ordinal date.

PR #65, opened 2011-01-23
Viraptor/caching

viraptor:
Cache objects aggressively (this helps tvdb api to not regenerate objects when not needed).
Locally, got rid of most of the useless tvdb_api.py:712(_getShowData) calls. Changes for HIMYM series:

Number of calls: 538 -> 137
Total seconds: 27.9 -> 8.4

Also stripped most of the sqlite writes happening during update.


midgetspy:
Pulled. Great work, thanks!

PR #69, opened 2011-01-25
Coming Episodes image scaling improvements; Frontline (PBS) scene alias

giantspook:
This changeset improves the scaling of posters and banners.  Most importantly, it corrects subtle aspect-ratio issues that occurred because the Image.thumbnail() method was used instead of Image.resize().  The former maintains image aspect ratio, meaning that the resized image was then scaled by the browser one or two pixels, resulting in jaggies.

Instead of returning the default image when no poster/banner is found, a 404 error is returned.  This avoids re-resizing and re-downloading the default image a bunch of times.  The default image is shown as a CSS background image.  I have avoided resizing the background images, thought this changes the appearance of things a bit (you may consider this a regression).

HTTP Etag (caching) support has been added to the thumbnail scaling too.  This means that the second time the image is requested the server can respond with 304 Not Modified before even loading or scaling the image.  This should help make things more snappy.


giantspook:
I've moved this stuff to the branch thumbnail-fixes.  I'm afraid that my poor (still learning) git-foo may cause you trouble as I rebased to upstream master in the process; if so, I apologize.

PR #73, opened 2011-01-31
Thumbnail fixes

giantspook:
Fixes to the coming episodes thumbnails, adding better compression, etags (browser caching), and correct image sizes.


midgetspy:
The placeholders are intentional... some images don't exist for some shows, showing a 404 is bad behavior IMO.

giantspook:
I made the placeholders CSS background images, so they show when the 404'd image fails to load.  They also only get loaded once, and are displayed before the poster/banner image loads.  I think that this is preferable to resizing and downloading the placeholder image many times.

I didn't resize the placeholder images as I figure smaller images should be generated from the original source (a vector file, I assume), but I can do so and add it to this branch request if you like.

giantspook:
I've added resized versions of the placeholder images with commit 2145ad9.

midgetspy:
Tried pulling this tonight, it doesn't seem to work in Chrome... the 404's aren't replaced with the placeholders, I just get broken images.

giantspook:
Hmm, I didn't have any issue with Chromium when I tried, but I'll look into it.  Thanks for giving it a look.

midgetspy:
I tried it on 9.0.597.83 beta, including clearing cache, didn't seem to help. It worked on Firefox though... if it works for you I'll pull it into a branch and get some others to beta test it.

giantspook:
I was able to reproduce and have fixed the issue.  Now the image tag is only inserted if the poster or banner exists in the cache, so we won't get 404s at all.

midgetspy:
  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy_cprequest.py", line 660, in respond
  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy\lib\encoding.py", line 193, in **call**
    self.body = self.oldhandler(_args, *_kwargs)
  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy_cpdispatch.py", line 25, in **call**
    return self.callable(_self.args, *_self.kwargs)
  File "C:\Users\Nic\workspace\SickBeard\src\sickbeard\webserve.py", line 1998, in comingEpisodes
    ep = dict(row)
TypeError: 'sqlite3.Row' object is not iterable

giantspook:
What version of Python and sqlite3?  I'm using 2.6.6., sqlite3 version 2.4.1 and don't see this.

giantspook:
I've added compatibility with Python 2.5, which I assume is what you're having problems running it under.

elmarkou:
After update todat to latest It will not show the real banners and posters anymore. It will only show the sickbeard poster

giantspook:
@elmarkou I have synced with master.  Does that fix the issue?

elmarkou:
fixed! thx

giantspook:
I give up.

PR #76, opened 2011-02-02
Tvtorrents support

tester22:
This pull adds support for the provider tvtorrents.


midgetspy:
I think before I pull this I will make some more changes for nzb vs torrents... cause the vast majority of people won't want this provider (and most don't want ezrss either).

Alfiegerner:
I would love this. Am I the only one using sickbeard just with torrents?

Alfiegerner:
Installed and tested, works great, thanks very much.

anothergene:
I would use this as well.  They have a huge collection!!

midgetspy:
Unless they have a search API this would only work for new episodes FYI.

tester22:
They don't have any API but I am working on a scraper that might work.
The only problem is that after an entire season has ended a user usually upload all files as one torrent and removes the single episode torrents. This limits the searching since we can't request a single fil from a torrent for download

midgetspy:
I know pretty much nothing about .torrent files but it could be possible to edit the torrent file to mark some files as skipped or remove them, yeah?

midgetspy:
According to wikipedia it looks like there's no reason it wouldn't be possible...

tester22:
But this would require alot of changes in sickbeard, currently it only downloads the files and marks them as snatched.
To implement this it would need to download a file that might be correct, edit it so that it only will download certain files that might be included in the torrent and if that succeeds mark the file as snatched. But for this provider this only relates to episodes that are older then the current running season so for most users they wouldn't need that functionality. And I think it would be fare to require them to manually download the torrent from the website if they need it.

rascalli:
I have an account onTVtorrents , so if it goes into a branch of midgetspy I am willing to test

Alfiegerner:
FWIW I'd be happy to have just for new episodes, if i want to pick prior series from tvtorrents i can do that manually.

midgetspy:
tester22: SB does its backlog searches one season at a time so this is actually pretty easy. It's already being done with NZBs. If SB finds a full season NZB it downloads it and tries to split it into separate NZBs and then treats them as individual results. The same could theoretically be done with torrents - split them in-place, treat them as separate results, and then download the ones that are needed. That said, I'm definitely not going to bother ;-P

midgetspy:
Pulled! Backlog would make this a pretty good option for some people if somebody wants to implement it :0)

Alfiegerner:
coool, thanks midgetspy! back on trunk now. Cheers tester22, nice one.

PR #80, opened 2011-02-06
Missed episodes

wolph:
Because I only run SickBeard once a week on one of my machines, I need to configure the "missed episodes" feature to ~2 weeks to make sure I never miss any episodes.

See this topic for more info: As described in this topic: http://sickbeard.com/forums/viewtopic.php?f=4&t=985&start=0


midgetspy:
This will only increase load on providers for a scenario that I can't imagine any other user having - doesn't belong in master IMO.

wolph:
Can you think of any other fix/workaround that would be appropriate for the master branch?

midgetspy:
Not really - running SB less than once a week is not how it's meant to be used and I've never heard of anybody else with such a problem... you can always just run your own fork and pull from my master to keep updated though.

midgetspy:
BTW I appreciate the contribution, I just have to be fairly vigilant about which features get added and which don't or SB will balloon and become much more complicated an more difficult to maintain :-|

wolph:
Ok, I've been running my own forks for a couple of months already (my github is out of date though). But I'd rather keep it synced with your master.

Perhaps an alternative solution might be to add a button to force a search for all items currently in the `comingEpisodes` view that are `missed`. Cause it seems that it regularly does not download all of those while it should be according to the code. That shouldn't really increase the load on the providers (I mean... how many missed episodes can you have in a week?) and still gives you the problem to search manually if for some reason the episode wasn't downloaded.

midgetspy:
What you're describing is what used to be "Force Backlog". A search for all missed episodes would include ALL missed episodes - there could be thousands if you have set old seasons to Wanted. This is done every 21 days automatically. In an effort to solve this exact problem, episodes less than a week old are also searched a few times a day... but if your SB is turned off for too long then you end up circumventing that feature. If you can think of another solution that doesn't encourage/allow people to search their whole backlog far too often I'm open to the idea...

wolph:
I'm describing a limited version that would only try the episodes that should have been aired the last 7 (or perhaps slightly more) days. I know it does that automatically once in a while, but at times I don't like waiting ;)

I don't think there is any need for a full backlog search, only the recently aired subset is needed.

midgetspy:
The limited backlog runs twice a day and searches for Wanted episodes aired within the last 7 days. It also runs when you start SB so I don't think that there is any need to allow it to be forced as well.

wolph:
Perhaps I'm just quite unlucky in that case since it always seems to take atleast 6 hours before it starts the search automatically after starting SB.

I'll keep an eye on the logs to see if I can find the reason for that unexpected behaviour.

PR #82, opened 2011-02-07
HTTP Proxy support for TVDB_API

tlusk:
# HTTP Proxy Support

The majority of SickBeard is written using urllib for http communication.  urllib has built in support to read the http_proxy variable in linux, and to check the registry settings in windows (or read the http_proxy env variable) and to use that proxy for http communications.

The problem is that the tvdb_api library uses httplib which does not support automatic proxy detection.  This causes a strange behavior where if a proxy is required and set, the communication to search for a show works normally, but trying to download the episode lists fails.

The changes I made add a http_proxy configuration that can be passed into the tvdb_api in order to configure httplib to properly use the proxy.  The system proxy settings are read from urllib and put into the tvdb_api_param on initialization, therefore all of sickbeard will now be using the http proxy.

httplib requires the SocksiPy library in order to do any kind of proxy communication, so it is included in the libs folder now.

### DIFF
--- a/lib/httplib2/__init__.py
+++ b/lib/httplib2/__init__.py
@@ -54,10 +54,14 @@ import hmac
 from gettext import gettext as _
 import socket
 
+# Try using local version, followed by system, and none if neither are found
 try:
-    import socks
+    import lib.socks as socks
 except ImportError:
-    socks = None
+    try:
+        import socks as socks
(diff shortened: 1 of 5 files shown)


tlusk:
I broke this out into a separate pull request to unclutter the other one.

PR #83, opened 2011-02-09
Libnotify notification support

giantspook:
This adds support for the libnotify notification system, used on Linux and *nix desktops.


midgetspy:
Pulled, thx.

PR #84, opened 2011-02-10
Sab Q limit

ob0t:
Have written a mod that allows users to check the SABnzbd queue size before submitting an NZB. It fires an API call to SAB to check queue size in MB against a configured limit - if the queue is too big the NZB is not sent to SAB and the title remains in the 'wanted' state. This prevents the SAB queue from getting too large and causing things to grind to a halt.

### DIFF
--- a/data/interfaces/default/config_episodedownloads.tmpl
+++ b/data/interfaces/default/config_episodedownloads.tmpl
@@ -193,6 +193,17 @@
                                     <span class="component-desc">Category for downloads to go into (eg. TV)</span>
                                 </label>
                             </div>
+                            <div class="field-pair">
+                                <input type="checkbox" name="sab_limit_q" id="sab_limit_q" #if $sickbeard.SAB_LIMIT_Q == True then "checked=\"checked\"" else ""# onclick="document.getElementById('sab_qlimit').disabled=document.getElementById('sab_limit_q').checked?false:true" />
+                                <label class="nocheck clearfix">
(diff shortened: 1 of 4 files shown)


rascalli:
I must say I like this option.
But will it then send again the next search ? (some people have a search every 15 min , won't that overload the API ?)
Or should it set to wait like 1 day ??

akuiraz:
This is a must have feature for those of us that are running their mediacenter's on lower end hardware as the amount of memory SABnznbd uses increases dramatically with every additional NZB added into the queue.  Hope to see this integrated into Sickbeard very soon!

unikitty37:
Does this work for all searches? It would be useful to be able to configure it to only trigger on backlog searches and not new episodes, so a person with a large queue doesn't have it bloated by a load of old shows, but they still get to see the new Dr Who without manual intervention...

akuiraz:
+1 for the option to only trigger on backlogs.

thezoggy:
i would think that you should mention the idea to sab vs having sb control sab.. sb should just throw everything it wants at sab.. sab should be smart enough to know how to handle things efficiently. you using sab 0.6.5? many things have been changed to sab with the 0.6.x dev to handle things better...

aussetg:
Could be useful.... My queues is 25 pages long / 1.5 Terabytes big

thezoggy:
sab already has a queue/quota manager in 0.7.x. so this pull may not be needed

PR #85, opened 2011-02-11
Notifo notifications

jeroen94704:
This adds support for Notifo notifications. I made this because there is no Growl/Prowl client for Android phones, and I absolutely detest Twitter :). Notifo uses the push-notification infrastructure of iPhone/Android/BlackBerry/Win 7 devices (although the latter two are not yet available). 

BTW: I've been wrestling a bit with Git (which is new to me), so while there are no test commits or reverts in the history, there may be a couple of not-too-relevant messages. I hope this is acceptable as-is.

Jeroen


midgetspy:
Pulled, thanks!

jeroen94704:
Awesome. Glad to be of help.

PR #86, opened 2011-02-14
Temporary work around for Issue 459

ShaneMcC:
Temporary work around for Issue 459 - add a "Default Show Directory" setting and use it in the Add Show page.


midgetspy:
The actual fix is almost done in the new_add_show branch so I'm not going to bother, thanks though.

PR #88, opened 2011-02-20
Additional Episode Number Style

mattprecious:
I added "02" to the episode number style options. I like to keep season folders and I prefer to not have season numbers in my filenames. I'm pretty sure this isn't uncommon.


midgetspy:
I'm not adding any new episode renaming stuff because I'm planning to redo it from scratch and the new system will support arbitrary naming. Thanks though!

PR #93, opened 2011-02-24
Use RegExp for sanitizing & Backlog searching with NZBMatrix documentary categories

nberlee:
Regular expressions are efficient in replacing characters.

I understand why the periodically RSS request only uses TV SD and TV HD categories (because of the 50 shows limit)..so, left that unchanged... 
I don't see why backlog searches could not search documentary categories.....?

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -98,10 +98,8 @@ def sanitizeFileName (name):
     >>> sanitizeFileName('a"b')
     'ab'
     '''
-    for x in "\\/*":
-        name = name.replace(x, "-")
-    for x in ":\"<>|?":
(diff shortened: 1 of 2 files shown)


midgetspy:
The backlog search uses the RSS too, if you look at the URL it's rss.php.

nberlee:
I meant, using a term= parameter (when backlogging) reduces the possible results by such a number, it is safe to add more categories (such as documentaries) to the search, because it probably won't come near the 50 limit.

midgetspy:
Unfortunately it comes close to and often goes over the limit as it is - a show with 20+ eps will already go over when you have SD XviD, 720p x264 and h.264 WEB-DL releases for each episode.

PR #94, opened 2011-02-25
Changes to common.Quality.nameQuality to improve accuracy

wrboyce:


midgetspy:
It definitely matters, nobody wants WMV or whatever. This introduces ambiguities and breaks existing settings ("hr.ws.pdtv.x264" would go from 720p to SD TV, for example), all to solve a problem I'm not aware exists :0)

wrboyce:
I have to say, I disagree (with the design decision), one quite a few counts:
- The assumption that certain codecs must be used for media to be of a certain quality is just incorrect.
- The assumption that /nobody/ wants WMV "or whatever" is incorrect, and should be a configurable option for SickBeard.

Would it not make more sense, to perhaps allow the user to provide a list of preferred codecs? Something like the code I've previously used to do this at https://gist.github.com/0580226f8c3c831ae4db could be helpful. Combine this with some config screens to allow the priorities to be changed and I believe we're onto a winner, leaving the current Quality checker to only concern itself with resolving a human readable name for a quality setting (for example, the current Quality checker would fish out all desired quality downloads available, say, anything HD. After that it could order the filtered list by the user's codec and resolution preferences, and then pick the top from the list - perfect).

I'll be happy to work with you guys to get the above gist implemented into SickBeard in a satisfactory manner.

midgetspy:
Nobody is assuming that codecs have anything to do with quality... the assumptions have to do with what the codec/resolution rules are for the scene. The current SB quality settings are tailored to work correctly with all possible qualities of scene releases.

Properly supporting qualities for non-scene releases is quite complicated and something I have no interest in implementing myself. It is a feature that very few users want and it is a lot of work. If you end up implementing this yourself in a way that doesn't impact the current behavior then I have no problems pulling it but I have no desire to work on it myself.

PR #95, opened 2011-02-26
Notification option for the Plex Media Server

sander1:


midgetspy:
Pulled to the plex_notifier branch, if you could test it for me I can pull to master. I don't have any access to an OSX install :-|

PR #98, opened 2011-02-28
Changed Growl support to make it work with Growl 1.2 on Snow Leopard (10.6)

sHorst:
Hi,
i changed the Growl implementation to the Growl.info reference implementation. Which works over UDP and not TCP. 
BR
Stefan

### DIFF
--- a/lib/growl/netgrowl.py
+++ b/lib/growl/netgrowl.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python
+
+"""Growl 0.6 Network Protocol Client for Python"""
+__version__ = "0.6" # will always match Growl version
+__author__ = "Rui Carmo (http://the.taoofmac.com)"
+__copyright__ = "(C) 2004 Rui Carmo. Code under BSD License."
+__contributors__ = "Ingmar J Stein (Growl Team)"
+
+import struct
+import md5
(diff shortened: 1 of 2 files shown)


midgetspy:
Does this still work with Growl for Windows?

sHorst:
haven't tried, since i do not have a windows but its the growl script from the growl.info svn so it should. i only commented the old code, so it can be changed with a flag, if it is not working any more

sHorst:
just checked with the current version of growl for windows and it works as expected.

sHorst:
some further infos to the windows/mac growl:

the udp version of the growl protocol is supported by both versions (but only basic messaging is working, but no images)
they both use different tcp protocols. (where images should work, but needs to do seperate implementation)

nifoc:
Please note that once Growl 1.3 (for Mac) is released, the current implementation will "magically" start to work, since GNTP support is scheduled for 1.3.

sHorst:
ok if that is so, than this patch is obsolete, do you know, when this update is scheduled? And for the time being I can use my Patch or use the gntp proxy

nifoc:
Sorry, I don't know when version 1.3 will be released … all I know is that they talked about 1.3 for probably almost a year by now.

PR #99, opened 2011-03-01
Alternative to ManuelAC's chmod

nberlee:
An alternative for ManuelAC's chmod def... It has more predictable behavior for novice users since it chmods according to the parent directory and also has no error on windows machines

I didn't change postProcessor.py as you might combine the two so more advanced user can set there mask using ManuelAC's code ...


nberlee:
Not sure why midgetspy's last commit is also in here ?!

PR #100, opened 2011-03-02
Alternative to ManuelAC's chmod

nberlee:
An alternative for ManuelAC's issue 152... It has more predictable behavior for novice users since it chmods according to the parent directory and also has no error on windows machines... also, meta files have inherit permissions


midgetspy:
Pulled, thanks.

PR #101, opened 2011-03-03
.gitignore minor update: .pyc, emacs scratch file, group sickbeard-specific ignores

mjc:
Figured it would make sense to group the sickbeard-specific ignored files at the bottom so that it's easy to find them all, should the file grow larger.  .gitignore supports comments so adding something like "#sickbeard-specific ignores below" would work too.

I added .pyc (as I end up with those as often as I end up with .pyo anyway), and *~ to ignore emacs auto-recover files, since there already are a few ignores for text editor and file browser litter.

Normally I wouldn't send a separate pull request for something so small but I didn't want it to get wrapped up in my next pull request (initscript enhancements) so I'm hoping this does the trick...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
-*.pyc
+*.py[co]
+.DS_Store
+*~
+*.bak
 cache/*
 cache.db
 config.ini
 Logs/*
 sickbeard.db*
 autoProcessTV/autoProcessTV.cfg
-*.bak
-.DS_Store
--- a/initscript
+++ b/initscript
@@ -10,7 +10,33 @@
 # Description:       starts instance of Sick Beard using start-stop-daemon
 ### END INIT INFO
 
-############### EDIT ME ##################
+### DON'T EDIT ME
+## place your customizations to these variables in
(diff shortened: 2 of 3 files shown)


midgetspy:
.gitignore should contain only things related to SB - if your environment requires extra ignores you should keep those in your fork IMO. I need to clean the file up and .pyc can be added, thx.

mjc:
Right, that's what my commit log basically mentioned (rather than the pull request). I'd assumed that there might've been trouble with other commiters forgetting or not realizing they needed to use .git/info/exclude, as is often the case with .DS_Store.

midgetspy:
Can you rebase this onto the init.ubuntu script? This got lost in the .gitignore pull req...

thezoggy:
the .gitignore suggestions in this pull request were taken into consideration when moving sab over to git. this pull can be squashed and redone with just the init script. please see: https://github.com/midgetspy/Sick-Beard/pull/199

thezoggy:
this pull request can be closed as the changes are already in master now.

PR #102, opened 2011-03-03
debian initscript enhancements: multiple instances, configuration in /etc/default

mjc:
As mentioned in the commit log; the initscript can be symlinked to /etc/init.d/anything, now, and the configuration will be picked up (based on the basename of the symlink) in /etc/default, if it exists.  This allows for multiple instances of sickbeard (each in their own separate directory, of course) to be started by initscripts.  

Additionally since the configuration edits would reside in a file not managed by git, the initscript itself can be updated by git or any other method safely in the future without losing or possibly incorrectly merging edits to the script.

For some reason the pull request feature won't remove my .gitignore commit from this request, I already sent a pull request about it separately to try and avoid this but oh well. :P

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
-*.pyc
+*.py[co]
+.DS_Store
+*~
+*.bak
 cache/*
 cache.db
 config.ini
 Logs/*
 sickbeard.db*
 autoProcessTV/autoProcessTV.cfg
-*.bak
-.DS_Store
--- a/initscript
+++ b/initscript
@@ -10,7 +10,33 @@
 # Description:       starts instance of Sick Beard using start-stop-daemon
 ### END INIT INFO
 
-############### EDIT ME ##################
+### DON'T EDIT ME
+## place your customizations to these variables in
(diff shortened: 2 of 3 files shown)


midgetspy:
You should do every change in a separate branch to keep them separate.

mjc:
Yeah, apparently github needs a cherry-pick request or something. haha.  I'll see about fixing something up so you don't have to pull both...

midgetspy:
I can always cherry-pick them myself now that they're here...

mjc:
I suppose, I sent myself a pull request to test it and didn't see an easy button, and so was figuring it might be easier for you if I just did it right (if so, mjc/Sick-Beard@80c9a141 is in a separate branch, heh).  I suppose I need to play around with github more often as I'm used to doing it all on the commandline via git-am and git-send-email...

HTH,
mjc

mjc:
Nevermind, this blasted script is still a bit buggy.  Closing the request for now, I suppose, and when it's ready and a bit better tested (along with perhaps a handy Makefile for installation purposes) I'll send another.

PR #105, opened 2011-03-09
NMJ background scan notification support

nberlee:
The Networked Media Jukebox, or NMJ, is the official media jukebox interface made available for the popular Popcorn Hour 200-series media player device.

Most configuration can be done automatically...


nberlee:
Changed so it works with the 20-3-2011 config commits

PR #108, opened 2011-03-14
basic icalendar / ics

anothergene:
I added ical/ics support for coming episodes.  Pretty basic, but it works.  No on the page yet.  

I want to expand it to include all coming episodes, not just the next.  possible give the user that option.


anothergene:
just need some help with getting the link on the coming eps page.  I assume that would be zoggy?  I would hate to mess up all his great work on the interface.

anothergene:
having some issues with this.  Need to reevaluate.

PR #109, opened 2011-03-20
NZB Method: NZBget (based on XMLRPC)

nberlee:
NZBget is a very light-weighted binary newsgrabber for nzb files (written in C++, and still maintained). It can sustain high constant download speeds on low cpu powered devices. (especially when you compare it to SABnzbd)  
This makes it popular on mediaplayers and NAS devices (mostly pre-installed).

This commit will add NZBget support (remote or on the localhost) so NZBget users (like me) won't have to wait before the nzb is picked up from the blackhole... And prioritize recently aired episodes just like in sab.py

The XML-RPC code is thanks to http://nzbget.sourceforge.net/RPC_API_reference

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -85,8 +85,8 @@
                                 <span class="component-title jumbo">NZB Method:</span>
                                 <span class="component-desc">
                                     <select name="nzb_method" id="nzb_method">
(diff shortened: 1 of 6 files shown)


midgetspy:
I think the config part of this will have to get redone after todays commits, thanks.

nberlee:
Yes, great timing! :)

I changed it.......
I hope you pull this soon, so I don't have to do it again... :)

DJClean:
This method seems to fail when using newzbin as provider.
For some reason it sends the newzbin login page to nzbget

PR #110, opened 2011-03-20
Added PID file support into SickBead with --pidfile=/path/to/file.pid command-line parameter

Djelibeybi:
Also added a sample Fedora init.d script and renamed the existing initscript to init.ubuntu for clarity.


Djelibeybi:
Fair call. I'll clean that up and re-issue a pull request later.

PR #111, opened 2011-03-21
Added PID file support into SickBead with --pidfile=/path/to/file.pid command-line parameter

Djelibeybi:
This pull request adds support for writing a PID file via the --pidfile=/path/to/file.pid command-line parameter. It also checks to make sure we can write the PID and cleans itself up on successful exit. Now that this support is in place, a proper init.d script for Fedora has been added and the existing initscript has been renamed to init.ubuntu for clarity.


midgetspy:
generica if you can confirm those changes work for ubuntu can you maybe fork so Djelibeybi can pull them from you so this can be pulled as a single feature?

Djelibeybi:
If the changes work, I'll just add them into my local repo. Much simpler than fork/pulling. :)

Djelibeybi:
I've modified init.ubuntu to use Sick-Beard's internal PID creation and daemonize options. This has been tested on Ubuntu Server 10.10.

midgetspy:
Pulled.

PR #114, opened 2011-03-29
daemon friendly options

travisghansen:
This adds 3 options
1. --datadir
2. --config
3. --logdir

All of these would be very helpful to keep data out of PROG_DIR which is not pretty for packaging.  This would be beneficial as well on embedded systems such as openelc which has a read-only root filesystem.


midgetspy:
FYI for next time, always work in a branch.

PR #116, opened 2011-03-31
Check filenames in the .torrent if the .torrent's filename doesn't give a quality

rmccue:
Occasionally, files will be missed because they have ambiguous filenames for the torrents, e.g. `(HDTV-LOL)`, although the filename of the actual file will be fine, e.g. `HDTV.XviD-LOL`

This patch fixes this by checking that first. It's somewhat inefficient, in that it requires downloading the .torrent file to check, and if it matches, that will then get downloaded again. However, this is not really a large issue.


bardelot:
Instead of adding a new parser you could use the existing one from hachoir which is already included in Sick-Beard.
e.g. to extract the filename use something like:

```
from lib.hachoir_parser import createParser
from lib.hachoir_metadata import extractMetadata
from lib.hachoir_core.error import HachoirError

try:
    parser = createParser(u"/home/user/file.torrent")
    metadata = extractMetadata(parser)
except HachoirError:
    return False

if not metadata.has('filename'):
    return False

filename = metadata.get('filename')
```

rmccue:
The problem with using that parser is that, AFAICT, we have to download the file first. Alternatively, we can use a stream from urllib2, but then we can't use `helpers.getURL`, and I believe we also lose the ability to parse gzipped torrents.

Should I convert this over?

thezoggy:
hachoir is for reading a stream.. 
https://bitbucket.org/haypo/hachoir/wiki/hachoir-parser
- torrent: Torrent metainfo file

PR #119, opened 2011-04-02
Notification fixes

JoshSchreuder:
- Fixed libnotify test button that was finding sbRoot incorrectly in JS
- Fixed Notifo if a bad username / API key was entered resulting in an IOError


dmwyatt:
zoggy said to post this traceback here....

http://pastebin.com/rUXujCaz  

All of a sudden getting a lot of failures on notifo notifications.  Not every file, but a good portion of them....

PR #124, opened 2011-04-05
Libnotify exception fix

JoshSchreuder:
- Catch GError exception 'display not found' when calling n.show() in libnotify.
- Modified .gitignore slightly as it was picking up cache.db-journal in commits

It's unclear to me why the comment in libnotify.py says that error handling doesn't work too well. 
This seems to work.

I took the libnotify fix from:
https://bugzilla.redhat.com/show_bug.cgi?id=667350

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *.pyc
 cache/*
-cache.db
+cache.db*
 config.ini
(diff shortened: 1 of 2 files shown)


midgetspy:
Awesome, thanks.

PR #125, opened 2011-04-05
Adding Another RSS Torrent Provider (Kickass www.kickasstorrent.com)

mr-orange:
http://www.kickasstorrents.com/

Signed-off-by: mr-orange mr_orange@hotmail.it


bardelot:
- You should also update the logging part, it still contains the references to EZRSS.
- And you may also want to filter the search results to get TV Shows only. e.g. use "?category=Other-TV&rss=1".
- I don't think you can search by air-date, thus you don't have to comment out that part.
- The spaces in the show_name parameter should be replaced with "-".
- If you do not need all functions you should remove those. e.g. "_extract_name_from_filename"
- You could probably also add an option to download verified torrents only, that would make it a little bit more special.

midgetspy:
I'm not sure how useful this would be as a provider, with wonderful release names on it like "The Event S01E09 Your World to Take (subITA)-WTRG", "Magical DoReMi Eng Dub all", and "opinionology1.MP4".... and it looks like all the good torrents come from EZTV anyway?

mr-orange:
@midgetspy & @bardelot the provider search "show_name-SXXEYY" and names like "the event" or "magical doremi" are not a problem, kickass torrent parse in automatic the correct showname (The-Event or Magical-Doremi) even if i search the showname with spaces, however i can modify the code to replace the ' ' with '-'.

@midgetspy Yes, a lot of good torrents are From EZTV, but sometimes EZRSS do not have 720p-HDTV show. An example are "Family Guy" and "Lie To Me", in this cases Kickass can snatch 720p releases. This type of failure has prompted me to write kickass provider.

@bardelot  Filter the search results to get TV Shows is a good idea but this type of filter does not work at all on kickass. An example is "Fringe", if you search this show a lot of good torrents are in other section like apps or movies. However the rss link seems indipendent from this type of filter.

@bardelot search by air-date can be implemented for sure, but i need to do more testing and option to download verified torrents only is a very good idea (probably a bit difficult to me provide this type of option)

@bardelot Whers is logging part referencing to EZRSS?.

bardelot:
Just search for "EZRSS" in the new provider and replace them with something like "kickasstorrent" otherwise you have two providers that both output the same logging / error messages.
The category filter does in fact work, but you're right the categories are often messed up. I don't know how you can search by air-date but I'm not that familiar with kickasstorrent at all.

mr-orange:
@bardelot thank you for pointing out the logging error. It' s fixed now.

akuiraz:
haven't tried this yet but i am definitely interested in seeing this provider integrated into sickbeard.  I do believe their URL has changed to http://www.kat.ph/ ... other than the actual address, I don't think anything else has changed?

PR #127, opened 2011-04-08
Initial SSL support 

msoute:
Adds ssl support for cherrypy.
Configurable via config -> general. requires a .crt and .key file (generated with openssl)


midgetspy:
Awesome, this is something that has been on my plate for a while, thanks. If you look in the SABnzbd code they have a method to auto-generate a cert if the OpenSSL library is installed, IMO that's required for this feature to be pulled since personally I have no idea how to make a cert if SB didn't do it for me ;-P

msoute:
Added auto generation for certificates. Certificates are auto generated of SSL is on but no certificate or key is provided in config. If SSL is on but config for certificates is incorrect SSL _should_ be disabled. This will need some testing on Windows as i can only test on *nix env.

Requires pyopenssl

msoute:
No reason and yes, will change / add those

midgetspy:
FYI in the future you should do your development in a branch so you can keep your master clean & synced.

midgetspy:
So this works without pyopenssl if you have your own certificates, correct?

smidley:
+1

msoute:
@thezoggy  removed it. Guess that if one knows how to create certificates then placing those in the correct dis isn't going to be a  problem.
Maybe i will add a text field later to make it possible to c'n'p the certificate information, should not be a problem since they should be in PEM format anyhow.

Also synced with origin/master.

midgetspy:
Closed cause I pulled the other similar request already.

PR #130, opened 2011-04-11
Allow Setting TVDB API Base URL in SB's config.ini

f0ff886f:
Allow setting the TVDB API Base URL in the config.ini (but do not expose it via gui as it should only be changed if an advanced user is changing things).

Previously I was always changing /lib/tvdb_api/tvdb_api.py but it breaks every time Sickbeard tries to update from Git when a change to tvdb_api.py was committed. This will allow me to use TVDB API hooks as described here: http://sickbeard.com/forums/viewtopic.php?f=8&t=487&start=20#p7324

This won't change the workflow for anyone who won't use the option, and abstracts the setting outside of hardcoded-inside-source.

### DIFF
--- a/lib/tvdb_api/tvdb_api.py
+++ b/lib/tvdb_api/tvdb_api.py
@@ -47,7 +47,7 @@ try:
     import lib.httplib2 as httplib2
 except ImportError:
     import httplib2
-    
+
 # Try using local version, followed by system, and none if neither are found
 try:
     import lib.socks as socks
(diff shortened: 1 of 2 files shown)


f0ff886f:
Hey midget, is there anything I can do to help this request along? I've pretty much stopped updating (as have a couple others) because it gets tiresome to do the merging each time.

If this won't be accepted just let me know, I'll think of a better way to change my TVDB hook (probably using the hosts file, so that the change is transparent to sickbeard)

Thermionix:
+1

MrAdam:
Looks like this was removed, any chance of adding it back to the config file?

Thermionix:
I have updated this pull request in; https://github.com/midgetspy/Sick-Beard/pull/562

MrAdam:
Thanks!

f0ff886f:
Closing this PR, looks like there is another one open #563 that also went stale.

PR #132, opened 2011-04-14
Fixed up Growl Notifications so that registrations are only done once (on

Technicolour:
Reasoning, when using Growl to forward notifications to your phone or even just using Growl things break when the notifications aren't set up properly. They also don't look very good (Growl for Windows) without an icon.

Fixed up Growl Notifications so that registrations are only done once (on Testing), also added icons.


midgetspy:
What if you never click the test button, they would just not work? If that's the case we should add some text to the config to indicate that it's required to send a test notification before it'll work.

Technicolour:
Yeah I was thinking that probably needed to be updated as well. Let me know if you think it requires anything more.

And yes you do need to register before any Growl notifications will work

Technicolour:
Is this suffice or should I go back and re work it a bit more?

alexw23:
Does not work.

Running Growl 1.6 on Mac OSX 10.7.3, the notifications come through fine but not the icon.

Console Output:
6/03/12 2:52:54.078 PM Growl: Not downloading icon for application SickBeard
6/03/12 2:52:54.213 PM Growl: Not downloading icon for notification Test of application Sick-Beard

Technicolour:
As per the GNTP spec passing a URL for the icon I would suggest that from the being in the spec and the logs you attached that it's either a Growl 1.6 problem or setting somewhere. It's also still working perfectly for me through Growl for Windows.

Technicolour:
*for the icon is allowed.

alexw23:
Sorry I'm running Growl 1.3.3

alexw23:
Here we go! https://github.com/guard/guard/issues/231#issuecomment-3695190 I will wait till 1.4!

PR #135, opened 2011-04-19
new coming ep banner/poster view w/new tooltip

thezoggy:
just ask if you really don't know whats all changed in this


thezoggy:
just re-based off current master 04/22/2011.

PR #137, opened 2011-04-21
My CherryPy 3.2 pull ;)

ghost:
It work with Python 2.7 installed via `Homebrew`.


thezoggy:
http://download.cherrypy.org/cherrypy/3.2.0/CherryPy-3.2.0-py2.win32.exe

http://www.cherrypy.org/wiki/WhatsNewIn32

ghost:
Yes? Why would we need the .exe file for the source?

Maybe for the Windows binary :)

thezoggy:
just was for midget/myself reference what was changed.

ghost:
OK. Thank you :)

midgetspy:
I will use 3.2 eventually but since it doesn't fix a single problem I'm aware of with SB I'm not planning to bother until I am looking for things to do or I find a reason I need it :0)

ghost:
Well, there is an almost two years span between your version, and the upstream 3.2 release. Which I truly believe contains many changes to the source, which could fix problems not know, as well as improve other areas where CherryPy is used, e.g. speed, etc.

But, I respect your opinion. Just trying to help out :)

midgetspy:
I appreciate it - I'm mostly just worried about causing new problems by pulling this version when there aren't any problems with the old one that I'm aware of. Trying to keep my workload at a minimum :0)

I'll pull it eventually, but in the meantime if you find any issues (fixed or caused) with 3.2 please let me know, thanks!

PR #139, opened 2011-04-29
Issue 886 & 389

kjoconnor:
Added updatetime to the TV object to allow for checking of last TVDB update time

Added Last Update Time to displayShow.tmpl

This will require running the following against sickbeard.db:
ALTER TABLE tv_shows ADD COLUMN updatetime NUMERIC;

This corresponds to Issue 886 here:
http://code.google.com/p/sickbeard/issues/detail?id=886
## 

It looks like GitHub put my other commit in with this request which addresses Issue 389 for iCalendar support:

http://code.google.com/p/sickbeard/issues/detail?id=389

The only problem is it might need a patched iCalendar library to work properly:

https://github.com/cozi/icalendar/pull/1


kjoconnor:
Is this not good?

kjoconnor:
I'll be splitting this into branches sorry, not too familiar with Git and branches and what not yet.

PR #140, opened 2011-05-02
Synology's NAS integration

zebulon501:
Provide the option to notify the media indexer of the Synology's NAS at the end of the post processing.


midgetspy:
What exactly does this synoindex do?

Diaoul:
Indexing multimedia content on Synology NAS. I think it is used to share medias through protocols like DLNA.

Hi zebulon, you here :)

zebulon501:
synoindex -a adds the file to the internal database of the UPnP MediaServer embedded into the NAS. Doing so avoid the long process of a full re-indexing of the file system. synoindex is a proprietary software from Synology, as its name suggests.

zebulon501:
@midgetspy: is there any decision about this feature? Is there any thing I need to change to get it pulled?

midgetspy:
Nope I've just been busy with the SB Index - I'll pull it soon.

midgetspy:
FYI in the future you should always work inside a branch and keep your master clean and synced.

zebulon501:
Thanks for the pull! and thanks for the git advice, was my first use of git, I'm used to more traditional VC (cvs, cvs, p4).

PR #141, opened 2011-05-03
Add sickbeard.db.dbFilename() method

jerub:
This will ensure that the mistake that was in mainDB.py a few times
where PROG_DIR and DATA_DIR were exchanged will not happen, if everyone
uses the new function to find out the location of the db.

This also means that upgrades will work from scratch properly if you use --datadir :)


jerub:
I amended the commit so that it doesn't have a NameError in os.remove(dbFilename)

PR #142, opened 2011-05-03
Issue 429 - Syndication

itsmegb:
Added a very basic Atom feed to the History page. Its not very pretty, but it does the job.


thezoggy:
this may not be needed anymore since you could just use the sb api to output history in json.. 
then just doing json -> atom..  ?
http://pastebin.com/raw.php?i=QQgLpnnG

itsmegb:
yea, i was going to close this anyway, it was never really implemented very well and i never used it myself so didn't bother looking back at it!

PR #143, opened 2011-05-04
Wrong line ending on first line of SickBeard.py

ghost:
cube> ./SickBeard.py
: No such file or directory

The new line ending fixed this problem.


midgetspy:
Thanks, but mixed line endings is gross and I am sticking with Windows line endings for the time being for various reasons. Just run it with "python SickBeard.py".

ghost:
In respect to your wish to not mix line endings you should review the current version of SickBeard.py. Only the first line has Windows line ending, the rest has Unix line endings.

PR #146, opened 2011-05-06
when adding a show first load episodes from thetvdb then load files from dir

lad1337:
as mentioned i dont see why the files should be parsed before we have info of them in our db
the only thing i can think of are the infos pulled from the nfo's but the tvdb lookup is done anyway


lad1337:
nm

PR #147, opened 2011-05-08
added SSL support for nzbmatrix

lgordon:
I changed http://www.nzbmatrix.com to https://nzbmatrix.com to support SSL.

I'm new to this so I hope I did this right.

Thanks,
Lyle

p.s. sickbeard is great.

### DIFF
--- a/sickbeard/providers/nzbmatrix.py
+++ b/sickbeard/providers/nzbmatrix.py
@@ -40,7 +40,7 @@ class NZBMatrixProvider(generic.NZBProvider):
 
         self.cache = NZBMatrixCache(self)
 
-        self.url = 'http://www.nzbmatrix.com/'
+        self.url = 'https://nzbmatrix.com/'
 
     def isEnabled(self):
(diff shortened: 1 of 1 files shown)


midgetspy:
Well that URL isn't used for anything except the link to the site. The actual url is rss.nzbmatrix.com and it doesn't support SSL (or didn't the last time I checked). Also FYI your commit changed every line because you committed with unix line endings rather than windows style.

lgordon:
Sorry about the line endings, I edited using the github editor since it was such a small change, I guess it altered the line endings to unix style.

re: SSL

The nzbmarix FAQ says (http://nzbmatrix.com/faq.php):

Do you provide SSL connection to the site?
- YES! Its not 100% completed yet, so you may have some "This page contains insecure items"

This is due to external images loading (banners and logos)

SSL Address
https://nzbmatrix.com

and the API info page says (http://nzbmatrix.com/api-info.php): SSL is supported, please use https://api.nzbmatrix.com instead.

PR #150, opened 2011-05-14
newzbin_ignoresource pull request

jaymzjulian:
Add support for ignoring newzbins quality setting in searches, since
it is often missing or incorrect for more obscure shows - the same as  the patch in issue #1515 where i failed to follow instructions :)


midgetspy:
Why not just set the quality to "Any"? Would be the same end result I think...

jaymzjulian:
because ps_rb_source is not set based on the quality setting, but hard coded to 3008, which seems to represent all sources - but will not return a result for which a source is not set, which is many of them. (is the correct fix to set this parameter based on the qualiy setting then?)

midgetspy:
Ah I see... _sigh_, yet another reason why newzbin is awful.

I think the solution here is to just remove the ps_rb_source restriction permanently, it should be covered in the quality stuff anyway.

jaymzjulian:
i agree with that solution - I only made it optional to try and not be contraversial :).  Assuming you don't require a new pull to delete the one line.

midgetspy:
Nope I committed the fix last night, thanks.

PR #151, opened 2011-05-15
IP Whitelist Modifications

bshep:
Added support for an ip whitelist to allow acces without authentication from certain networks.  Includes updated settings template and config file handling routines.

The code that disables basic_auth is very hacky but I couldnt find a way for cherrypy to honor a change in the tools configuration dynamically, if you know of a better way let me know!

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -632,7 +632,7 @@ class ConfigGeneral:
     @cherrypy.expose
     def saveGeneral(self, log_dir=None, web_port=None, web_log=None, web_ipv6=None,
                     launch_browser=None, web_username=None,
-                    web_password=None, version_notify=None):
+                    web_password=None, version_notify=None, web_ip_whitelist=None):
 
         results = []
 
@@ -666,6 +666,7 @@ class ConfigGeneral:
         sickbeard.WEB_LOG = web_log
         sickbeard.WEB_USERNAME = web_username
         sickbeard.WEB_PASSWORD = web_password
+        sickbeard.WEB_IP_WHITELIST = web_ip_whitelist
 
         config.change_VERSION_NOTIFY(version_notify)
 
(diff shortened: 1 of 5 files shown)


midgetspy:
This shouldn't be a UI config option IMO, should be an advanced (config.ini) option only. Other than that looks good, I haven't tested but I will give it a shot soon.

bshep:
Cool let me know if you want me to remove the UI config options or make other changes

midgetspy:
Yeah the config stuff will have to be removed before I pull, or I will do it eventually when I have a few minutes :0)

thezoggy:
does this have ipv6 support?

bshep:
I'm pretty sure it doesn't, but I dont have an ipv6 setup so I cant test.

It wouldn't be hard to implement though since the hard part was figuring out how to disable the authentication dynamically and that part wouldn't really change

sickbeard/webserve.py
```diff
@@ -666,6 +666,7 @@ class ConfigGeneral:
         sickbeard.WEB_LOG = web_log
         sickbeard.WEB_USERNAME = web_username
         sickbeard.WEB_PASSWORD = web_password
+        sickbeard.WEB_IP_WHITELIST = web_ip_whitelist
```
midgetspy: With no config UI this'll erase the whitelist on config save I believe.

bshep: You are correct.

removing the line now...

bshep:
Just noticed there are some changes in webserve.py that are in the current version but not in my version.  

I'm a little new to git so I'm not sure if I need to do anything for this patch to apply without reversing those changes or it git will just do the right thing.

Let me know if I should do anything on my side.

midgetspy:
Nope it will merge it automatically as long as you haven't changed the same
lines in which case I'll have to fix it manually.

bshep:
Found some more of the leftover UI config code and removed it, also reverted sickbeard/webserve.py since there should be no changes to that file.

Also added some documentation to the functions and moved some of the code around to make it look a bit better.

PR #154, opened 2011-05-16
support for torrent site freshon.tv (rss feeds)

pwr:
Hi,

I've added support for the torrent site http://freshon.tv (formely tv.torrents.ro).

Cheers,
-pwr


midgetspy:
How many torrents does this site have? How many users?

pwr:
No idea on the number of users (I don't have any connection with them, I'm just a user).
The site hosts torrets of TV shows only, and there are about 10k active torrents listed on the site, half of which are in HD.

PR #155, opened 2011-05-16
added all github relevant variables into config

SpLord:
for easy use of forks and branches, removed the hardcoded git repo and put it into the already existing version file


midgetspy:
Good idea, thanks.

SpLord:
added the exception to only show when the account is not midgetspy

ghost:
`git rebase -i HEAD~5`

And change the four at the top to `squash`. That will merge all your commits to one :)

SpLord:
done thx
@Mindtooth: i'm new to git, svn user :-)

ghost:
Hehe :)

We all were at one time or another. That is why the web is so great to share the knowledge.

Good luck on your future Git adventures!

sickbeard/version.py
```diff
@@ -1 +1,3 @@
+SICKBEARD_ACCOUNT = "midgetspy"
+SICKBEARD_REPOSITORY = "SICK-BEARD"
```
midgetspy, line 2: It occurs to me that version.py is actually auto-generated in the windows builds so I think we should just move these two variables... probably can just put them in **init**.py I guess.

jorgenpt, line 2: Shouldn't this be Sick-Beard and not SICK-BEARD?

jorgenpt:
I've implemented some related / overlapping functionality in https://github.com/midgetspy/Sick-Beard/pull/229
Specifically, it allows you to configure GITHUB_USER (or SICKBEARD_ACCOUNT as it's called in this patch) from the web interface, which decides where it pulls from. Ideally, SICKBEARD_VERSION would also be configurable from there :-)

Finally, I think the "right" approach to this for GitUpdateManager is to use git fetch instead of gh.commits.forBranch, and compare master to origin/master. Maybe a GithubUpdateManager that subclasses from it and provides the nice URL for the list of changes if "git remote -v" tells us the code is hosted on GitHub.com would be a good separation?

sickbeard/version.py
lad1337, line 2: +1 on both prev comments ... if moved to init.py make sure to change the refs in versionChecker/inc_top accordingly

midgetspy:
Closing this because it's not finished enough to pull and doesn't look like it's being worked on. If it is updated in the future open a new request please.

PR #156, opened 2011-05-16
language flags

SpLord:
for a better overview of the shows language added language flags to the show home site and the show site


midgetspy:
Can you squash this into one or two commits please? Other than that I think I am going to pull it but I will refactor it so that if all shows are english then no flag column is shown at all. I think this is the default case so no need to show a whole column of "English" flags most of the time.

SpLord:
Argh

midgetspy:
I can just cherry pick that commit, no worries.

thezoggy:
I brought up adding flags to midget on the language drop down and he mention this branch existed. While the things I had in mind are different than what you have done I have pretty much added everything you wanted to be done in this branch with the exclusion of the home screen modification. I put the flags on the Manage screen since I figure using the mass edit would be one place it would really benefit from. The language flags are on the displayShow / editShow / and also in the Adding show logic.. there is still yet a name map to be added so that instead of the 2 letter codes the user can actually see the english text instead (FR -> French). I'll talk to midget about possibly expanding this to the home screen. Here is my branch for more details and preview: https://github.com/midgetspy/Sick-Beard/pull/174

PR #157, opened 2011-05-17
added support for hard-linking files instead of copying them on posix systems

pwr:
Hi,

On POSIX systems (Linux/OS X), files can be hard-linked instead of copied, if they reside on the same filesystem.  This means there will be two file entries pointing to the same data block.
There are two major advantages to using hard-linking when possible: first, the operation is much faster (especially for big files), because only a new file entry is created, no data copy takes place; second, a lot less disk space will be used because both files share the same data.  This should help when handling many TV shows with HD episodes :).

I've added a method linkOrCopyFile to helpers.py, which attempts to create a hard-link instead of a file copy; if the operation fails (e.g. the source and destination are on different file systems), an exception is raised and it falls back on the regular copyFile.  I've also replaced all calls to copyFile with linkOrCopyFile in postProcessor.py, image_cache.py and helpers.py. 
I've tested it a bit, and it works as expected both when linking is possible and when not; however I only have a Linux machine to test it.

Cheers,
-pwr

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -398,12 +398,24 @@ def copyFile(srcFile, destFile):
         pass
 
 
+def linkOrCopyFile(srcFile, destFile):
+    if os.name == 'posix':
+        try:
+            ek.ek(os.link, srcFile, destFile)
+        except OSError:
+            # source and destination may be on different partitions, so hard-link is impossible
+            copyFile(srcFile, destFile)
+    else:
+        # link certainly not supported
+        copyFile(srcFile, destFile)
+
+
 def moveFile(srcFile, destFile):
     try:
         ek.ek(os.rename, srcFile, destFile)
         fixSetGroupID(destFile)
     except OSError:
-        copyFile(srcFile, destFile)
+        linkOrCopyFile(srcFile, destFile)
         ek.ek(os.unlink, srcFile)
 
 
--- a/sickbeard/image_cache.py
+++ b/sickbeard/image_cache.py
@@ -142,7 +142,7 @@ class ImageCache:
             ek.ek(os.makedirs, self._cache_dir())
 
         logger.log(u"Copying from "+image_path+" to "+dest_path)
-        helpers.copyFile(image_path, dest_path)
+        helpers.linkOrCopyFile(image_path, dest_path)
         
         return True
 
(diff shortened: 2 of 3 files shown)


thezoggy:
is your os really not that smart to know that your moving a file from point a to point b that it doesnt just moves the pointer?

pwr:
SickBeard _copies_ files, doesn't move them. Stop trolling.

Diaoul:
Well you can have SB move files instead of copying. I don't get it.

nberlee:
thezoggy isn't trolling!

Unless KEEP_PROCESSED_DIR is set it moves the file... Or when the destination is on different mounting point, the python move function will copy/delete...

[code]         # move the episode and associated files to the show dir
            if sickbeard.KEEP_PROCESSED_DIR:
                self._copy(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES)
            else:
                self._move(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES)
[/code]

pwr:
The files might be under some other program's management (e.g. torrent downloader), with its own policy of removing files (e.g. when reaching a certain share limit).  Moving the files from the source folder will certainly confuse it and maybe even cause it to re-download the files.
If the user chooses to move the files in SB, this will not impact in any way.
If the user chooses to copy them, hard-linking instead of copying (when possible) will certainly be faster and conserve disk space.

icovada:
He isn't trolling and I endorse pulling this into the main code.

If you use bittorrent to download the files and you want to keep seeding them, as of now the options offered by Sickbeard are 
1) Move them to your final folder, causing the bittorrent client to complain that the files are missing or, worse, redownload them
2) Copy them, taking up space with duplicates.

I hope this will go in the main code soon. In the meantime, I'll patch my own code myself

RichieB2B:
Excellent! Please add this to the main code. Works great on my QNAP NAS.

snown:
I too forked Sick-Beard to use hard links, but if it's pulled into the main code it will be one less thing for me to maintain.

+1

xcorp:
How about using a symlink instead, for us people that has a separate download disk.

fenduru:
:100:

PR #159, opened 2011-05-20
temporally set the locale to en_US so the date string we get from newzbin

lad1337:
temporally set the locale to en_US so the date string we get from newzbin can be parsed.
implemented as helper functions.

tested on mac osx with python 2.7 and a default locale "de_DE"
k thx by


lad1337:
got a better implementation with the use of http://labix.org/python-dateutil as date parser and not this hacky way

will be a new pull request

PR #160, opened 2011-05-22
Added initctl script for ubuntu

sj26:
Added an launchctl script for ubuntu which I greatly prefer to the antiquated BSD rc system.


sj26:
This is already in the repo from another contributor; closing.

PR #165, opened 2011-06-01
mod_proxy_html mangles the JSON that Sickbeard outputs

kruton:
CherryPy 3.2 includes a convenient decorator to indicate that an exposed
method should output JSON. Switch to using that instead of json.dumps
where possible to make it more clear what the intention of the method
is.

Unfortunately, the setStatus method has a parameter "direct" which
indicates whether it should output JSON or plain HTML. If this
definition is refactored, it can use the decorator as well.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -28,6 +28,7 @@ import datetime
 
 from Cheetah.Template import Template
 import cherrypy.lib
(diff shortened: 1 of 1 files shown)


kruton:
Note this is needed to use Sick-Beard through a proxy that inspects content like mod_proxy_html

PR #167, opened 2011-06-05
Removed https from Newzbin to allow sickbeard to download successfully

ozeraser:
Removed https from Newzbin to allow sickbeard to download successfully and not hammer the newzbin servers with requests.


midgetspy:
Yep, set autocrlf to false or input in your git settings to prevent git from automatically changing them.

ozeraser:
Thanks guys i will see if i can sort it and try again.

PR #173, opened 2011-06-25
quality tags, tableClick fixes, coming episode adjustments

thezoggy:
Added: css styling to make the show quality text on the home page look like a 'tag', plan to expand this cosmetic change to other sections later on.
Fixed: tableClick only affects left click now and will not kick off if shift is used. This should resolve those that want to shift click to open shows in new tabs.
Changed: Now the table td hover effect is purely css based rather than jquery adding/removing classes, should be more efficient and less prone to conflicts. Added class 'tvShow' for its identifier, also have tableClick use this class instead of a selector.
Changed: Made the verbiage on the coming episodes page mimic the displayShow page. Force -> Search. Also updated the non-list views to use the tvdb icon and search icon as well.

Preview:
http://zoggy.net/sb-ui-tweak1.png
http://zoggy.net/sb-ui-tweak2.png
http://zoggy.net/sb-ui-tweak3.png

### DIFF
--- a/data/css/comingEpisodes.css
+++ b/data/css/comingEpisodes.css
@@ -18,10 +18,10 @@
 .tvshowTitle a {
     color: #FFFFFF;
     float: left; 
-    padding-top: 5px; 
+    padding-top: 3px; 
     padding-left: 8px; 
-    line-height: 17px;
-    font-size: 16px;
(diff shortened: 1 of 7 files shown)


midgetspy:
Does the tableClick fix ctrl+click too?

PR #174, opened 2011-06-25
ui-flags -- Show language flags where appropriate 

thezoggy:
Changed: Replace the 2 letter language abbreviation codes with image of the language's flag instead.
Added: Added a column on the Manage page so that language flag is shown.
Added: Show the language flag on the show page as well in the edit show drop down.

Preview:
http://zoggy.net/sb-flag1.png
http://zoggy.net/sb-flag2.png
http://zoggy.net/sb-flag3.png


midgetspy:
Does the lang column still show up if all shows are the same language? If not I will add that before pulling

thezoggy:
lang column is only on the manage page and not on the HOME screen (i figured we could address that when you got back), i made the edit column slightly narrower to offset the addition so that the same amount of data is shown

Senaxx:
When this pull request is implemented this could also be used for another pull-request. The subliminal subtitle implementation. Currently it shows "srt" or the en,nl when multiple languages are downloaded. As I asked Diaoul to implement it in his Sickbeard fork he mentioned this pull-request. Subliminal integration could also profit from this pull-request.

thezoggy:
re-based, fixed conflicts. should be good now.

Diaoul:
As subliminal supports ISO-639-1 languages, it would be greate to have all flags for ISO-639-1 languages.
@thezoggy: do you think this is possible? where did you get those flags?

PR #175, opened 2011-06-27
Added logging of IP and username when basic_auth fails

bshep:
This adds information to the log about the IP address and username used when someone fails to login.

This clarifies the "Authentication error, check cherrypy log for more details" error that shows up when random people stumble upon your sickbeard install.

### DIFF
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -48,6 +48,7 @@ def initWebServer(options = {}):
             args = [status, message, traceback, version]
 
             logger.log(u"Authentication error, check cherrypy log for more details", logger.WARNING)
+            logger.log(u" - URL = %s" % str(cherrypy.request.path_info), logger.WARNING)
             logger.log(u" - IP = %s" % str(cherrypy.request.remote.ip), logger.WARNING)
             auth_header = cherrypy.request.headers.get('authorization')
             if auth_header:
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -42,13 +42,23 @@ def initWebServer(options = {}):
 
         #HTTP Errors
         def http_error_401_hander(status, message, traceback, version):
+            import binascii
+            import base64
             args = [status, message, traceback, version]
-            if int(status) == 401:
-                logger.log(u"Authentication error, check cherrypy log for more details", logger.WARNING)
+
+            logger.log(u"Authentication error, check cherrypy log for more details", logger.WARNING)
+            logger.log(u" - IP = %s" % str(cherrypy.request.remote.ip), logger.WARNING)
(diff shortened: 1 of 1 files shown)


mxhob1:
Would be interested in seeing this merged. I might fork it off and see if I can get it to temporarily ban access from IP's which fail x amount of times in y minutes (ie 5 failures in a minute) as my server is internet-facing.

baz1860:
I'm all for the increased info in the logs, but isn't the actual banning of IP x for length of time y something more suited to iptables or the like?

mxhob1:
@baz1860
 you're correct although to my knowledge (which is albeit limited) there is no way that you could separate failed authentication attempts from successful ones via iptables. If you implemented a flat x access attempts in y time from any IP through iptables, that would ban anyone who authenticated repeatedly, even they were successful authentications, although the scenarios for such a case would be quite limited.

Ideally, I'd have it monitor the failures as part of the authentication function in Sick Beard, and then if an IP were to violate it would just ignore all authentication attempts from that IP for x time afterward and forward to a generic 403 page.

baz1860:
@mxhob1 
-I INPUT  -p tcp -m tcp --dport 22 -m state --state NEW -m recent --set --name SSH--rsource
-I INPUT -p tcp -m tcp --dport 22 -m state --state NEW -m recent --update --seconds 180 --hitcount 4 --name SSH--rsource -j DROP
added to iptables would deny anyone making 4 failed SSH logins within 180 seconds be banned. I would imagine this could be adapted to stop the Sickbeard login. Change the dport figure to whatever you have sickbeard and I guess do something with the --name tag. I'm not going to pretend to be an expert on iptables, so you'll have to do some googling as to what that --name tag should be.

mxhob1:
@baz1860 
As noted in my last post, it's possible to limit the amount of attempted sessions per IP over a set timeframe, but iptables has no way to discern whether they are successful attempts, or failures. But it probably doesn't make much of a difference for SB though, since you would only ever realistically have one session open.

For anyone reading, make note to increase the hitcount argument to n+1, where n is the amount of times you want somebody to be allowed to authenticate. The reason for this is that if, for example, n equals 5, and you failed four requests. The fifth time you attempted to authenticate, you would be banned regardless of the outcome of your attempt since you would have reached the limit.

baz1860:
@mxhob1 
Fail2ban might be a better option for you then. It looks for failed logins and bans based on that information. I'd forgotten about it until a moment ago. Works very well on servers. I can't speak to how easy it would be to add the sickbeard log to the app - but I can confirm that it stopped the many many attempted ssh logins to the server I once had.

mxhob1:
@baz1860 
Ah! Perfect! Let's see if I can get it going with Sickbeard when I rebuild my server in the next couple of days :smiley:

PR #177, opened 2011-06-28
Made show titles bolder on the main page

camflan:
With the new quality badges, the front page is a little busy and the titles tend to get lost. I made them bolder so they stand out and take control of the row a bit better. I don't think it's the solution to stand time (I think the main page could use a re-org) but for now, I think it helps quite a bit.


midgetspy:
FYI in the future a screenshot is great for pull requests like this so I don't have to test it out locally to decide if I like it or not :0)

PR #179, opened 2011-06-30
Added iCal support

kjoconnor:
Added iCal support as per issue http://code.google.com/p/sickbeard/issues/detail?id=389

It makes use of PIL if it's available and needs the iCalendar library - http://pypi.python.org/pypi/icalendar/2.1 .  The library may need a patch (https://github.com/cozi/icalendar/pull/1) to work properly with Google Calendar though.

Tested on iPhone & Google Calendar.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -2552,7 +2552,7 @@ class WebInterface:
     # Provides a URL to generate an .ics/iCalendar subscribable URL to have an upcoming episode schedule
     @cherrypy.expose
     def ical(self):
-	from icalendar import Calendar, Event
+	from lib.icalendar import Calendar, Event
 	import random
 
 	logger.log(u"Receiving iCal request from " + str(cherrypy.request.remote.ip))
@@ -2571,7 +2571,7 @@ class WebInterface:
         
         # Create an iCalendar object with FixedOffset as air times are in EST
         cal = Calendar()
-        from icalendar import FixedOffset, UTC
+        from lib.icalendar import FixedOffset, UTC
 	# Create the EST tzinfo object
         EST = FixedOffset(-240, "EST")
         cal.add('version', '2.0')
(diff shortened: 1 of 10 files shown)


tester22:
Why not add the module to the library folder in SB, that reduces the need for the user to install modules to make it work "out of the box".

bjeanes:
I'd love to see this!

camflan:
Was going to do this myself, nice work. Pull this soon :)

smidley:
+1

kjoconnor:
I've resubmitted it twice already with no input from @midgetspy as to
why it's not being accepted. If I can get some guidance as to what's
wrong with it I'll fix and resubmit again.

On Jun 27, 2012, at 4:57, aanderson
reply@reply.github.com
wrote:

ghost:
+1 

Would love to see this added

sickbeard/webserve.py
```diff
@@ -2549,6 +2549,117 @@ def comingEpisodes(self, layout="None"):
 
         return _munge(t)
 
+    # Provides a URL to generate an .ics/iCalendar subscribable URL to have an upcoming episode schedule
+    @cherrypy.expose
+    def ical(self):
```
midgetspy, line 2554: The indentation in this function looks very screwy in this diff, is there a problem with tabs vs spaces?

sickbeard/webserve.py
```diff
@@ -2549,6 +2549,117 @@ def comingEpisodes(self, layout="None"):
...
+            	event['uid'] = datetime.date.today().isoformat() + "-" + str(random.randint(10000,99999)) + "@Sick-Beard"
+            	
+            	# Add the event to the Calendar object
+            	cal.add_component(event)
+        
+        # Pass the calendar string to the tmpl file
+        t = PageTemplate(file="ical.tmpl")
+        t.calendarString = cal.as_string()
+        
+        return _munge(t)
```
midgetspy, line 2661: Why not just return a string?

kjoconnor:
@midgetspy I redid this in #464

PR #182, opened 2011-07-01
Added TVTorrentz.org torrent RSS support

ghost:
Used the same layout as damoxc (thnx for that)

TVTorrentz has different naming rules opposed to other trackers, also the RSS feed does not include the filename. For singles this doesnt matter since the torrent name == scene release name, therefor it grabs the quality from the torrent URL. Been testing it for half a week now and it works flawlessly.

Let me know if you need more information or anything like that.

PS, Season packs will not work atm unless its a Scene released pack (then the .torrent name is still valid), when its a custom/internal season pack its usually named <showname> - <"season" + number> + <quality>.torrent


icovada:
TVTorrentz.org closed. I think we can unfortunately delete this pull request

ghost:
Yeah, im afraid you are right :-(

Maybe someday .... ;-) for now i closed the request.

PR #183, opened 2011-07-02
Ui enhancements - tooltip / tweaks / coming ep reorder

thezoggy:
Fixed the nowrap issue in Chrome for the airdate columns. 
Added quality 'tags' to the coming episode page.
Changed: Moved over to a much newer and nicer tooltip (qtip2) jquery plugin.
Changed: No longer use the info image as a placeholder for the ep summary tooltip.
No need to waste space for an image and padding, now we use the airdate or ep field where applicable.
Reordered the coming episode list view to be more logical as the most important columns should be first rather in the middle of the table.
Chrome related fixes galore.

Preview:
http://zoggy.net/sb-qtip-displayshow.png
http://zoggy.net/sb-qtip-comingep.png

*Not in the preview: Tweaked the css for the class that wraps around the tv show name in various places so that it now stands out a bit more against recent changes.

preview of css class:
http://zoggy.net/sb-manage4.png


thezoggy:
Re-based and updated module. Also included some chrome css fixes for some of the recent changes.

PR #186, opened 2011-07-09
Allow fullscreen mode for iOS when added as a webApp

neoflex:
Just added one line in the header to allow Sickbeard to be launch as a fullScreen webapp on IOS devices


neoflex:
sry, should have test more. It's not a good idea since anyway clicking on any link within the webapp will launch a safari window.

PR #187, opened 2011-07-10
Upgraded: Cherrypy 3.2.0rc1 -> 3.2.0

thezoggy:
Looks like everything works exactly the same and so far no errors. I would recommend having some of those people with jacked up import problems to test this out first before pushing out to master branch. 

Looking through all the code changes it's mainly documentation stuff. 
The JSONtools lib got a pretty decent rewrite but nothing is broken in sb because of it. 

There are a few new files that are part of this upgrade that I figure you may want to take a look at. 
One is cherrypy's module they have to help deal with multiple python version support and another is for logging/stats for the app. 

cherrpy_cpcompat.py
 -- Compatibility code for using CherryPy with various versions of Python

```
CherryPy 3.2 is compatible with Python versions 2.3+. This module provides a
useful abstraction over the differences between Python versions, sometimes by
preferring a newer idiom, sometimes an older one, and sometimes a custom one.
```

cherrypy\lib\cpstats.py
 -- A package for collecting and reporting on program statistics


jorgenpt:
And by now 3.2.2 is out :-)

thezoggy:
I purposely only updated to 3.2.0 as there was a pretty decent timeline gap..  the sab devs warn us of going to latest as it has ssl issues... this branch was mainly just for testing to see what/if anything that changed from our version.

3.2.0rc1 = 2009-10-17
3.2.0 = 2011-02-25

latest is:
3.2.2 = 2011-10-18

cant find their changelog for it... my hope was to get midget to test 3.2.0 final and then slowly over time make it up to 3.2.2... my fear is that bugs/issues will come up with such a giant leap forward.

jorgenpt:
Wow, didn't know there was such a huge difference between rc1 and final :)

That makes sense!

On Nov 22, 2011, at 7:32, thezoggy
reply@reply.github.com
wrote:

PR #191, opened 2011-07-19
Sab API url

mattijs:
The Sabnzbd API url is malformed. This causes snatched nzb to fail because they can't be send to SABnzbd. 
This diff poses some simple additions to fix the url. Maybe something more solid like a regex check would cover config errors like values with the http:// or the trailing slash in them


thezoggy:
Just set your sab url correctly. Please close this pull request.

PR #198, opened 2011-08-31
fedora and ubuntu init script modifications

chuckination:
Tweaking ubuntu and fedora init scripts to split out the data directory from the code directory.  Tweaked ubuntu init script to block until the daemon exits when stopping.


ghost:
Wouldn't `datadir=${SB_DATA-~/.sickbeard}` be a better idea?

PR #204, opened 2011-09-15
Added priority selecter for SABnzbd

adrianmoisey:
This allows you to set the priority of old (>7 days) and new episodes sent to sabnzbd. It may be useful for people who have more then just sickbeard pushing into SAB or for people like myself who run 2 sickbeards and want one of them to be a lower priority.

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -172,6 +172,49 @@
                                     <span class="component-desc">Category for downloads to go into (eg. TV)</span>
                                 </label>
                             </div>
+
+                            <div class="field-pair">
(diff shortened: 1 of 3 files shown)


adrianmoisey:
Is 2 a force?
No mention of it here: http://wiki.sabnzbd.org/api#toc36

adrianmoisey:
I've added it as an option

adrianmoisey:
Left to die I assume?

adrianmoisey:
Done

thezoggy:
if you want one to be lower you could just set the category you told sb to use.. to have a set priority rather than actually use the value passed from sb.

PR #207, opened 2011-09-25
Very small fix to handle occasional SSL errors that Prowl seems to generate from time-to-time

jiphex:
Occasionally after something downloads and Sick Beard tries to notify me, the Prowl API seems to generate some kind of an error related to SSL negotiation (which looks like a server-side issue).

All I've added is a try:except block to handle the errors produced when this happens, which should mean that SabNZBD+ is properly notified of the post-processing script completion, even if the notification failed.

Also, thanks a lot for Sick Beard. It's made a massive difference to how I watch TV.


jiphex:
Bah. Misnamed the error, closing this for now.

PR #210, opened 2011-10-03
Very small fix to handle occasional SSL errors that Prowl seems to generate from time-to-time

jiphex:
Occasionally after something downloads and Sick Beard tries to notify me, the Prowl API seems to generate some kind of an error related to SSL negotiation (which looks like a server-side issue).

All I've added is a try:except block to handle the errors produced when this happens, which should mean that SabNZBD+ is properly notified of the post-processing script completion, even if the notification failed.

Also, thanks a lot for Sick Beard. It's made a massive difference to how I watch TV.

(This was previously Pull request #207 - closed and reopened here)

### DIFF
--- a/sickbeard/notifiers/prowl.py
+++ b/sickbeard/notifiers/prowl.py
@@ -18,6 +18,7 @@
 
 from httplib import HTTPSConnection
 from urllib import urlencode
+from ssl import SSLError
 
 import sickbeard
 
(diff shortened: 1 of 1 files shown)


flz:
+1 for this request. Would be even better if it tried up to 3 times to deliver the notification as it seems to fail pretty frequently.

thezoggy:
cherrypy has ssl problems... why we are on an old version.. sab devs said they wish they never upgraded.

PR #215, opened 2011-10-11
Teeveeprovider

nls44:
Added a new provider that has pretty much exclusively scene content. Very fast after pre also, so should be a great resource.


thezoggy:
this honestly isnt needed as the whole purpose of that efnet channel is a request/fill channel to get things on usenet... and all of their content is already being indexed on any of the major providers... also those are some poorly named commits :P

nls44:
... What makes a provider distinctive is the speed content gets uploaded, the quality of the uploads and the speed of indexation. This provider scores big on these 3 points. First of all TV shows appear here faster than anywhere else on Usenet, they can be had 2-3 minutes after pre. Secondly, since it's only scene rips, the quality is great and naming is standardized, unlike the various crap you can find on Usenet. And finally the speed of indexation is amazing, .nzb's get created upon finishing the upload so it's available merely seconds after it's been uploaded.

*I agree on the naming of my commits though, but that's what you get when you commit at 4-6 AM :P

nls44:
Oh, also, it's not a request channel. Yes, it's possible to make requests there but its main purpose is automated uploading of TV episodes to Usenet.

lgordon:
I'm not sure if my vote counts for anything but there isn't very much overhead associated with another provider and teevee has good content so it would be a nice additions to sick-beard

akuiraz:
agreed... i like the idea of another provider.  i see no negative impact for including it.

janhuntjens:
+1, Would love to have another provider to fall back on. Seeing those 'your request limit has been reached' messages from my main provider in the logs scares me!

andrew-from-toronto:
Throwing my vote in for this as well. I don't see the harm in another provider, if people don't want to use it they can simply not enable it!

lgordon:
Any chance of this being pulled?

midgetspy:
My only concern is whether the site owners want the site to be automatically screen-scraped by SB. I imagine they won't be particularly enthused.

PR #218, opened 2011-10-16
Added PLEX Compatible multi-ep. "Extend." Renaming style.

chrisallen:
ADDED: Support for PLEX Compatible Multi-Episode Extend. (Note this only supports foo.s01e01-e02 & foo.S01E01-E02)
Added as option "3" to avoid issue with previous users settings.


chrisallen:
Have pushed corrected code. Thanks for spotting that. :-)

PR #224, opened 2011-11-05
If using SABnzbd, include a link to it in the top nav bar.

ericpeden:
Trivial change that fixes http://code.google.com/p/sickbeard/issues/detail?id=245


scottwf:
This link won't work for remote computers though will it? I've got sick beard and sabnzbd on the same computer so I have localhost listed in the SB config but I mostly access it from another computer so should I change the config to the external ip of the server?

ericpeden:
You're right, this patch won't handle that case. I'm not sure how to deal with that short of adding another config variable, and that seems like overkill.

yaleman:
@scottwf why not just change your config to the hostname of the machine that SB/Sab is hosted on? You shouldn't lose functionality...

PR #229, opened 2011-11-22
Configurable index page and gh user

jorgenpt:
This adds two general configuration options:
- Default page: Decides where the user is sent when he visits /
- Github fork: Decides which github user's fork the SourceUpdateManager pulls its updates from, for development purposes.

In addition it does some minor whitespace cleanup.

This resolves http://code.google.com/p/sickbeard/issues/detail?id=1756 and http://code.google.com/p/sickbeard/issues/detail?id=1757


lad1337:
i thought about this and why dont we just set the git repo url ? why limit it to different users but same repo name (and github)

sure changing it to a repo that does not merge / fit might break sb but same goes for "suddenly" switching he fork
in addition selecting the branch should be done too

i am not saying you should implement it just my idea of a optimal solution ... i might do

Diaoul:
Too limited IMO, it will break if I rename the repo.
No branch selection.

Why bother making a UI for that as it is moslty an advanced feature, it shouldn't be available to user.

I did similar stuff with a bit more customization and no UI here :
https://github.com/SickBeard-Team/SickBeard/commit/197a715436652ca039c13d4bb1b31b2ecafb5ce1

Diaoul:
Also, consider spliting your pull requests as there are many different features (4, 1 per commit) in the same pull request

jorgenpt:
@lad1337: Limiting to github is needed because we use the "github download tarball" feature to download an updated version - it won't work with non-github hosting. If you want to host it on something other than github, you should use the GitUpdateManager :-)
@Diaoul: Out of the 334 forks of Sick-Beard, only 3 have renamed the repository. Frankly, I don't see why you would rename it unless you were actually _forking_ the entire project, going in a different direction. I should probably split the configurable index-page & gh-user parts, but the first two commits are minor stylistic fixes that don't really have features of their own. I'd agree on not creating a UI preference for an "advanced" feature, but there's precedent already: Listening on IPv6, HTTP Logs, and Logging directory are definitely advanced options that most users never change. Ideally there'd be probably be a separate "advanced" configuration page, that had these options, but that's outside the scope of this change.

I agree that there should be branch selection, but that's something that can be easily added on top of this. :)

What's this "Sickbeard-Team" thing? There's no link to it on sickbeard.com.

Diaoul:
- Listening on IPv6 and HTTP Logs won't break SickBeard if misconfigured
- Logging directory isn't an advanced option as it creates file in a location

midgetspy:
Closed as per my comments on google code.

PR #237, opened 2011-11-28
Trakt notifier integration

daften:
This adds support for trakt.tv notifications. When an episode is downloaded, it is automatically added to the user's collection.


akuiraz:
I'm not sure I understand what exactly this will be doing... would it basically be sending Trakt.tv the Show Name and and Episode information to update the online library with the downloaded episode?

daften:
Basically: yes
On trakt you can add items to your "collection". When using sickbeard, you have to add it manually each time a new episode is downloaded. With this, it can be done automatically.

thezoggy:
cant you just use xbmc to do this? seems like a lot of people use sb + xbmc

daften:
Maybe, but I don't
I have a boxee box, so this is the best way to do this...

daften:
I've checked, and the trakt plugin for xbmc apparently supports 2 way sync even, so for xbmc users, that is the best solution. But not everybody uses xbmc. Hence also the support for e.g. NMJ ...
So this seems like a good solution (to me). I've tested it, and it works and I don't think I've missed any major items. :)

lad1337:
why is a simplejson update in here ?

akuiraz:
I can understand daften's point... the code has been written and it works.  Not to mention, I would imagine that now we are one step closer to integrating closer with Trakt.tv and Sick-Beard to allow use of the Trakt collection (or any list) for importing the users current TV Shows to Sick-Beard.  Another idea may be to make more sense of the "Archived" status.. I think that the idea of that status is say, you had the episode, you watched, and then you deleted it.  Is this correct?

daften:
@lad1337: simplejson update to allow for JSONDecodeError, which was not in the version in sickbeard original. This can be changed, but I don't think it's a good idea. The existing functionality keeps working with the update...
@akuiraz: Indeed, all good points, but I'd need more info to provide better integration :)

thezoggy:
is JSONDecodeError in json or only simplejson ?

daften:
I've tried json, and no luck. It is not mentioned on any of the main docs it seems.

thezoggy:
sorry looks like you only have just the name at that level.. not much is exposed

daften:
It certainly seems so. But as  you said, we still need to eliminate the potential for bugs. Maybe by limiting the number of splits to 3?
Can I add commits to this pull requests now? How, cause I can't find it in the interface here.

lad1337:
notifiers are called after the location is stored in the db
you could import the db and query for the file name you will get the show id and from there you can get the tv.show obj
i would consider this hacky and the proper way would be to expand the caller params but this is much more complex

you already added one commit ... just keep working on this branch

daften:
This is basically how I do it now, but I don't use the filename, because the complete path is stored. This seems like a faster query now.
And yes, it is hacky, but it's the only way I found how to do it. Maybe in the long wrong, the notifier framework will be changed so the episode and show objects can be passed :)

daften:
Changed the split, this should fix the potential issue.

thezoggy:
would your change still miss things like: Mongrels.2x05.720p_HDTV_x264-FoV

daften:
Yes, but can the ep_name be given like that?
I guess it can if the user changes the renaming settings.
So change to checking on location column in tv_episodes table. I'll get cracking :)
Thanks for the feedback so far, keep it coming

lad1337:
ok ... the non libary notifiers are called before it is saved to db ... look at 
https://github.com/midgetspy/Sick-Beard/blob/master/sickbeard/postProcessor.py#L702
this is before line the line ep_obj.saveToDB()

donno why but you might need to chang that or write your hook like the xbmc notifier does

daften:
Ah, ok, I will change my plugin to work as update_library and pass the necessary objects along. I didn't see this in the code before, thanks for the hint. :)
This will most likely be something for tomorrow, since it's 2 AM now and I have to work in the morning.

lad1337:
what are we talking about just do it like the synoindex_notifier notifier
https://github.com/midgetspy/Sick-Beard/blob/master/sickbeard/postProcessor.py#L715

2AM hey ... nice timezone

daften:
Yes, that was the plan
yeah, belgian time :)
Thanks for the comments, helps a lot.
One more question: now i use for statements to loop over the sqlResults objects i get back from the dbconnection (provided by sickbeard). Since I know only one result should be returned each time, what would be the cleaner approach. I don't think I can use Cursor from sqlite3 and then a fetchone, or can I?

lad1337:
you issue a sql query and get the row...
but you dont need that ... you just call the trakt notifier by hand and send you the ep_obj

daften:
Dammit, yes, it contains everything, including show information...
I'll blame it on the lateness and me being tired ;)

daften:
Ok, changed it with the risk of being a zombie at work (appropriate, since I'm watching the walking dead right now).
Tested it by manual post-processing an episode and it works.
Thanks for the feedback so far, let's get this baby merged as soon and especially as bug-free as possible :)

lgordon:
I use a netgear neotv (not xbmc) and the trakt integration directly with sb would be cool!

simonk83:
Yep, nice one daften, this'll be great :)

daften:
Thanks :)
Now we wait for the pull ;)

jamiew:
Excited for this, looking forward to mainline integration

daften:
Is there anything I can do to get this pulled quicker? :)
I see boxcar notifications have been pulled already, while being requested 2 days later.

lad1337:
we have to test the simplejson update since we make heave use of it in the upcoming api

daften:
Ah, ok, makes sense :)
Let me know if there's anything I can do to speed this up. I can do this without the simplejson update if necessary.

daften:
I see the api has been merged ... ;)

thezoggy:
yes, finally one thing off our plate before we can start looking at other pull requests

midgetspy:
The "Test" button doesn't ever return for me. I put username "a", password "b", apikey "c" and it just goes to a spinning ajax wheel forever. The log says:

trakt_notifier: Call method account/test/%API%
trakt_notifier: tries 1
trakt_notifier: Calling method http://api.trakt.tv/account/test/e, with data{"username": "a", "password": "7a81af3e591ac713f81ea1efe93dcf36157d8376"}

daften:
@midgetspy What system are you testing on?
On my system (mac os x), the method call fails and the spinning bar is replaced with an error message within 2 sec.

midgetspy:
Looks like the problem is your test method doesn't work on python 2.5:

The URL /home/testTrakt?api=e&username=a&password=o gives me:

  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy_cprequest.py", line 660, in respond
  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy\lib\encoding.py", line 193, in **call**
    self.body = self.oldhandler(_args, *_kwargs)
  File "C:\Users\Nic\workspace\SickBeard\src\cherrypy_cpdispatch.py", line 25, in **call**
    return self.callable(_self.args, *_self.kwargs)
  File "C:\Users\Nic\workspace\SickBeard\src\sickbeard\webserve.py", line 2013, in testTrakt
    result = notifiers.trakt_notifier.test_notify(api, username, password)
  File "C:\Users\Nic\workspace\SickBeard\src\sickbeard\notifiers\trakt.py", line 64, in test_notify
    return self._notifyTrakt(method, api, username, password, {})
  File "C:\Users\Nic\workspace\SickBeard\src\sickbeard\notifiers\trakt.py", line 98, in _notifyTrakt
    stream = urllib.urlopen("http://api.trakt.tv/" + method, encoded_data)
  File "c:\Python25\lib\urllib.py", line 84, in urlopen
    return opener.open(url, data)
  File "c:\Python25\lib\urllib.py", line 192, in open
    return getattr(self, name)(url, data)
  File "c:\Python25\lib\urllib.py", line 336, in open_http
    return self.http_error(url, fp, errcode, errmsg, headers, data)
  File "c:\Python25\lib\urllib.py", line 349, in http_error
    result = method(url, fp, errcode, errmsg, headers, data)
  File "c:\Python25\lib\urllib.py", line 659, in http_error_401
    errcode, errmsg, headers)
  File "c:\Python25\lib\urllib.py", line 357, in http_error_default
    raise IOError, ('http error', errcode, errmsg, headers)
TypeError: EnvironmentError expected at most 3 arguments, got 4

daften:
Isn't this a bug in python 2.5? (see http://mail.python.org/pipermail/patches/2006-December/021247.html)
I'll try to find a workaround though, would using urllib2 be an option?

daften:
I've just created a Virtual Machine with ubuntu, installed python 2.5 (see http://kovshenin.com/2010/12/installing-python-2-5-on-ubuntu-linux-10-10-2743/) and tried to reproduce, but it doesn't fail here. I'll change to urllib2, could you see if that fixes the problem?

midgetspy:
Merged.

midgetspy:
http://sickbeard.com/forums/viewtopic.php?f=4&t=3517

Broken from the trakt merge - can this notifier not use the json module that already exists in SB?

daften:
The problem lies indeed in the updated simplejson module (specifically the speedups file). I'll take a deeper look into it tomorrow.
In the meantime I've opened a pull request that should solve this problem.

PR #238, opened 2011-11-29
Do not process hidden directories on posix systems

rgerrits:
Fixes issue 1730: Sickbeard scanning .AppleDouble folders again and mistaking the 741 byte files for actual episodes


thezoggy:
is this needed anymore? as we recently pulled another request that looks like it did this very thing: https://github.com/midgetspy/Sick-Beard/pull/333

`Ignore .AppleDouble and .DS_Store when post-processing.`

rgerrits:
Ignoring .AppleDouble directories was the main reason for my pull request, so #333 will do.

PR #244, opened 2011-12-04
SickBeard API

thezoggy:


cpny:
Been using the api branch without problems + the extension that takes advantage of the api on sickbeard , looking forward to it :)

colinhumber:
As have I. No issue on my end either. Everything's been running very smooth.

PR #246, opened 2011-12-05
Update metadata of wdtv

mila76:
I have modified wdtv metadata creation make it useful (using last firmware) like issue 1761 http://code.google.com/p/sickbeard/issues/detail?id=1761

Missing some tag but a good starting point.

I'm not a python programmers and i don't know sickbeard sorce code so, for now, i dont know how add "title" and "series_name".
With this two all useful tags is done


mila76:
updated my code...  last one is bugged, commit wrong patched files.
now i have added title, series_name, genre and actor

mila76:
good, new patch is out

PR #253, opened 2011-12-11
Sending notifications when SickBeard update is available

Selissi:
Tested notifications for Notifo, Prowl, Growl, Boxcar and Twiiter.
Not tested Libnotify but if should work.
First commit kinda messy a lot of empty lines insert and deletion due to silly setting in EasyEclipse for Python.
Huge thank you to lad1337 who showed me the "right way" I did only machine copy/paste :)
Please consider it pulling it into master.

Selissi


simonk83:
Sounds good to me.  Could you add NotifyMyAndroid while you're at it?  

Perhaps it'd be worthwhile adding an auto update option as well, similar to CouchPotato v2.

PR #259, opened 2011-12-14
Bugfix for Linux where c speedups for simplejson don't work

daften:
On linux, the encode json data function didn't work due to some bug in the c code. I've disabled the speedups, which shouldn't make any difference in performance (little data is encoded)


thezoggy:
more simplejson issues.. http://www.sickbeard.com/forums/viewtopic.php?f=4&t=3529
i think we should revert the trakt.tv pulls until we can get everything figured out

thezoggy:
and http://www.sickbeard.com/forums/viewtopic.php?f=4&t=3517

daften:
I'll revert to the simplejson version from before the upgrade and change my code, give me about 2 hours :)

daften:
Pull request opened: #260
This should eliminate all the issues with json from trakt notifier

PR #262, opened 2011-12-19
Metadata Support for Synology NAS

BigCabbage:
Hi,

I added metadata support for Synology NAS, since I loved having Sickbeard on my NAS as it became available recently. I just missed correct metadata support for the build-in mediaserver.
Build it based on the mediabrowser one and used it for a few weeks. Now I decided to contribute back.
Hope you like it,

Frans.

### DIFF
--- a/sickbeard/metadata/__init__.py
+++ b/sickbeard/metadata/__init__.py
@@ -16,10 +16,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Sick Beard.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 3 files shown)


HolgerW1:
I agree with cytec, at least DSM 4.1 and the Video Station package don't use the meta data files that are currently created by Sick Beard.

It seems the Synology Video Station is looking for a JSON encoded file "SYNOVIDEO:TV_RECORD" in the directory "/@eaDir/<FILENAME>/" for TV entries.
The fields used inside the JSON are "channel", "title", "start" and "description".

Edit:
this meta data file is only used for TV records. TV episodes don't use any meta data.
See "/var/packages/VideoStation/target/plugins/syno_file_assets/episode.inc.php"

PR #265, opened 2011-12-29
Add search option for backlog search frequency

cwc:
Under Search Options, there's now a field for setting the number of days between backlog searches. Default is still 21, minimum is 4.

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -48,6 +48,17 @@
 
                         <div class="field-pair">
                             <label class="nocheck clearfix">
+                                <span class="component-title">Backlog Search Frequency</span>
+                                <input type="text" name="backlog_search_frequency" value="$sickbeard.BACKLOG_SEARCH_FREQUENCY" size="5" />
+                            </label>
+                            <label class="nocheck clearfix">
+                                <span class="component-title">&nbsp;</span>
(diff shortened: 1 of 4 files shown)


thezoggy:
this shouldn't be needed.. as anytime something is changed to wanted.. the backlog kicks off.. plus we've already had several providers harp about sb users hammering their sites.. this will only further increase the api restrictions / get sb banned.

owlboy:
thezoggy has a great point, we don't want SB becoming useless because we have no good providers.

thezoggy:
if you really think your backlog should run ever 4 days.. you either have sb misconfiguration.. invalid qualities for a show.. either case.. all you will end up doing is pissing off the providers. its not like THAT much content will be added in 4 days that will apply to most peoples backlogs. generally things get reposted after a tv show ends.. or when a new boxset comes out. if you didn't find it a few weeks ago.. chances are you really wont find it next week.. you are better off expanding your search providers to include p2p stuff (nzbmatrix/newsbin) or torrents.. if it really isn't found on the reliable providers that auto tag stuff (scene). searching every 4 days for 'Get Smart' in 720p bluray is never going to net you any results... and if your backlog is so huge that you cant wait 3 weeks before it starts eating away at it.. maybe you should get out of the house or start watching some of the crap that you already have.

midgetspy:
This was intentionally removed from SB and won't be pulled. There is almost no situation where this is a useful change.

PR #266, opened 2011-12-30
SB API updates

thezoggy:
New api command: Show.Pause - toggle the pause state of a show in sb.
Grouped the show.getbanner/poster/pause together as a sb api update (0.2). Rather than having people rely on a build number that may or may not be out of sync of master. In other-words, this should allow for coders to detect things a bit easier.


gugahoi:
Thanks Zoggy!

PR #267, opened 2011-12-31
Using numbers for the date format used in logging.

lueo:
I am using a Traditional Chinese Window 7 system. When I click on the "View Log" button on the web interface, it shows the following errors:
"UnicodeDecodeError: 'utf8' codec can't decode byte 0xa4 in position 0: invalid start byte ...."

After debugging, I found that the filehandler of logger.py use "%b-%d" as date format. In Asian country, "%b" shows like "一月", and it saves as local codepage while other messages are utf-8 string. Thus when webserver.py tried to read it back as utf-8, it just failed.

I also changed the date format of logger to default python style (only numbers), so that webserver.py could read it in a more universal way. Otherwise the "View Log" page will be blank since the regular expression can't match anything ("Jan-01" becomes "一月-01", then match failed).

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -86,7 +86,7 @@ class SBRotatingLogHandler(object):
     
         file_handler = logging.FileHandler(self.log_file)
         file_handler.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


midgetspy:
The right solution is to fix how the web UI reads in the log rather than changing the format of the log to be less readable.

lueo:
Which file in the Web UI to be modified? I can help to code the change.

PR #272, opened 2012-01-08
Added new separator stype compatible with default theRenamer settings

acasajus:
Current separators are
- -: Show Name - S01E02 - Ep Name
- space: Show Name - S01E02 - Ep Name

And I added
- period: Show Name.S01E02.Ep Name

theRenamer (www.therenamer.com) uses the last separators by default. It's a very used tool to rename episodes in windows.


midgetspy:
No naming enhancements are being accepted as a custom naming solution is on the way.

PR #273, opened 2012-01-10
Added changelog contents via tool tip to the Config page if applicable.

thezoggy:


thezoggy:
this is no longer needed as changelog moved to github wiki.

PR #274, opened 2012-01-16
FIX: Keep original file doesn't keep reprocessing and nofiying the same ...

acasajus:
...file again and again

Fixes http://code.google.com/p/sickbeard/issues/detail?id=1538 and http://sickbeard.com/forums/viewtopic.php?f=6&t=848

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -620,6 +620,10 @@ class PostProcessor(object):
         
         # set the status of the episodes
         for curEp in [ep_obj] + ep_obj.relatedEps:
+            splitStatus =  common.Quality.splitCompositeStatus( curEp.status )
+            if splitStatus[0] == common.DOWNLOADED:
+                self._log(u"Skipped setting status to SNATCHED. Episode already DOWNLOADED", logger.DEBUG )
+                continue
             curEp.status = common.Quality.compositeStatus(common.SNATCHED, new_ep_quality)
         
         # check for an existing file


thezoggy:
so you wanna add code to 'fix' people with bad configurations?

acasajus:
It happens to most people downloading with torrents. If you check the url's, people are solving this problem using custom scripts. 

What's the 'good' configuration that prevents this behaviour?

lad1337:
doing this helps:
http://code.google.com/p/sickbeard/issues/detail?id=1538#c1

does you fix still work with initial and archive ? question is does the initial download status contain downloaded if so this will break it

i would sugest blocking it on a config/gui lvl ... not allowing "scan folder" and "keep originals"

acasajus:
Reading the comment also helps

> I unchecked 'Keep Original Files' and the endless notifications stopped, but SB should be able to handle this scenario for the people who (for whatever reason) actually want to keep the original files.

For instance, to keep seeding the torrent after it has been downloaded and processed.

midgetspy:
Ideally you shouldn't be post processing the folder where your torrents are seeding from but you should have your torrent client copy completed downloads to a "finished" folder that SB can process.

acasajus:
The problem with doing that is that the torrent client is not "sickbeard exclusive" people use their torrent clients for more stuff than just sickbeard. If I do that then ALL the stuff the torrent client is downloading will be copied and only the stuff that sickbeard knows about will be moved somewhere else. Anything that sickbeard doesn't know about will be duplicated and I'll have to manually remove it. By post-processing the seeding folder I don't have to clean up manually the sickbeard post processing folder.

acasajus:
any more comments?

cgriffin4:
A fix to this problem would be extremely helpful.  It looks like my options at this point are to manually copy the files to a folder that SickBeard processes, have SickBeard post-process constantly, or not seed.  None of these options are considerably good.

acasajus:
Just get the changes from my branch

cgriffin4:
@acasajus Thanks, that would work.  If the official release isn't patched I will be forced to do that, I was hoping to avoid it since I'm not familiar with the process of compiling into a package my Synology NAS will install.

midgetspy:
This would mean that replacing episodes with higher qualities would no longer work, so I can't merge this.

acasajus:
With a minor modification I think I can also check if the quality is the same or worse. Would you accept it then?

PR #277, opened 2012-01-24
Changed the newzbin provider url to use newzbin.com

cianmcgovern:
I've changed the urls in the newzbin provider back to the newzbin.com url.
I expect newzbin will release an update on whether they plan on staying with the newzbin.com domain soon.

I've also included a commit that changes the launch parameters in the init.ubuntu script as per issue: https://code.google.com/p/sickbeard/issues/detail?id=1778

### DIFF
--- a/init.ubuntu
+++ b/init.ubuntu
@@ -34,7 +34,7 @@ RUN_AS=SICKBEARD_USER
 DATA_DIR=~/.sickbeard
 
 # startup args
-DAEMON_OPTS=" SickBeard.py -q --daemon --pidfile=${PID_FILE} --datadir=${DATA_DIR}"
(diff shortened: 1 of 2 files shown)


thezoggy:
http://www.sickbeard.com/forums/viewtopic.php?f=4&t=3716

midgetspy:
Newzbin.com is going away, the change stays. If you break the non-newzbin change out I can do it separately.

Diaoul:
Cherry pick !

cianmcgovern:
I put the pull request in when the newzbin2.es domain would no longer work before they switched to it fully yesterday. Hardly cherry picking...

PR #278, opened 2012-01-24
sabnzbd poller

lad1337:
it will poll sabs history every 30(hardcoded) seconds 10(hardcoded) items at a time. the sabToSickbeard script is not needed any more (but still working, both should NOT be used at a time could lead to a double pp)

it will filter out everything that has not the configured sab category
to stop post processing the same file over and over again it will compare the path of the "new" file with the post precessed files in the history
this means it will use the database every 30 seconds (bad for hdd spin downs)
with an advance option (in the config.ini only) it can be disabled (default is ON !)

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -32,7 +32,7 @@ from threading import Lock
 from sickbeard import providers, metadata
 from providers import ezrss, tvtorrents, nzbs_org, nzbmatrix, nzbsrus, newznab, womble, newzbin
 
-from sickbeard import searchCurrent, searchBacklog, showUpdater, versionChecker, properFinder, autoPostProcesser
+from sickbeard import searchCurrent, searchBacklog, showUpdater, versionChecker, properFinder, autoPostProcesser, sabPoller
 from sickbeard import helpers, db, exceptions, show_queue, search_queue, scheduler
 from sickbeard import logger
 
@@ -70,6 +70,7 @@ showQueueScheduler = None
 searchQueueScheduler = None
 properFinderScheduler = None
 autoPostProcesserScheduler = None
+sabPollScheduler = None
 
 showList = None
 loadingShowList = None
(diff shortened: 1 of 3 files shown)


nifoc:
What's wrong with using sabToSickbeard?

I sometimes have stuff in the "tv" category that I _don't_ want SB to process, which right now is easily done by removing the post-processing option for that particular item. Your poller would make that impossible, right?

lad1337:
yes you would need a dedicated sickbeard category
i could extend it to match the nzb name in sickbeards history too but that would prevent you from adding stuff from the outside that you still want sickbeard to pp ... this can also be an option
then the sab poller would only pickup stuff that sickbeard snatched

why ? because the most complicated setup is the sabtosickbeard script and its the most common error / problem ppl have and this would only need the initial connection to sab and you are done ... (and you dont have to enter your sb pw in a text file)

and besides you can still use the sabtosickbeard scipt (you should just disable the poller ^^)

thezoggy:
would sab poller work with web_root (reverse proxy peeps)?

lad1337:
web_root: yes i dont see an issue with that
i should mention that it only works if sb and sab are on the same machine this restriction is given because the path, to a file, is handed to sb as sab sees it therefor the path might not be valid for the box sb runs on
but i think the sabtosickbeard script has the same restriction

ps the poller is also aware of the nzb name and sends this to the post processor this will help identify the file even if it was named differently then the nzb

TFenby:
#445

TFenby:
Is this going to get pulled any time soon?

joehillen:
Is there a way to make it detect and warn if the sabtosickbeard script is enabled, or maybe even auto-disable it? 

I suspect this would be a breaking change for a lot of people and a lot of guides would have to be rewritten.

TFenby:
The toSickBeard.py scripts use the same interface as the manual post-processing page, so I can't see it being auto-detectable. The poller can be disabled by default, with the notice to disable the script if you enable it.

This is such a better solution, though.

joehillen:
Yeah, it would be a shame to have to disable it by default.

TFenby:
Check this out: http://wiki.sabnzbd.org/api#toc8

You can see which scripts are configured. That could be matched against sabToSickBeard.py. Yes, someone could have renamed the file, but doing it this way would be pretty safe, I'd think.

Or, to be 100% safe, you could check if there's _any_ scripts enabled. Then perhaps show a notification of some sort to let the person know?

thezoggy:
http://thezoggy.github.com/sabnzbd/api/#get_scripts

getting the scripts is one thing, but really you need to see if the script is set for the category.. awhile back we talked about just setting the script manually from sb via sab's api. but honestly you really dont need to even do that if you have intelligence to query sab and check the history for the nzo_id.. which you get back when adding a nzb (if you use json/xml -- we currently use plain text.. this needs to be changed). anyways just so many minor things and on the back burner.

thezoggy:
note that my api doc is far from complete or even current.. just one of the zillion side projects i've yet to finish :(

TFenby:
So you want to upgrade to the JSON API and use the nzo_id to identify the nzb during post-processing?

thezoggy:
ill have to go dig up my notes to give details but the whole concept should be..

if using sab, when adding a file we should store the nzo_id. that way we can check in on it later (sabpoller) to see the status. rather than doing some clunky thing like searching history. then one could easily nuke that failed download / retry / etc

and then once we go to pp the file, we can nuke the nzo_id from sb as it's no longer needed.

now where/how to store the nzo_id... maybe have some table of active_downloads.. that stores the name/provider/time/retry count/nzo_id/etc that you could use with the thing that processes the logic on handling failed downloads. or you could go a little lower in complexity and just append this nzo_id to the history table and go off that..

TFenby:
Is there a more appropriate place to discuss this? I'm thinking about just adding a new table for tracking nzb (and probably torrent) info, rather than spamming up the history db.
I don't want to start a huge discussion here unrelated to the patch, though.

thezoggy:
jump on irc.freenode #sickbeard sometime

PR #279, opened 2012-01-26
Cleans up stale pid files in the init.ubuntu script

tecnobrat:
start-stop-daemon's behavior is to check for the existence of a pid file, but doesn't ensure the pid listed inside is running.  Instead it checks that the pid file exists, and if the exec command exists in the process-list.  Since the python command isn't unique enough (/usr/bin/python), it thinks its still running.

This simply cleans up pid files which don't have a live pid in them, causing the rest of the behavior to work properly.

### DIFF
--- a/init.ubuntu
+++ b/init.ubuntu
@@ -52,6 +52,13 @@ if [ ! -d $DATA_DIR ]; then
     chown $RUN_AS $DATA_DIR
 fi
 
+if [ -e $PID_FILE ]; then
+  PID=`cat $PID_FILE`
+  if ! kill -0 $PID > /dev/null 2>&1; then
+    echo "Removing stale $PID_FILE"
+    rm $PID_FILE
+  fi
+fi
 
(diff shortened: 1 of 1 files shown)


midgetspy:
Was this addressed with https://github.com/midgetspy/Sick-Beard/pull/282?

tecnobrat:
Different solution to the same problem.

The other solution removes the pid file no matter what.  I remove the pid file only if its no longer running.

That solution has a major bug, if you call start when its already running, it will delete the pid file and then bail because it can't bind to the port.

Because the pid is now removed, stop will no longer work.

Mine, it will simply bail out because its running already and leave the old pid file intact.

blobby:
With this latest script, i get stale pid files lingering on reboot that means that SickBeard wont start up without manual intervention anymore :(

midgetspy:
I believe the pid files have been sorted, correct me if I'm wrong.

PR #281, opened 2012-01-26
testSABnzbd appends host with trailing slash if required

cianmcgovern:
I noticed a small bug when I was setting up sick beard for the first time. When I put the SABnzbd url into the configuration and then clicked TestSABnzbd, the test would fail because the URL didn't have a trailing slash. When the config is saved, a trailing slash gets appended to the URL if it doesn't have one. However, because  TestSABnzbd used the details from the URL fields before they were saved, it didn't append the URL with a trailing slash thus causing the test to fail. I've added a simple if statement to counteract this issue.


lad1337:
+1

PR #285, opened 2012-01-28
Tablesorter related upgrade and tweaks

thezoggy:
Upgraded jquery.tablesorter-2.0.3 -> 2.0.25.2.
With the upgrade came a bunch of new features, parsers, and widgets that we could leverage to improve things.
- Replaced our date parser in lieu of the isoDate parser built in.
- When sorting images, the sorting routine uses the image's alt tag rather than the url now... much better results.
- Updated our quality parser to work with the quality tags (uses textExtraction).
- Added filter widget to tablesorter for the coming episode and history pages.
- Added re-sizable widget to the tablesorter on the home page to test for feedback.
- Added stickyheader widget to the tablesorter on the mass update (manage) page.


thezoggy:
will resubmit at a later date

PR #286, opened 2012-01-28
Added the nolaunch parameter to the solaris SMF (init.d) script as per issue #1778

ghuntley:


midgetspy:
This is now fixed in source by having nolaunch automatically be set when in daemon mode.

PR #290, opened 2012-02-04
Post Processing Symlink Functionality

jness:
This commit adds symlink as one of the options for Post Process.

With this functionality users torrenting can continue to seed after processing,
and without the draw back of keep two separate files on the filesystem 
(this was the only option with Keep Original Files).

Jeffrey-

### DIFF
--- a/data/interfaces/default/config_postProcessing.tmpl
+++ b/data/interfaces/default/config_postProcessing.tmpl
@@ -58,6 +58,18 @@
                                 <span class="component-desc">Keep original files after they've been processed?</span>
                             </label>
                         </div>
+                        
+                        <div class="field-pair">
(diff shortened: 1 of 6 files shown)


ThomasDK81:
Hardlinks was the solution. Using cp -val copies the dir and hardlinks the files.
If anybody is interrested we made this solution: https://github.com/AzzABTN/BTN-Sickbeard-Post-Processing

junalmeida:
I personally prefer the idea of @thomasdk81. Copy, delete, simlinking download to tvshows. But we have to take care of file system. Most of Unix filesistems are OK. At windows, only ntfs is capable of simlinks. I don't know if ntfs supports hard links.

junalmeida:
I will test these changes at my torrent branch.

PR #295, opened 2012-02-09
tivo configuration was is ignored

IsThisThingStillOn:
There was no check if tivo was configured, resulting in execute of the code every time postprocessing ran.


midgetspy:
I've fixed this independently, thanks.

PR #296, opened 2012-02-10
Ubuntu init script changes for unique groups and home dirs

billmoritz:
In my installation, the SickBeard user has a different group name than
the user name.  Also my Sickbeard user has a home directory that is not
in /home.


thezoggy:
this can be closed out, this was resolved in the ubuntu init rewrite a year ago. users define their overrides in the etc/defaults

PR #297, opened 2012-02-11
Add 1080i HD TV support as separate category

unikitty37:
I added support for a separate 1080i category, so that people wanting stuff that only seems to have been released in 1080i can get it without a lot of manual downloading.

The new category is treated as higher quality than 720p/1080p, which is subject to some debate to say the least (well, not in the case of 1080p). Generally, you'd leave it disabled unless you need it, though.

I've also added it to the ANY preset, but left it out of the HD preset due to the above issue, as some people will prefer 720p to 1080p.

(Thought: will the changing of the category values in the Quality class cause problems with existing databases — will they need to be flushed or updated?)


midgetspy:
Yes this will break any existing config settings, they would need to be migrated to the new values.

midgetspy:
Closing this since it would break all existing users' quality settings. The necessary bit shifts would need to be made to the quality field in all show and episode entries in the database and their quality settings in config.ini. If this is done just make a new pull request please.

PR #298, opened 2012-02-11
SSL Support Added

xombiemp:
You can use and existing cert and key or it will generate a self-signed cert and key if they don't exist in the specified path from the config.  I used SABnzb as a guide.
This allows for a configuration of http only, https only, or simultaneous http and https on different ports.


midgetspy:
Pulled, thanks.

PR #299, opened 2012-02-14
Add TV show options to the general config view, with the following options.

flex1986:
TV show update on startup, this works together with the --force option at startup. Making it optional in the config makes it easier for novice users.

TV show update time, the current update time is hardcoded to 3 o clock. Making it optional gives users that don't have sickbeard running 24 hour the ability to update the TV show content automatically.

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -299,7 +299,7 @@ def main():
         sickbeard.launchBrowser(startPort)
 
     # start an update if we're supposed to
-    if forceUpdate:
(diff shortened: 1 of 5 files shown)


flex1986:
This option is related to the following topics, and I picked it up because I also run into a lot of out of date show information because my laptop is not on at night.

http://sickbeard.com/forums/viewtopic.php?f=3&t=3836&p=17168
http://sickbeard.com/forums/viewtopic.php?f=4&t=611&p=14454

flex1986:
Closed, still needs some fixing

PR #300, opened 2012-02-14
Set file ownership to parent dir owner after move/copy.

nmorey:
I'm running Sickbeard on a Thecus NAS which is a tightly controlled environment and it is not easily possible to configure the sickbeard user (thus it's root).

Copied files are already chmod like their parents so I added a similar behaviour for owner and group.


nmorey:
Oh yes I missed copy2 in shutil.
I guess I just need to write a wrapper that chmod the file and then calls sickbeard script

PR #301, opened 2012-02-14
Redirect to listening port on reboot after config change

xombiemp:
This will redirect to the proper url after a reboot if the
config changed the port or protocol (http/https) that the server is
listening to.
Previously it would just time out.

This changes the datatype to jsonp for the ajax request to is_alive to allow for cross domain requests, which it is when you change the protocol or port.

If going from http to https or https on a different port, the script will wait 5 seconds after the shutdown and then automatically redirect to the new url.  This is because if the SSL certificate is self-signed, or otherwise untrusted, the ajax request will break.


xombiemp:
I'm going to submit a new pull request so that the commits are cleaned up.

PR #304, opened 2012-02-17
Redirect to listening port on reboot after config change

xombiemp:
This will redirect to the proper url after a reboot if the config changed the port or protocol (http/https) that the server is listening to.
Previously it would just time out.

This changes the datatype to jsonp for the ajax request to is_alive to allow for cross domain requests, which it is when you change the protocol or port.

If going from http to https or https on a different port, the script will wait 5 seconds after the shutdown and then automatically redirect to the new url. This is because if the SSL certificate is self-signed, or otherwise untrusted, the ajax request will fail silently.

### DIFF
--- a/data/interfaces/default/restart_bare.tmpl
+++ b/data/interfaces/default/restart_bare.tmpl
@@ -1,6 +1,10 @@
 <script type="text/javascript" charset="utf-8">
 <!--
 sbRoot = "$sbRoot";
+sbHttpPort = "$sbHttpPort";
+sbHttpsPort = "$sbHttpsPort";
+sbHttpsEnabled = "$sbHttpsEnabled";
+sbHost = "$sbHost";
 //-->
 </script>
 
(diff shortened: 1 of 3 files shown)


thezoggy:
why are you creating a new pull request each day?.. it creates a lot of email spam

xombiemp:
Sorry. I wanted to clean up the pull request so there weren't so many commits in it.  I'm trying to get a handle on git.  What's the best practice for updating my branch when there are new commits in the sickbeard master?  Should I be rebasing to stay up to date or just leave my branch alone and let whoever merges it in worry about it?

midgetspy:
I'm not sure if you rebase and force the refs to update whether the pull request will update accordingly or not. You can always merge/pull from my master to keep yours up to date but unless you need to resolve conflicts in your branch it's easier for both of us if you just let it get out of date and when I pull it the conflicts will auto-resolve or I can rebase it myself.

xombiemp:
Ok, sounds good.  I'll just leave it alone now. I got all the kinks worked out too so it should be good to go.

midgetspy:
This doesn't work in Chrome going from https -> http. It pops up a bar warning that there is insecure content and if you click "Load anyway" it just goes to the old https URL. Doing http -> https worked.

xombiemp:
I couldn't reproduce the error in Chrome on Mac or Windows, but I could in IE9.  The above commit fixed the problem for me, see if it fixes it in the Chrome that was giving you the error.

ghost:
Breaks compatibility with reverse-proxy setups.

Tried to redirect me to http://[:47193, so it's not ipv6 compatible either.

PR #307, opened 2012-02-20
Don't download higher quality if the episode file has already been deleted

Anthirian:
This pull request adds functionality to Sick Beard to allow users to enable the option to not download better qualities of episodes if the original file is no longer present on the system. This is useful for people who use Sick Beard to download episodes, watch them and then immediately delete them from their hard drives. The functionality is off by default and can be turned on via the Search Settings config page.

This functionality works for the backlog searches as well as the rss searches. It is not fully finished yet, as I am still in doubt as to what to do if one episode of a multi-episode or a season has been deleted. You probably wouldn't want to have to download an entire season if that means you get 2 or 3 episodes in higher quality, but the question is where to draw the line?

PS. The commits have gotten somewhat messy, as my other pull request's commits somehow seem to have merged into this one. I have reverted that request's files to the originals, but they still show up in the files changed tab. This request's commits start from d03414ed6b1480f57a0166866b35221eb32efb0c onwards.


rascalli:
WHy would you set a higher quality for archive if you will delete after watching ?

Anthirian:
The reason for this is that you sometimes watch an episode before a higher quality becomes available. Other times however, you may not have the time to watch a new downloaded episode straight away. Until you do, you still would want Sick Beard to download the higher quality episodes. This is not an issue if you never delete your episodes, but it is when you do, or when you have a limited plan with your usenet provider.

PR #311, opened 2012-02-27
Auto create show folder if missing during post-processing

lebedevcoder:
This allows users to configure Sick-Beard to automatically create show folders if they are missing during post-processing. There is also an option to have Sick-Beard skip the creation of the show folder when a new show is added. Finally, during Sick-Beard's file scan routine, if a show directory is missing, and auto-create is true, then Sick-Beard will behave as if the directory is there, but episodes and/or season directories are missing (ie, behave as if the episode files have been deleted, and update the database to reflect that).   

This is to address Issue 136: http://code.google.com/p/sickbeard/issues/detail?id=136 

This is especially helpful for users who use WDTV, Asus O!Play, or other media players that do not hide empty folders in their UI. 

This is my first time using git-hub, so please let me know if I did something wrong (or could have done something better).

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -154,6 +154,7 @@ TVTORRENTS_HASH = None
 
 TORRENT_DIR = None
 
+ADD_SHOWS_WO_DIR = None
 CREATE_MISSING_SHOW_DIRS = None
 RENAME_EPISODES = False
 PROCESS_AUTOMATICALLY = False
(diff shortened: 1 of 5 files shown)


stevenj:
This patch is awesome. After i sustained a catastrophic hdd failure it saved me countless hours manually recovering everything. (I still have to re-dl anything I lost, but at least I am still up to date). This needs to be in mainline so everyone can benefit.  I love the feature of not creating show directories on add also, so much cleaner when browsing.  The only thing that could make it better would be adding the options to the GUI, but that's a nit pick and it doesn't affect me at all.

stephenm00:
Would love to see this merged into the mainline!

midgetspy:
Thanks, this is merged into the development branch (where all new features go now) for testing.

kierr:
Can you confirm what the status is of this patch? I am using the development branch, but folders are no longer created if they do not exist.

New location ... does not exist

I have the create_missing_show_dirs = 1 and add_shows_wo_dir = 1 set in my config file.

thezoggy:
the show folders would be re-created during the post-processing routine. if your not pp a folder.. then no they wouldnt be created

PR #313, opened 2012-02-28
Fixed url redirection after restart/update

mikaelq:
The restart page will now redirect to the configure protocol (http/https) but keep the host/port/location fixing the issue for people running proxies to connect to SB.


midgetspy:
Based on the description, if the user changes their port then the restart won't work anymore?

Also there are a bunch of extra changes in this pull request, best practice is to make a new branch that contains only the changes you want me to pull.

mikaelq:
Thanks for tip on the branch. I resubmitted a new pull request, fixing the problem with the port.

PR #316, opened 2012-02-28
Fixed bad restart redirection for people using a proxy.

mikaelq:
I modified the js to only generate a url from the config if connecting to localhost, otherwise it will just refresh.


xombiemp:
This defeats the whole purpose of the auto redirect. Instead of actually fixing the problem with using a proxy you just made it so it only works with localhost and nothing else.

mikaelq:
I agree, this code is bad. But I think the auto redirect needs to be thought about again. If anyone is using a proxy to connect to sickbeard(so not using the sbHost hostname) on every restart they will be redirected to the sbHost which fails. Building the sbHost url from config variables seems like a bad idea, it should be done from the global vars.

PR #317, opened 2012-02-28
Don't download higher quality if the episode file has already been deleted

Anthirian:
This pull request adds functionality to Sick Beard to allow users to enable the option to not download better qualities of episodes if the original file is no longer present on the system. This is useful for people who use Sick Beard to download episodes, watch them and then immediately delete them from their hard drives. The functionality is off by default and can be turned on via the Search Settings config page.

The reason for this functionality is that you sometimes watch an episode before a higher quality becomes available. Other times however, you may not have the time to watch a new downloaded episode straight away. Until you do, you still would want Sick Beard to download the higher quality episodes. This is not an issue if you never delete your episodes, but it is when you do, or when you have a limited plan with your usenet provider.

This functionality works for the backlog searches as well as the RSS searches. It is not fully finished yet, as I am still in doubt as to what to do if one episode of a multi-episode or a season has been deleted. You probably wouldn't want to have to download an entire season if that means you get 2 or 3 episodes in higher quality, but the question is where to draw the line?

PS. This is a rewrite of pull request 307 (https://github.com/midgetspy/Sick-Beard/pull/307), which messed up the other branches since I modified stuff in the master branch.

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -36,6 +36,18 @@
                         </div>
                         
                         <div class="field-pair">
+                            <input type="checkbox" name="check_existence" id="check_existence" #if $sickbeard.CHECK_EXISTENCE == True then "checked=\"checked\"" else ""# />
+                            <label class="clearfix" for="check_existence">
+                                <span class="component-title">Check Files' Existence</span>
+                                <span class="component-desc">Check files' existence before downloading higher qualities?</span>
+                            </label>
+                            <label class="nocheck clearfix">
+                                <span class="component-title">&nbsp;</span>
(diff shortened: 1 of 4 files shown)


thezoggy:
couldnt you just not have an archive quality setting? then the ep will just remain in 'ignored' after its been deleted?

Anthirian:
I thought so too at first, but then only one of the qualities you
selected will get downloaded, instead of downloading higher qualities
until the highest has been downloaded like the archived qualities
allows.

Anthirian:
Any update on whether this could get pulled?

thezoggy:
should be based off the development branch not master.
should squash your commits for a cleaner rewrite of history. 
also in the last 8 months a bit of changes have happened.. you prob want to double check the custom_naming changes didnt break anything in it.

PR #319, opened 2012-03-01
Fix auto redirect for ipv6

xombiemp:
This fixes the auto redirect for IPv6.  It was tested with the latest
Safari, Firefox and Chrome on Mac and the latest IE, Firefox and Chrome
on Windows.

Others have mentioned that the redirect does not work with a reverse proxy in place.  If someone could give me instructions on how to set this up I could try to figure out how to fix the redirect for that too.


ghost:
My pull request has one way to solve the reverse proxy issue.

xombiemp:
Sweet, nice going.

PR #322, opened 2012-03-04
Add 1080i HD TV support

unikitty37:
I've added a migration for the database and config.ini, and everything appears to be working correctly.


midgetspy:
You also need to migrate episode statuses which can include qualities (downloaded/snatched). Also we would need to do a check that if somebody has ANY selected that we update their quality to include 1080i or else all their shows will magically turn into "Custom" quality since the preset has changed.

midgetspy:
So this only works on Newzbin, correct? Have you tried any other providers?

thezoggy:
tried this out tonight. nzb.su found some 1080i releases but nzbs.org/sbi/nzbmatrix would not find anything. also this pull broke the backlog overview page

thezoggy:
this is no longer needed since we added the the 1080p stuff awhile ago. please close

PR #326, opened 2012-03-07
Ignore failed downloads

TFenby:
So, here's a patch that tracks in the history database when a download fails. (Based on it starting with _FAILED_ during post-processing). Additionally, there's a setting on the post-processing page that determines whether or not failed downloads are deleted after being marked.

The initial database update takes foreeever, though. Can't do an update table when adding a primary key, unfortunately.

### DIFF
--- a/autoProcessTV/autoProcessTV.py
+++ b/autoProcessTV/autoProcessTV.py
@@ -41,8 +41,7 @@ class AuthURLOpener(urllib.FancyURLopener):
         return urllib.FancyURLopener.open(self, url)
 
 
-def processEpisode(dirName, nzbName=None):
-
+def processEpisode(dirName, nzbName=None, failed=False):
     config = ConfigParser.ConfigParser()
     configFilename = os.path.join(os.path.dirname(sys.argv[0]), "autoProcessTV.cfg")
     print "Loading config from", configFilename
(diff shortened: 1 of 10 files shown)


TFenby:
Almost ready

thezoggy:
you should refer to the custom_naming branch. we modified the db slightly and store original release name and sizes of release.

TFenby:
Would you rather I do that? This is functional but for a few specific cases, and this is the most popular and oldest issue open on Google Code. The sooner this features is added to master the better.

thezoggy:
I was just saying, that we already modify the db to include some info that you may find useful for your feature implementation. Also you can see how to handle db migrations and config upgrades. custom_naming is almost done and will be merged into master.. thus your current pull request would have to be rebased.

TFenby:
I'll check it out, thanks.

TFenby:
It's crashing with the following, so I'm just going to continue with my patch as it is.

  File "/root/git/Sick-Beard/sickbeard/db.py", line 159, in _processUpgrade
    instance.execute()
  File "/root/git/Sick-Beard/sickbeard/databases/mainDB.py", line 420, in execute
    if not int(cur_ep["file_size"]) and ek.ek(os.path.isfile, cur_ep["location"]):
  File "/root/git/Sick-Beard/sickbeard/encodingKludge.py", line 56, in ek
    result = func(*[x.encode(sickbeard.SYS_ENCODING) if type(x) in (str, unicode) else x for x in args])
UnicodeEncodeError: 'latin-1' codec can't encode character u'\u20ac' in position 131: ordinal not in range(256)

PR #328, opened 2012-03-09
Allow .mkv as extension for .dvdrip

unikitty37:
Sick Beard assumed .mkv files to be HD TV even if they were DVD rips.

I use MakeMKV for all my rips, which of course means I end up with everything importing as HD TV rather than SD DVD; I'm not sure adding .mkv to the list of extensions checked will have a negative impact on anyone, as the filename contains .dvdrip anyway. (I assume .bdrip.mkv should still be counted as SD DVD quality?)

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -126,7 +126,7 @@ class Quality:
 
         if checkName(["(pdtv|hdtv|dsr).(xvid|x264)"], all) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDTV
-        elif checkName(["(dvdrip|bdrip)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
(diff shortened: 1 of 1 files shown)


thezoggy:
mkv = hd container. use avi or mp4 for sd.

thezoggy:
also... dvdrip can only be sd. dvd = sd.

unikitty37:
Why use avi or mp4 when MakeMKV rips straight to mkv? It's stupid to spend time remuxing everything…

This patch isn't intended to make things comply with scene rules; it's intended to make things useful for people ripping their own discs. How exactly does it hurt anything to have .dvdrip.mkv treated as SD DVD?

midgetspy:
Container (extension) has nothing to do with how it detects quality except as a last resort. Just name them dvdrip.x264.mkv or dvdrip.xvid.mkv or whatever you want.

PR #330, opened 2012-03-11
Issue 1541: Multi shift-select checkboxes

bramp:
http://code.google.com/p/sickbeard/issues/detail?id=1541

These two commits allow multiple checkboxes to be selected in a single go by clicking the first box, then shift clicking the second box. All checkboxes between the two will either be selected or unselected.


midgetspy:
Nice, thanks.

PR #332, opened 2012-03-13
Updated searching for new shows to also query each word individually

dstjacques:
For a lot of shows and movies it's difficult to know which format of the title the TV database is using. Modified the searchTVDBForShowName() function to also query each word in the search field individually. The concatenation of all search words is the first thing searched to keep the most relevant results appearing first. Results for each individual word appear afterwards as secondary results.

As an example searching for "law and order svu" would now search for "law", "and", "order", and "svu" and the results would include the desired show "Law & Order: Special Victims Unit" for which the spelling was unknown.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -26,6 +26,7 @@ import re
 import threading
 import datetime
 import random
+from sets import Set
 
 from Cheetah.Template import Template
 import cherrypy.lib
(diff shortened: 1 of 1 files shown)


midgetspy:
Can you give any examples of shows which were hard to find before but are found more easily now?

dstjacques:
The first time I had tried searching for "Dragon Den" instead of "Dragons Den".

Later I tried finding "Law & Order SVU" but couldn't. I find that this helps a lot with finding shows/movies that have acronyms, symbols like "&", and other non obvious spellings.

PR #334, opened 2012-03-14
Notification to execute command on snatch

xtsinc:
Added category 'Other' and included a notification to execute a command when a .torrent file finishes downloading, ie is snatched, not when the download itself is finished.

In my case, the .torrent file is downloaded to the local machine, the notify_snatch fires and executes an rsync to push the .torrent files to a remote seedbox server's watch directory.  The files are downloaded on the seedbox and sent automatically to the Sickbeard server's incoming download folder where they are then post_processed and any extra scripts are run.  Without the new notify_snatched to run the rsync, no .torrent files would make it to the seedbox to be downloaded and the post_processing extra scripts would not run (unless a manual download was completed).

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -805,7 +805,41 @@
                     </fieldset>
                 </div><!-- /trakt component-group //-->
 
+				<br />
+                <h1>Other</h1>
+                <br />
+				              <div id="core-component-group11" class="component-group clearfix">
+                    <div class="component-group-desc">
+                        <h3>Execute Command</h3>
+                        <p>Execute a command when the download finishes.  Useful for rsync of .torrent files to a remote seedbox</p>
+                    </div>
 
+                    <fieldset class="component-group-list">
+                        <div class="field-pair">
+                            <input type="checkbox" class="enabler" name="use_commandnotify" id="use_commandnotify" #if $sickbeard.USE_COMMANDNOTIFY then "checked=\"checked\"" else ""# /> 
+                            <label class="clearfix" for="use_commandnotify">
+                                <span class="component-title">Enable</span>
+                                <span class="component-desc">Should Sick Beard execute a command notification?<br /><br />
+                                </span>
+                            </label>
+                        </div>
+                    <div id="content_use_commandnotify">
+                        <div class="field-pair">
+                            <label class="nocheck clearfix">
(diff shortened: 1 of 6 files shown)


midgetspy:
Ah - this feature isn't appropriate for inclusion in SB master, sorry. This is the only application I can think of for such a feature (and is a very uncommon case) and IMO should be accomplished by something OTHER than SB (watch script).

If it was implemented as a hidden option like extra_scripts I would consider including it, if you feel like rewriting it. But honestly it would be WAY simpler to just make yourself a watch script to watch your torrent dir and act accordingly when it finds a .torrent file.

xtsinc:
I must respectfully disagree that this case is very uncommon.  As more private trackers are being added as providers, this feature will be used more and more since many users of these tracker communities utilize seedboxes to maintain the integrity and speed of the tracker as well as their required hit & run/ratios.

Unfortunately a watch script to accomplish this would be a poll/time based solution whereas this allows for actions to be executed exactly when needed and no other time, eliminating waste and providing the most efficient solution.  Since this is only needed because torrents are being actively downloaded by SB, I feel like the proper place to address the issue would be in SB.  I could re-implement this as a hidden notification of sorts.  Perhaps when the notifications are fired, the last one to be fired could be one that looks for extra_snatched_scripts and executes that.  Would that be acceptable? Would you like an option in Notifications to turn this functionality on?  Any other suggestions?

Thanks for the feedback.

midgetspy:
Well the number of SB torrent users is low, and the number of those users using seedboxes is even lower. Even if this is needed for every seedbox user I believe it's an uncommon use case.

A watch script would be polling, yes, but the performance of a script in a 1s sleep loop is so negligible that the waste/efficiency argument is purely theoretical in my mind. After all your torrent client is polling too and you seem content to use rsync to provide the file to it :0)

But all that said, if you write this as an advanced option I will include it. I wouldn't make it an actual notifier, just add a call to your function in search.py after it writes to history and sends the snatch notifications. The function can be basically a copy of the way the existing extra_scripts works (in fact you can feel free to refactor extra_scripts from postProcess.py so you can use common code in both places and just pass them in a list of scripts to run).

xtsinc:
Sounds good.  I haven't had a chance to dig into how the current extra_scripts implementation is handled, so your outline helps.  I'll see what I can code up and submit a new pull when everything is in order.

PR #337, opened 2012-03-18
Enhancment: Added Edit/Submit buttons to top of mass update page.

bungiman:
Makes it slightly easier to manage if the user has two places.

### DIFF
--- a/data/interfaces/default/manage.tmpl
+++ b/data/interfaces/default/manage.tmpl
@@ -64,6 +64,12 @@
 
 <form name="massUpdateForm" method="post" action="massUpdate">
 
+<table id="preMassUpdateTable" class="sickbeardTable tablesorter" cellspacing="1" border="0" cellpadding="0">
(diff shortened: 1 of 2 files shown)


bungiman:
Enhancement Screenshot

bungiman:
So I think that this enhancment broke the bottom buttons. I am going to take a look. Happy to listen to any advice.

EDIT: Fixed!

thezoggy:
that looks kinda funny.. what about putting them under the header? (aka 1st row). anyways.. why is this even needed? cant scroll to bottom (end key) ?

bungiman:
For me I was sorting by something where all the shows of relevance were visible. Since I was mostly using the mouse to select my actions I found it was easier when I had access to the buttons at the top. In fact I kinda assumed they would be with the normal feel of SickBeard. I can certainly add it in as a row. Just wanted something put out.

PR #339, opened 2012-03-21
Converted cookie references over to localStorage (html5).

thezoggy:
This saves HTTP requests and keeps the scope of the data local to the machine.
DOM storage has been around for awhile and basically any browser EXCEPT IE6/7 is supported (oh no!).
This means we could retire the use of the cookie/cookiejar/JSON js libs.. _However, I've left them alone for now._

### DIFF
--- a/data/js/browser.js
+++ b/data/js/browser.js
@@ -137,9 +137,13 @@
                 };
         }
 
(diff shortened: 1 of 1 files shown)


lad1337:
+1

PR #341, opened 2012-03-24
Header is always visible

enriquemorenotent:
Sometimes for long running tv shows, the page can get too long. I thought its a good idea to have the menu always visible :)

### DIFF
--- a/data/css/default.css
+++ b/data/css/default.css
@@ -24,6 +24,16 @@ padding:0;
 #upgrade-notification div{background-color:#c6b695;border-bottom:1px solid #af986b;padding:7px 0;}
 #header-fix{*margin-bottom: -31px; /* IE fix */height:21px;padding:0;} 
 
+.mainheader{
+    position: fixed;
+    width: 100%;
+    z-index: 1;
+}
+#contentWrapper{
+	position: relative;
+	top: 140px;
+}
+
 #header {
 background-color:#fff;
 padding: 5px 0;
--- a/data/interfaces/default/inc_top.tmpl
+++ b/data/interfaces/default/inc_top.tmpl
@@ -141,6 +141,7 @@ table.tablesorter thead tr .headerSortDown { background-image: url("$sbRoot/imag
 </head>
 
 <body>
+<div class="mainheader">
 #if $sickbeard.NEWEST_VERSION_STRING:
 <div id="upgrade-notification">
     <div>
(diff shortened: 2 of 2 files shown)


thezoggy:
i do not like this. this defeats the purpose of using stickyHeaders in the table or the fact we have a 'jump to top' in the bottom right when the menu is not visible. plz do NOT pull this

enriquemorenotent:
Agreed, with this the use of the top button becomes irrelevant.

What is stickyHeaders? I didnt see this option in the configuration screen.

bungiman:
Can you post a screenshot?

thezoggy:
http://zoggy.net/sb-upgrades3.png

bungiman:
I actually like this. What if you add another row above it for edit/submit buttons too? Then the user would have access to submission always too,

enriquemorenotent:
That screenshot is not really what it should do... weird. Can you post a full-page screenshot. so I can check what is going on?

thezoggy:
the screenshot shows what the stickyheaders does.. why i think your pull request and this current feature would be a train wreck.

bungiman:
Sorry, I think I got confused. thezoggy, the SS you posted was of stickyheaders? How is that enabled? 

If what I saw was _not_ added functionality from this pull request, then I think this should be closed.

enriquemorenotent:
@bungiman you think it should be closed? You dont even know what this pull request does! :D

bungiman:
Yeah, that is part of the reason though :P Can _you_, the author, please post a screenshot of what is supposed to be going on?

enriquemorenotent:
There you go :) http://goo.gl/ZQ0y2

bungiman:
Thank you :) So personally, I don't mind it. In fact it could be useful and makes navigation feel like a typical program where your "state-changing" stuff is at the top. With that being said I think you should make sure it does the following.
1. Plays well with stickyheader, since it is already in the codebase.
2. Since it makes the "Top" link completely irrelevant, you should remove that too.

If 1&2 are done, then I would probably pull it into my fork and vote that it go into the main branch.

thezoggy:
fyo, this pull request destroys the ability to use sb on a mobile or small resolution.

bungiman:
@thezoggy css media queries should fix that though.

@enriquemorenotent I guess media queries becomes task 3 on the list.

PR #342, opened 2012-03-25
Option for post-processor to update destination directory timestamp

aforty:
On OSX and a hard drive formatted with exFat, the timestamp on the destination directory (such as ./Season 01/) of a file doesn't get updated. This may simply be an exFat issue, I had this happen sometimes in Windows too. This messes up my particular media player of choice (Plex) where the scanner will ignore the directory based solely on the directory timestamp. So even though there's a new file in the directory (with its correct timestamp) the directory will not have its timestamp updated and thus will be ignored by the scanner because it assumes that nothing changed. 

This may not even require a visual switch in /config/postProcessor. Touching the directory timestamp seems like something that should happen regardless and the helpers.touchPath() method could always be called, simply to ensure that this has happened. Thoughts or possible complications with doing this always?


ghost:
Just out of curiosity.

Does it make a difference if files are copied instead of moved to destination directory?
(Config>Post Processing>Keep Original Files)

aforty:
No, made no difference. Really starting to think this is an exFat issue.

midgetspy:
This sounds like a pretty localized issue that could be solved with an extra_script?

aforty:
Well it's "localized" to a specific formatting but cross platform, so I'd say it's a real issue. Of course this issue is not with Sick Beard at all but I think it may be too much to ask of users who happen to be using exFat to "know" to install an extra script. Can we agree that touching a file modified date should happen when a file is copied or moved to a new destination? If we can then I'd argue that what is the harm in the code doing so explicitly? So I'm proposing getting rid of the UI option I wrote and simply touching the file path explicitly (behind the scenes, every time) upon move/copy, just in case the OS doesn't do so itself. 

I personally know two other people using exFat and have run into this issue in some shape or form (not with Plex but other things), so I'd assume there are more out there that could be experiencing this issue but not knowing the root cause. 

Thoughts?

PR #345, opened 2012-03-27
API: Sickbeard Server Filesystem Listing

ghost:
http://sickbeard.com/forums/viewtopic.php?f=8&t=4106

This adds a new command sb.remotedirs that allows you to recieve directly listings from the sick beard server. 

When you add a new root or and existing show on the web ui you are given a "choose directory" popup that allows you to 'browse' the directories on the sick beard server. This simply adds the same functionality as that but for the api system.


thezoggy:
also be sure to run the code change through pep8 and pyflakes to catch things. note we do not follow the 80 char limit

midgetspy:
Hmm... trying to decide what I think about the security implications of this.

Diaoul:
I do not want a program with API access to read my FS if that's what the pull request is about.

midgetspy:
Yeah I don't like the idea of having this potentially exposed.

PR #347, opened 2012-04-01
Add Check permissions before changing permissions

ghost:
Starting point: ERROR result in logfile POSTPROCESSER :: Failed to set permission

Also possible ERROR POSTPROCESSER :: Failed to respect the set-group-ID 
when target directory has different group-ID.

On linux when download user and SickBeard user are not the same
(like in Zebulon packages for Synology NAS) moving files will 
result in download user still owning the files.

SickBeard always tries to chmod permissions (even when already correct) 
this results in error, because only root or owner are allowed to do this.

Add Check if change is necessary (permissions already correct) like in fixSetGroupID
Add Check if user is allowed to change permissions.

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -446,6 +446,13 @@ def chmodAsParent(childPath):
     if childPath_mode == childMode:
         return
 
+    childPath_owner = os.stat(childPath).st_uid
+    user_id = os.geteuid()
+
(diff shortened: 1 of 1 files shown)


Diaoul:
+1 for this one. Permissions give many boring errors

Adeiko:
Hoping this gets added, my log is full of this errors :(

PR #348, opened 2012-04-01
Add pid files to the .gitignore file to prevent accidental commit of pid files.

ghost:


ghost:
git config --global core.excludesfile ~/.gitignore_global

PR #349, opened 2012-04-01
new provider: nzbindex.nl

nielsw:
Hi,

i've added nzbindex.nl as a search provider, would you mind taking a look at it and/or even adding it as an optional search provider?

Thanks for all your work so far / the time you may take looking at this.

Kind Regards,

Niels Wijnja


mizzos:
+1 on this.

basrieter:
+1 on this one to. We need more free search providers.

pamarama:
+1 for me too

sickbeard/providers/nzbindex_nl.py
```diff
@@ -0,0 +1,149 @@
...
+
+		results = []
+
+		for curItem in items:
+			(title, url) = self._get_title_and_url(curItem)
+
+			if not title or not url:
+				logger.log(u"The XML returned from the NZBindex.nl RSS feed is incomplete, this result is unusable: "+data, logger.ERROR)
+				continue
+			if not title == 'Not_Valid':
```
midgetspy: Should be != for clarity

sickbeard/providers/nzbindex_nl.py
```diff
@@ -0,0 +1,149 @@
...
+			if title.lower().count(showName.lower()) == 1:
+				titleStart = title.lower().index(showName.lower())
+				titleEnd = title.index(" ",titleStart)
+				title = title[titleStart:titleEnd]
+			elif title.lower().count(showName.lower()) > 1:
+				titleStart = title.lower().rindex(showName.lower())
+				titleEnd = title.index(" ",titleStart)
+				title = title[titleStart:titleEnd]
+			else:
+				title = "Not_Valid"
```
midgetspy: I'd just use None for this rather than "Not_Valid"

schumi2004:
Why not just nzbindex.com instead of *.nl ?

ghost:
Why not just nzbindex.nl instead of *.com ?

Try "nzbindex" as keyword in Google, that will give you your answer.

thezoggy:
.com = us gov control and can easily be shutdown per ICE.

sickbeard/providers/nzbindex_nl.py
nielsw: Fair enough, was a leftover from 'not title', will be changed.

sickbeard/providers/nzbindex_nl.py
nielsw: Again a leftover, will change it.

nielsw:
The name of the searchprovider isn't really relevant imho, we could change it to just 'Nzbindex' to avoid confusion but in the meantime it'll be .nl as it's hosted in the Netherlands.

midgetspy:
OK I tested this and I'm not going to be pulling it. This is not an index it's simply a usenet listing. An index filters and categorizes the content so that we can search it, this site doesn't. Usenet indexes have complex sets of regexes to decipher the release title reliable - we would need this too to make this provider useful. I tried searching for an ep of a popular show and it couldn't find any because it was parsing the titles wrong - and to parse them right is an indexers job, not SB's.

PR #352, opened 2012-04-04
Fix keywords search Exception for adding new shows

ghost:
While the new keyword search breaks the searchterm into elements to find more shows,
it returns nothing when one of the elements throws an Exception.

See also:
https://github.com/midgetspy/Sick-Beard/commit/52ac48d134f929700ae67e5bee80aa6e120d1c72#commitcomment-1170920

Add a new show 'hangs' on certain keywords.

Example search: white collar 
So keywords = ['white collar', 'white', 'collar']
first 
http://thetvdb.com/api/GetSeries.php?seriesname=white+collar&language=en
OK continue 

http://thetvdb.com/api/GetSeries.php?seriesname=white&language=en
Fails because XML is wrong. 

Log doesn't show Exception and search in webinterface seems to 'hang'.

I think it should at least return the already found.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1581,11 +1581,12 @@ class NewHomeAddShows:
 
             try:
                 seriesXML = etree.ElementTree(etree.XML(urlData))
-            except Exception, e:
-                logger.log(u"Unable to parse XML for some reason: "+ex(e)+" from XML: "+urlData, logger.ERROR)
-                return ''
+                series = seriesXML.getiterator('Series')
 
-            series = seriesXML.getiterator('Series')
+            except Exception, e:
+                # use finalURL in log, because urlData can be too much information
+                logger.log(u"Unable to parse XML for some reason: "+ex(e)+" from XML: "+finalURL, logger.ERROR)
+                series = ''
 
             # add each result to our list
             for curSeries in series:


ghost:
Thanks.

Does it make sense only to search for alternatives if a whole show's name is misspelled/ not correct?
If the whole show's name has results then showname is probably correct.

Something like a break at the end of the for searchTerm in keywords loop.

```
        if searchTerm == keywords[0] and results != []:
            break
```

thezoggy:
people complain that they cant find what they are searching for. with tvdb you cant search aliases only the exact show. so searching for 'law & order' becomes challenging because you dont know if the official listing is '&' or 'and' for example. the more precise solution will be at the top. if by your logic we did what you want.. then searching for 'top gear' would just return that.. not the multitude of top gear versions (aus/us/etc). and searching for top gear usa.. you would never get any results. anyways id rather it show more results than not enough. you dont hear anyone complain about too many shows to pick from.. we deff hear when the users think there should be results and there is none. for whatever the reason.

ghost:
Actually searching for a show like 'top gear' you will get the multitude of top gear versions (http://thetvdb.com/api/GetSeries.php?seriesname=top+gear&language=en)

But I see your point. When someone wants 'Law & Order SVU' and only searches for 'Law and Order' then you will only get the original show as a result.

My main concern was, that it can take longer to display the results when searching for a long title
with a lot of common words (especially when thetvdb has a busy day).

But if more results makes people happier, best leave it as it is.

PR #356, opened 2012-04-08
Accept comma separated list for Plex Media Server hosts

sander1:
Accept comma separated list for Plex Media Server hosts (just like in the XBMC notifier) for the rare case someone wants or needs to update more than one Plex library.


midgetspy:
What does the test function do with multiple hosts?

sander1:
The test function needs some work, it only tries to test to connect to the Plex client, not the server part. I'll see if I can come up with a solution.

thezoggy:
was taken care of in pull #524 with the xbmc/plex rewrite

PR #358, opened 2012-04-09
cross-domain api

brettjurgens:
allow other domains (and ports) to access the api.

useful so that you can make calls to the api from a different port/server, using a javascript ajax call


brettjurgens:
didn't think too much about security, if we were to whitelist only localhost, it would also fix the port issue.

i know jsonp would fix it, but i could not get that working (jQuery throws an error with the API and jsonp datatype)

thezoggy:
use reverse proxy then ?

midgetspy:
Not going to pull this, please see https://github.com/midgetspy/Sick-Beard/pull/406

PR #360, opened 2012-04-10
Fix for recent GH API change

thezoggy:
Update the max commits to detect when people are out of date (or ahead)


lad1337:
+1

midgetspy:
Heh I did this locally and never pushed... whoops. Thanks.

PR #363, opened 2012-04-13
[FIX] Issue 1876: Allow SD x264 TV Snatching

xbmcfanboy:
http://code.google.com/p/sickbeard/issues/detail?id=1876
Someone posted code that seems to work. I took that code and made this pull request.


midgetspy:
This will only affect times when both episodes are posted/found simultaneously, but it can't hurt.

PR #355, opened 2012-04-06
kat.ph torrent support

junalmeida:
This is a provider to support http://kat.ph torrents. I'm testing it for 7 months and it proves to be very valuable on snatching quality torrents for my tv shows.

PR #364, opened 2012-04-13
KickassTorrents Provider

cwc:
This contains @dinoboy197's KickassTorrents provider as well as some bugfixes.

Let me know if I need to clean anything up before it can be pulled. I can squash my bugfixes into a single commit if necessary.


midgetspy:
Is this any different than https://github.com/midgetspy/Sick-Beard/pull/355?

cwc:
Ah, did not see that one.

PR #367, opened 2012-04-13
Nonzero result code on error.

twoolie:
Make url error return non-zero result code to allow sabznbd to display retry button.


midgetspy:
Thanks, nice fix. I've pulled this to the development branch, where all new code will be going from now on.

PR #368, opened 2012-04-14
displayShow poster art, other fixes

giantspook:
- adds the show's poster to the top of its displayShow page in the summary block
- fix the banner size to match the CSS (was making images blurry)
- fix the content-type header to match the image type


midgetspy:
Pulled to development except the displayShow changes, not sure about that yet as it's being worked on elsewhere also.

PR #369, opened 2012-04-14
Adds a 720p 5.1 quality.

zglurb:
Adds a 720p quality in the custom quality list.
It does not work with Newzbin because of Newzbin's lack of audio filter.

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -62,6 +62,7 @@ class Quality:
     HDWEBDL = 1<<3 # 8
     HDBLURAY = 1<<4 # 16
     FULLHDBLURAY = 1<<5 # 32
+    HD51 = 1<<6 #64
 
     # put these bits at the other end of the spectrum, far enough out that they shouldn't interfere
     UNKNOWN = 1<<15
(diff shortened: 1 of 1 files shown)


thezoggy:
quality is for video not audio. dont think this would get pull.

zglurb:
I was about to close it anyway because 5.1 files won't be matched by HDTV et Web-DL filter anymore. I was thinking of adding "720p 5.1" to the Any and HD quality presets but with your comment, I am not sure anymore.

Do you think a checkbox "Download only episodes with a 5.1 audio" in the editShow page would be accepted? It needs more development but I'd do it if it has a chance to be pulled since I'd really like this feature in Sickbeard.

thezoggy:
well its hard to enforce audio stuff.. there are certain networks that dont do it.. for example mad men web-dl = 2channel. but hd broadcast are 5.1. then also some shows just dont have it at all..  
imo, if people are really that picky about their audio..they can just use the bluray rip quality under custom.

zglurb:
Some shows are broadcasted in stereo while the web-dl version have 5.1 (NCIS for exemple). For those shows, I just wait a few days for the web-dl to be released.

But other shows are broacasted in 5.1 and released in stereo in a 700MB file (for one hour shows). Those formats didn't exist last year (at least, I don't think so) and Sickbeard always picks them since they are the first to be released. I just want to be able to ignore them and only download the ones above 1GB.

PR #370, opened 2012-04-17
Newzbin "720p WEB-DL" searches fixed

azcom:
Resolved issue in newzbin.py where no results were ever found for searches with "720p WEB-DL" quality selections if newzbin is the only configured provider.

New Web-DL source attribute now added to Newzbin so changed newzbin.py search code for "720p WEB-DL" to use it and fix this long standing issue.


midgetspy:
Great, thanks. I've pulled this to the development branch, where all new code will be going from now on.

PR #371, opened 2012-04-18
fix sql query for multi episods int he cache

lad1337:
the sql query would not catch multi episode releases 
the extra %'s around the episode string lets the LIKE also catch multi eps


midgetspy:
Pulled to development

PR #374, opened 2012-04-18
debianized init.ubuntu with test for /etc/default/sickbeard

gargolito:
to configs out of the init.ubuntu file. this way updates won't break configs of modified init.ubuntu

### DIFF
--- a/etc/default/sickbeard
+++ b/etc/default/sickbeard
@@ -1,0 +1,24 @@
+############### EDIT ME ##################
+# path to app
+APP_PATH=PATH_TO_SICKBEARD_DIRECTORY
+
+# path to python bin
+DAEMON=/usr/bin/python
+
+# Path to store PID file
+PID_FILE=/var/run/sickbeard/sickbeard.pid
+PID_PATH=`dirname $PID_FILE`
+
+# script name
+NAME=sickbeard
+
+# app name
+DESC=SickBeard
+
+# user
+RUN_AS=SICKBEARD_USER
+
+# data directory
+DATA_DIR=~/.sickbeard
+
+############### END EDIT ME ##################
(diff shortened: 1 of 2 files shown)


ghost:
I think the --datadir was introduced so users on the same system 
can have there own sickbeard.db, cache and Logs.
Using the ~ can do just that, providing the startscript is run by the user.

In this case the init.ubuntu is always run as root. 
So ~ expands to home folder of root, this is a no no.
See comments https://github.com/midgetspy/Sick-Beard/pull/366

So users want to fix this 'homefolder' issue in different ways.

Like using /home/${RUN_AS} or get homedir from /etc/passwd
The downside of these fixes are, that homes are not always in /home and
you can't depend on /etc/passwd being there.

So when one of these fixes gets merged, then other users want to change this back 
or try something else.

gargolito:
The contents of /etc/default/sickbeard were pulled from the configuration stanzas at the top of init.ubuntu the change I amed and I am requesting does not change in anyway the functionality of init.ubuntu UNLESS /etc/default/ubuntu exists AND it has declared variables with values other than those already in init.ubuntu - these variables are supposed to be changed by ubuntu users so whatever method of declaring any one variable is moot. You are right that if you use the shorthand ~/.sickbeard, the dirname would be in the context of what user the service is running as, and that this would cause problems in instances where more than one user uses sickbeard in one machine. It will also cause problems for users that do not understand the distinction between user context and service accounts. The change I requested is minor and will not affect anyone who is already using init.ubuntu.

giantspook:
PIDFILE should be in /run, not obsolete /var/run per http://lwn.net/Articles/436012/

init.ubuntu should really all be an upstart job instead of an init script, though.

PR #375, opened 2012-04-19
Better GitUpdateManager._find_git_branch()

mikenerone:
GitUpdateManager._find_git_branch() had the job of determining the currently-checked out git branch. This change replaces it with an "official-er" technique for doing that (See "git help symbolic-ref" for details). More reliable, future-proof, and faster, to boot.


thezoggy:
i dont think this works if the user doesnt track the branch... sline tried it earlier and she had problems

midgetspy:
If you don't track the branch then how would you update anyway?

PR #376, opened 2012-04-20
Unittest additions

lad1337:
added simple full pp test

and a yml for http://travis-ci.org/
you might want to add your repo to travis to

my one is at http://travis-ci.org/#!/lad1337/Sick-Beard

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,12 @@ autoProcessTV/autoProcessTV.cfg
 server.crt
 server.key
 
+#  SB Test Related   #
+######################
+tests/Logs/*
+tests/sickbeard.*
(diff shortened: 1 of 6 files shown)


thezoggy:
we need to add docstrings to each test so we know what we are testing and what should be returned. like do we have a test that tries to load each provider? or notifiers? or metadata? and will it update automatically when new ones are added? what about our logger.. will it test that ? can we have it test logging rollover (set really low rollover limit / populate with fake data to force rollover) and then compare size? have it check connectivity to tvdb / providers / etc before running test? etc

thezoggy:
also run through pyflakes/pylint/pep8

PR #377, opened 2012-04-21
jQuery and jQuery UI updates

thezoggy:


thezoggy:
pulled into ui-html5. closing

PR #384, opened 2012-05-01
Fixed up issue in build 495 with CSS styling on Sort Arrows

demyte:
Hi,

A small fix for CSS style on the sort arrows in the table headings where they were not set to no-repeat

Cheers

James


thezoggy:
you just need to clear browser cache.. no 'fix' is needed

notice we break apart the images from the css so we can do sbRoot (reverse proxy):
https://github.com/midgetspy/Sick-Beard/blob/master/data/css/tablesorter.css

PR #385, opened 2012-05-02
Add zip support tvdb_api

ghost:
Tried to stay as close as possible to the original 1.5 version.

If enabled (default) download language.zip file, extract/ unzip all xml files to tvdb cache_dir.
Try using the unzipped xml files first to get data.
If fails/ not possible etc uses the 'original' url as fallback.


thezoggy:
we already got zip support added to the official lib, https://github.com/dbr/tvdb_api

ghost:
I see.

But SickBeard doesn't use the official (1.5) lib. It's modified.

When 'replacing' SickBeards version with the official lib, 
you have to fix: http_proxy, cache=recache, cache_dir and http cache-control too.

thezoggy:
yes, which is what we did. https://github.com/lad1337/Sick-Beard/tree/tvdb_api_update

thezoggy:
i've shared your pull request with the rest of the dev team to investigate though :)

jsturgis:
I can confirm that on Ubuntu 12.04 I replaced the sickbeard tvdb_api.py with the latest official lib without any changes and everything works well, also it fixes the xml error I was getting with the sickbeard version.

ghost:
Switching to development. Made a rebase mess.

PR #389, opened 2012-05-07
Separated Update and Notify XBMC Hosts

WordMasterRice:
Updates the config to allow for separate hosts to update XBMC and to be notified.  This change is primarily useful for people with multiple XBMC boxes that want all boxes notified but only 1 to actually do the updating.


thezoggy:
cant you just use sql backend for xbmc? and have all non primary xbmc boxes use that sql db?

WordMasterRice:
That is what this is for.  There is no "Primary" as far as the MySQL backend is concerned, all of the boxes can update the database.  Because of this you generally only want one doing the updating, so if you have 2 boxes running you don't want them both trying to update the same database.  Conversely you would like them both to get the notification that a download has started or finished.

thezoggy:
gotcha. and we can send the update or rescan to multiple (, delineated) list of clients?

WordMasterRice:
When you are using the SQL backup you should only send the update to 1 client, (normally a server that runs XBMC as well) and then you send the download popup (which is purely cosmetic) to all the clients.

Basically all I did was leave the update/rescan piece alone, and created a new list for clients that are essentially "Read Only" and shouldn't be sent the update command, only the little popup notification.  This allows XBMC to work in the client/server sort of way that the SQL backend was intended for.

thezoggy:
right, but not everyone uses sql backend.. just making sure we dont take away existing funcationality

WordMasterRice:
Correct, I could add an option so that if you have a simple setup that the Update host is used throughout, but it is pretty self explanatory in the config screen that I modified.

ghost:
Just a general question.
When a configuration setting is changed,
do you also have to 'fix' existing configuration.

For example in this case. 
When SickBeard is updated then notifications are handled with a 
different option. I guess current configuration 'breaks'.

Do one need to add something like:

```
XBMC_NOTE_HOST = check_setting_str(CFG, 'XBMC', 'xbmc_note_host', '')
    if XBMC_NOTE_HOST = '':
        XBMC_NOTE_HOST = XBMC_HOST
```

What about when option names are changed?
Lets say someone later on wants to change XBMC_NOTE_HOST to XBMC_NOTIFY_HOST 
or XBMC_HOST to XBMC_UPDATE_HOST

What is general rule of thumb?

Anyway.

Just a question about this update.
Is the scenario always 'update all libraries' or 'update only one library'?
Or can there be a mix. Say 3 machines shared database and 1 machine own database.

Otherwise I think a checkbox XBMC_UPDATE_ONLY_ONE_LIBRARY instead 
of text XBMC_NOTE_HOST could also be used.
Then just put a break at the end of the update_library for loop

```
if sickbeard.XBMC_UPDATE_ONLY_ONE_LIBRARY:
    break
```

WordMasterRice:
You bring up a good point, if you upgrade the current functionality does break but I might have a way around that though.

I wouldn't simply check to see if XBMC_NOTE_HOST is empty because their may be a legitimate reason that it is empty.  It would be safer to check and see if the option is set in the ini and if it isn't assign it to XBMC_HOST to maintain current function.  The user can then change it if they wish to leverage the split.

Both of the fields support the comma delimited host option, so in your case you would put 1 of the hosts from the shared DB, along with the machine that runs its own database in the Update field, and put all 4 in the notify field.

I did think about an UPDATE_ONLY_ONE option to allow putting in the host only once but I didn't want to add yet another option.

ghost:
I think, we mean the same.
XBMC_NOTE_HOST will be empty if it's not in the config.ini
because it's the default value (last value) in: XBMC_NOTE_HOST = check_setting_str(CFG, 'XBMC', 'xbmc_note_host', '')

When using UPDATE_ONLY_ONE I meant an on/off checkbox INSTEAD of a textbox. 
So it is not really an extra extra option :-)
You fill in all the hosts. So all of them are notified, but if UPDATE_ONLY_ONE/ SHARED_DB is on 
then 'update' stops early (after first host)

But I don't know.

With an extra textbox it's more flexible.

With a single checkbox (on/off) there's less room for mistakes 
(don't have to 'fix' upgrade, because you can default to off)

WordMasterRice:
Ahh I see what you mean.  I think that having just the check box leaves another hole that would have to be addressed at some point where you might want to update 2 boxes and notify 5 or something, which the check box couldn't handle.

ghost:
Sure, textbox is more flexible.

thezoggy:
is this still being updated or ready to test before merged?

WordMasterRice:
There is one change that I want to make tonight and it will be ready for testing.  I'd like it to default to whatever XBMC_HOST is in XBMC_NOTE_HOST isn't in the configuration so that for everyone who already has it setup the way they want it, it will continue to work as normal.

WordMasterRice:
Sorry about the delay, the pull request should be ready to go for testing.

WordMasterRice:
Sorry about that, just before I did the commit I downloaded the github for windows client and it did some funky stuff.

ghost:
No worries. I hope they can cherry-pick the good stuff.

Otherwise try rewriting history :-)

See http://git-scm.com/book/en/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages
If you get it right on local then I guess you have to do a git push with --force option

neoflex:
+1 for this, would be very usefull for people with multiple xbmc instances

midgetspy:
I haven't reviewed all of the changes but I will say off the top: a feature like this is useful to a VERY small portion of SB users - if it should be in master I think it should be an "advanced" (aka config-only) option. I'd suggest simply adding a config-only field which is a list of IP:ports and if it's filled it will override the notification setting for which XBMCs to update library for. I think that'll add the functionality without needlessly complicating things for the 99% use case.

mxhob1:
Has there been any progress on this since the last comment? This has just become relevant to me (5 XBMC machines with a central SQL database). 

If not, I should be able to take some time to bring it in line with @midgetspy's interpretation so it can get merged into master soon.

thezoggy:
this has been taken care of with pull #524 with the xbmc rewrite. now we only send library updates to the first host. this should allow people that use sql backend to just set their 'master' server first.

WordMasterRice:
So just out of curiosity, now it will be broken for everyone using separate databases?

gregs007:
Yes, it is indeed broken now for those of us not using the sql backend (i.e. running different versions of xbmc across different media servers).

PR #395, opened 2012-05-23
Update the scene exception db before checking for updates

jtatum:
This addresses an issue that can occur when using a zipball/tarball of a tagged version. The version checker encounters an exception which ends the thread before the scene exceptions are populated. Swapping the order so the more stable exception retrieval runs first fixes the issue.

### DIFF
--- a/sickbeard/versionChecker.py
+++ b/sickbeard/versionChecker.py
@@ -48,10 +48,10 @@ class CheckVersion():
             self.updater = None
 
     def run(self):
-        self.check_for_new_version()
-        
         # refresh scene exceptions too
         scene_exceptions.retrieve_exceptions()
+
+        self.check_for_new_version()
 
     def find_install_type(self):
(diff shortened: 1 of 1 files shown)


midgetspy:
What exceptions are thrown, can you post a log? A better fix would be to catch those exceptions and log them and then carry on.

thezoggy:
@jtatum any update on this?

jtatum:
So, basically, I was using homebrew to install SB on OS X. Homebrew gets a specific revision of the source from GH, but doesn't do it as a clone, it just copies it in (having its own update mechanism).

This is probably an unsupported way to do things, but in terms of actual functionality, the only thing it broke was the scene exception DB since the update mechanism fails before the scene exceptions get updated. In any case, I just switched to installing it from Git like normal, so...

I can definitely update the pull to record the exception, though, if desired.

PR #403, opened 2012-06-02
Make restart code work when behind reverse proxy

giantspook:
the restart code used to not take the web_root case into account. it would redirect to the wrong location following a restart.

I have tested this patch in the following cases:
- http -> http restart (no web_root)
- http -> https restart (no web_root)
- https -> https restart (no web_root)
- and of course, HTTP-via-reverse-proxy -> same, with web_root configured, apache for the proxy

the original code didn't handle the https -> http case, and i haven't added support because anyone who manages to get https working well enough to switch back to http should be able to figure it out themself!

### DIFF
--- a/data/js/restart.js
+++ b/data/js/restart.js
@@ -1,78 +1,74 @@
-if (sbHttpsEnabled != "False" && sbHttpsEnabled != 0) 
-	var sb_base_url = 'https://'+sbHost+':'+sbHttpPort+sbRoot;
-else
-    var sb_base_url = 'http://'+sbHost+':'+sbHttpPort+sbRoot;
+var force_https = (sbHttpsEnabled != "False" && sbHttpsEnabled != 0);
+// We use the pathname and .. segments so we don't need to worry about how a
+// proxy may alter the path.
+var url_stem = '//' + window.location.host + window.location.pathname + '../../';
+var old_base_url = window.location.protocol + url_stem;
+// The config might have changed to enable HTTPS, in which case we'll need to
+// check a new URL and redirect there.  Since web_root can only be changed by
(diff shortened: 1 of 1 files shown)


xombiemp:
Actually, the original code did handle the https -> http case.  I just tested it in Safari, Firefox and Chrome on a Mac and all four cases worked great in each browser.

xombiemp:
I tested your branch with no proxy or web_root and here are my results with Chrome on Mac:
http://localhost:8081/ -> http://localhost:8082/  WORKS
http://localhost:8082/ -> https://localhost:8082/ WORKS
https://localhost:8082/ -> https://localhost:8081/ Error 102 (net::ERR_CONNECTION_REFUSED)
https://localhost:8081/ -> http://localhost:8081/  Error: The restart has timed out

I tried these four tests with the current master and they all worked.

Also, did this commit not fix the problem? 
https://github.com/midgetspy/Sick-Beard/commit/1c26733063a17cb188d27a7dd173075caede60dd
I ask because I don't currently use a reverse proxy.

giantspook:
interesting, i didn't think to test changing the port since it doesn't make much sense behind a reverse proxy. i will hack on it a bit.  thanks for your help!

the commit you linked actually contains a bug: it uses the [X-Forwarded-Host header](http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#x-headers) wrong (it's like the Host header and includes the port if it's not the default, so the X-Forwarded-Port header is redundant).  i think that restart broke for me about this time, since Apache sends the X-Forwarded-Host header by default in reverse proxy mode, but not the X-Forwarded-Proto.  i will probably need to change my approach a bit.

PR #406, opened 2012-06-07
API support for JSONP

mattmelling:
Adds support to webapi for JSONP calls using the 'callback' or 'jsonp' request argument.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -158,9 +158,13 @@ class Api:
     def _out_as_json(self, dict):
         """ set cherrypy response to json """
         response = cherrypy.response
+        request = cherrypy.request
         response.headers['Content-Type'] = 'application/json;charset=UTF-8'
         try:
             out = json.dumps(dict, indent=self.intent, sort_keys=True)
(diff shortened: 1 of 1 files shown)


piejanssens:
I understand what it enables you to do but could you sketch a situation where this could be useful?

Live UI for MediaFrontPage maybe?

mattmelling:
It's useful for me as I'm writing a HTML/JS app to query the state of some services on my LAN from my phone's browser, and regular AJAX requests will not work because they will violate the origin policy. As you suggested a utility like MediaFrontPage could use JSONP to implement a live UI even if SB was running on a different machine. My app is already able to query SABnzbd and a couple of other services which are able to handle JSONP requests, but it would be excellent if it could query Sick Beard as well.

gaetan-petit:
Very usefull!
Much cleaner and way more secure than : https://github.com/midgetspy/Sick-Beard/pull/358

midgetspy:
Great, pulled into the development branch. In the future please just submit pull requests against development directly, thanks!

mattmelling:
Fantastic, will do in the future.

PR #407, opened 2012-06-09
Fix unicode fix episode names

ghost:
Just added a missing ek.ek to fixEpisodeNames.


midgetspy:
Thanks, but this function has been replaced in the custom_naming branch which will be pulled soon so I'm not going to pull this now. custom_naming should be merged in the next couple days.

PR #408, opened 2012-06-09
Fix no dir manual postprocessing

ghost:
When no folder is supplied in webinterface Manual postprocessing, dir is not None but ''

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1475,7 +1475,7 @@ class HomePostProcess:
     @cherrypy.expose
     def processEpisode(self, dir=None, nzbName=None, jobName=None, quiet=None):
 
-        if dir == None:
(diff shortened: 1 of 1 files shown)


sickbeard/webserve.py
```diff
@@ -1541,7 +1541,7 @@ def index(self):
     @cherrypy.expose
     def processEpisode(self, dir=None, nzbName=None, jobName=None, quiet=None):
 
-        if dir == None:
+        if dir == None or dir == '':
```
midgetspy: This can be "if not dir" which covers both cases. Good catch.

ghost: I considered changing it to "if not dir"  first, but thought we don't want  'dir = True' to enter processTV.processDir either.

But I'll change it.

PR #409, opened 2012-06-09
Fix season tumbnails always downloading

ghost:
Check first if there are already season thumbnails.

Otherwise if you have already season thumbs, it downloads them anyway.
Finds out the imagefile already exists and not write them.

### DIFF
--- a/sickbeard/metadata/generic.py
+++ b/sickbeard/metadata/generic.py
@@ -215,8 +215,11 @@ class GenericMetadata():
     
     def create_season_thumbs(self, show_obj):
         if self.season_thumbnails and show_obj:
-            logger.log("Metadata provider "+self.name+" creating season thumbnails for "+show_obj.name, logger.DEBUG)
-            return self.save_season_thumbs(show_obj)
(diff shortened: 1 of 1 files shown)


sickbeard/metadata/generic.py
```diff
@@ -445,6 +445,9 @@ def save_season_thumbs(self, show_obj):
         # to present to user via ui to pick down the road.
         for cur_season in season_dict:
 
+            if self._has_season_thumb(show_obj, cur_season):
```
midgetspy: This should be up on line 217 where the rest of the checks are - not sure why this one got missed though, thanks.

ghost: I tried it on line 217 first, but couldn't figure out how to get 'season' there.

ghost:
It's on L217 now, using show_obj.episodes.iteritems()
Guess this also 'solves' Fix an issue where season banners don't always download for the most recent season https://github.com/midgetspy/Sick-Beard/pull/386

PR #411, opened 2012-06-09
Redirect when show dir exists

ghost:
Just an idea. 
When you add a new show and the folder already exists, redirect to add existing show.


chrispaolini:
That seems like a good change.

PR #412, opened 2012-06-09
Fix better quality proper

ghost:
When the status of an episode is snatched (proper) it never gets replaced
by backlog if you want a higher quality.


thezoggy:
just wanted to say that thank for your recent submissions. they are small but noteable changes. also glad to see you follow pep8 style guidelines :) I've passed on my recommendation for merging all your submissions into development as soon as we wrap up custom_naming this week

midgetspy:
Yep, greatly appreciated.

PR #417, opened 2012-06-16
Add webhook notification

bouk:
This adds the possibility to make Sick Beard send a POST request to a specified URL upon snatch or completion of a download. The request has three parameters:
- ep_name (the name of the episode)
- notification_type ('snatch', 'download' or 'test')
- notify_string (a human readable version of the notification type)

I tried my best to have the code in the same style as the rest of the project
There's is no icon yet for the config page 

This is useful for people who want to make their own custom notification, with this they can just throw a PHP script on a server somewhere and let that do the work instead of having to modify Sick Beard.


midgetspy:
This is what "extra scripts" is for: http://code.google.com/p/sickbeard/wiki/AdvancedSettings

bouk:
Ah, you're right. Might still be useful for some people maybe? Else I'll
just close this
Op 18 jun. 2012 00:10 schreef "Nic Wolfe" <
reply@reply.github.com>
het volgende:

midgetspy:
Thanks for the contribution in any case.

lgordon:
Maybe you could post the script as a gist then add a link in the wiki so others can find it, I posted my script: https://github.com/midgetspy/Sick-Beard/wiki/Extra-Scripts

thezoggy:
http://code.google.com/p/sickbeard/wiki/AdvancedSettings

PR #423, opened 2012-06-26
fix silly upstream error 'NoneType' object has no attribute 'makefile'

ghost:
Updated Class Http(object)
copied def _conn_request(self, conn, request_uri, method, body, headers) from httplib2-0.7.4
and added ssl_SSLError = ssl.SSLError


midgetspy:
Awesome, thanks!

PR #208, opened 2011-09-29
Issue 1629: Add full Synology index command set

mreid-tt:
This attempts to add the complete 'synoindex' command library to Sick-Beard as well as function calls for existing I/O operations in the PVR app, namely creating folders as well as adding and removing files from the library.

PR #430, opened 2012-07-06
Issue 1629: Add full Synology index command set (addendum)

mreid-tt:
This amends the previous pull request #208 which added the complete 'synoindex' command library to Sick-Beard. The specific fix is to include the notifier for the creation of a new TV show.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1797,6 +1797,8 @@ class NewHomeAddShows:
             redirect("/home")
         else:
             helpers.chmodAsParent(show_dir)
+            # do the library update for synoindex
+            notifiers.synoindex_notifier.addFolder(show_dir)
 
         # prepare the inputs for passing along
         if seasonFolders == "on":


mreid-tt:
Any idea when you guys might get around to pushing this?

mreid-tt:
It seems that somewhere in the mix some of the original code from pull request #208 (specifically the trigger to add folders) seems to have been lost. Rather than creating a convoluted mess (plus I don't know exactly when my original code got lost - maybe in the switch to merging from the development branch), I'm going to withdraw this request and submit a new one combining both needs.

Roeland54:
You didn't submit a new request. Are you still working on a solution? It would be very helpful and it looks you are pretty close.

mreid-tt:
Yes, this is true... I wrote to the author the following note some months ago with no response:

I previously submitted some code under pull request #208 for which some of it is no longer in the current master. Additionally based on my observation that you seem to be moving toward my suggestion of having the I/O functions executed in the abstract 'helpers' function, the addendum pull request #430 now no longer seems sensible (so I cancelled it).

I also note you have code in the helpers to determine if a file is media or not as well as a lot of new code in the post processor for moving files. Now the synoindex has all these functions included in the notifier I coded in pull request #208 but I need some guidance from you as to how I should attempt to reintegrate this into the project. Shall I now be considering all addendum to the helpers funtion only? It just seems to me that you are not complete with this transition (for example I see system function calls to create folders both in the postprocessor directly and called from there to the helpers).

I would like some guidance on how I can fix the now broken Synology code for folder handling or alternately, should I wait for a few more revisions for the project to use the helpers function for all I/O and then try to fix it?

mreid-tt:
Okay, I took the time to do a quick patch and submitted as pull request #591. Hopefully it will get merged soon... Be sure to comment in the new pull request to give it some attention.

PR #431, opened 2012-07-06
SickBeard UI - html5

thezoggy:


arogl:
With the change in the custom naming code should the logic of line 130 in data\interfaces\default\manage.tmpl be reversed, or should the column heading for "Season Folders" be changed to "Flatten Folders"?

thezoggy:
yeah just the header label is wrong, i'll fix that in a moment. thanks for catching this!

twe4ked:
Looks great guys, thanks!

PR #435, opened 2012-07-15
Add refreshDir to testRename

ghost:
Making sure episode location in database is same as files on disk.
Otherwise when a file is deleted or renamed outside SickBeard,
testRename raises exception: Cannot find 'location' while searching for 'location'


ghost:
Maybe better to just catch the exception.

PR #436, opened 2012-07-16
Add ability to set SABnzbd/NZBGet download queue priority for each show

lebedevcoder:
config.ini option "display_queue_priorities = 1" in order to display queue priority controls/info.
Download queue priorities can be defined for both recent episodes and
older episodes The priorities available for NZBGet are "Default" and "High".
The "Default" priority causes the NZB to be added
to the bottom of the download queue, while the High priority will cause the 
NZB to be added to the top of the queue. All of SABnzb's priorities are
available. Using the "Default" priority will cause SABnzb to assign the 
NZB with the priority for the Category of that NZB.

Sick-Beard defaults to assigning recent episodes a priority of
High, and older episodes a priority of Default. It is possible to
set different priorities when adding a new show, as well as change the
Sick-Beard defaults. The priorities for an existing show can be changed
within that show's edit page.

Priorities are valid only for shows being downloaded using
SABnzbd or NZBGet. They have no effect on torrent downloads or when
sending NZBs to a black hole. All options and info regarding priorities
will be hidden if neither SABnzbd or NZBGet is configured as the
download method even if "display_queue_priorities = 1" .


lebedevcoder:
I forgot to mention that these changes are to address Issues 568 and 406:
http://code.google.com/p/sickbeard/issues/detail?id=568
http://code.google.com/p/sickbeard/issues/detail?id=406

mxhob1:
Definitely interested in seeing this pulled into development. Hopefully it will be!

villain:
any chance this will be committed soon? with shows being DMCA'd having a force priority for those being taken down is a good solution

piejanssens:
+1

itsChrisPhillips:
+1 here as well. I don't "NEED" this feature in my situation, although it would be a "nice to have." However, I can see use for it and especially in place like here (Australia), On-Peak\Off-Peak data plans are pretty much the norm. That and our data rates are appalling at best! Any ideas where this is at? I'm considering maybe I refresh my Python skills (haven't coded in it for YEARS) and contribute to this if I can.

PR #437, opened 2012-07-16
Added a post-processing option that removes a TV Show year from the episode filename when relevant.

dhayab:
Years in brackets are incorrectly viewed as Season-Episode by the Apple TV Media Browser parser, which retrieves erroneous metadata.

Eg. Show Name (2005) - 1x01 - Ep Name >>> Show Name - 1x01 - Ep Name

### DIFF
--- a/data/interfaces/default/config_postProcessing.tmpl
+++ b/data/interfaces/default/config_postProcessing.tmpl
@@ -185,6 +185,18 @@
                         </div>
 
                         <div class="field-pair">
+                            <input type="checkbox" name="naming_strip_year" id="naming_strip_year" #if $sickbeard.NAMING_STRIP_YEAR then "checked=\"checked\"" else ""#/>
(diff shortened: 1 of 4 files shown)


midgetspy:
All new changes now go into the development branch and unfortunately the whole renaming system has been redone so this won't work there. But that said I don't think this feature is something I want in SB except as possible an advanced option (aka config.ini only, no UI).

The reason is that by stripping the year the file can no longer be recognized as being part of the correct show. Combine that with the fact that most apple TV users will just use XBMC and this is an option that very few users will need.

Thanks for the submission though, if you re-submit against development with no UI changes I will look over it again. Or feel free to just keep it in your fork for other apple TV users.

dhayab:
Oh, nice, I didn't follow the development of this project for the last few weeks, the custom naming looks great. I'll adapt my code so it works with the development branch and keep it on my fork. Thanks for your input.

PR #438, opened 2012-07-18
Fix TVDB XML error for the daily show

jsturgis:
I assume you already have this fix planned but thought I would submit it anyway


midgetspy:
Should be fixed as of pull request 438, but thanks.

PR #440, opened 2012-07-22
Added a new torrent search provider called TwitterTorrents that uses twitter streams similar to @eztv_it

djay:


alexgrist:
Definitely need this or something similar added, ezrss has become unreliable and doesn't appear to update frequently.

Tried and tested this since ezrss stopped working and it's a been a great replacement.

junalmeida:
Nice, my branche have 3 new torrent providers, and tranmission and utorrent direct communication.
If this is useful i can make a pull request.

alexgrist:
Would be good to see that included too, although NZB seems to be focus it would be great to have more Torrent options and especially direct communications with uTorrent.

sickbeard/providers/twittertorrents.py
```diff
@@ -0,0 +1,97 @@
...
+        data = self.provider.getURL(url)
+
+        return data
+
+    def _parseItem(self, item):
+
+        (title, url) = self.provider._get_title_and_url(item)
+
+        if not title or not url:
+            logger.log(u"The XML returned from the TvTorrents RSS feed is incomplete, this result is unusable", logger.ERROR)
```
arogl, line 90: Need to change TvTorrents to TwitterTorrents?

thezoggy:
i would suggest running your code through PEP8 to fix some of the cosmetic/whitespace issues. with regards to pep8, you can ignore `E501,E201,E202,E261` as we dont conform to the 80 char limit or the too many branch logic concerns.

codeman334:
I'm in dire need to implement this. But when I overwrite the changed files, SickBeard crashes and never starts. Any ideas? I'm running alpha build 496.

codeman334:
I'm actually really glad no one responded.  I was so eager to make this happen that I didn't realize this fix doesn't work on the binary install of SickBeard.  Installed Python+Cheetah and all was golden...Updates perfectly too.

jacobjohn90:
Why can't i find any of the changed files? I would really like to update these changes so I can use eztv''s torrent!

djay:
You have to checkout my forked version at https://github.com/djay/Sick-Beard which includes the changes.
It's probably out of date now and I need to remerge it but it still works fine for me. Pity sickbeard doesn't have a proper plugin architecture :(

djay:
Actually I think https://github.com/djay/Sick-Beard/tree/piratebay_twitter is the best one to pull from as it has piratebay plugin merged in.

PR #441, opened 2012-07-22
Add optional full episode output to season list in API

pleo1:
I modified the `show.seasons` API method to accept an optional `full` parameter which allows the `description` and `location` to be included in the season list in addition to the current fields.  The original behavior of this method is unchanged, as `full=0` is the default.


midgetspy:
Please submit this against the development branch, all new features go there now. And thanks.

PR #446, opened 2012-08-05
Patch to remove some special Characters from filenames to circumvent errors

mneuhaus:
On my Windows machine with NTFS partitions Sick-Beard somehow managed to put several files onto it with special characters in it like : , ", ? and so on.

This pr fixes this by removing such characters from the to be renamed filename

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1452,5 +1452,7 @@ class TVEpisode(object):
         if naming_use_periods:
             finalName = re.sub("\s+", ".", finalName)
 
+        finalName = ''.join(i for i in finalName if i not in r'\/:*?"<>|')
+
         return finalName
 


thezoggy:
you using the development branch ? whats your renaming scheme? you have sab renaming? why i ask is because those characters are already taken care of by the sanitizeFileName function. we also trim spaces and .'s from start and ending filename amongst some other things

mneuhaus:
weird, although i'm not using the latest development branch to be honest, i've just checked the helpers.renam_file and sanitizeFileName which are identical to the current development branch, and yet i somehow managed to get that crap to my disk. i'll have to test that again sometime this week with the latest dev branch.

midgetspy:
Yeah I'm going to close this for now, if you can reproduce the bug in development then submit a patch against development and I'll pull it. Thanks.

PR #451, opened 2012-08-08
Add fix duplicate shows and fix orphan episodes in MainSanityCheck

ghost:
Added two more sanity checks for the sickbeard.db

Duplicate Shows: Shows with the same tvdb_id
Remove all but one.

Orphan episodes: Episodes with no tvdb_id parent in tv_shows
Remove all

### DIFF
--- a/sickbeard/databases/mainDB.py
+++ b/sickbeard/databases/mainDB.py
@@ -28,27 +28,60 @@ from sickbeard.name_parser.parser import NameParser, InvalidNameException
 class MainSanityCheck(db.DBSanityCheck):
 
     def check(self):
+        self.fix_duplicate_shows()
         self.fix_duplicate_episodes()
+        self.fix_orphan_episodes()
+
+    def fix_duplicate_shows(self):
+
+        sqlResults = self.connection.select("SELECT show_id, tvdb_id, COUNT(tvdb_id) as count FROM tv_shows GROUP BY tvdb_id HAVING count > 1")
+
+        for cur_duplicate in sqlResults:
+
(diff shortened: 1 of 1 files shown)


midgetspy:
Are these scenarios that you've encountered? If so I'd like to try to reproduce them as well.

ghost:
I have only seen the MultipleDBShowsException in the forum,
and didn't like resorting to external tools there:

http://sickbeard.com/forums/viewtopic.php?f=4&t=4789
http://sickbeard.com/forums/viewtopic.php?f=4&t=4715

The only way I got multipleDBShows in my test database, when I used SQL to put it there.

The Orphan episodes I found on my own (test) system  too. 
Don't know how long they have been there, so don't know what made them stay.
Orphan epsiodes normally isn't a big deal, only when you check total downloads.
I found them trying to figure out why the total download count was not correct in this question http://sickbeard.com/forums/viewtopic.php?f=4&t=4774 (was another problem though)
This person http://sickbeard.com/forums/viewtopic.php?f=4&t=4771 jumped in and had orphans also.

thezoggy:
prob an issue when the ep get deleted from tvdb but we didnt delete them on our end. ive had this happened with topgear when they shuffle episodes around in specials..

ghost:
For the orphans, the only thing I could think off:

```
    myDB.action("DELETE FROM tv_episodes WHERE showid = ?", [self.tvdbid])
    myDB.action("DELETE FROM tv_shows WHERE tvdb_id = ?", [self.tvdbid])
```

The second delete doesn't check if the first one succeeds. 
But I guess it only fails if the first delete was against a locked database and therefore didn't delete the episodes
and the database was unlocked just before/ during the second delete.

PR #452, opened 2012-08-10
Add NMJv2 notifier

racquemis:
The already present NMJ notifier doesn't work correctly on the popcorn hour 300 series. 
On the 300 series the device no longer returns the NMJ location from the telnet commands the notifier issues.
By manually entering the URL for the API call it was able to start the scanner on the device. However after recent firmware upgrade the device now requires a second API call to specify the scan directory.

At setup the NMJv2 notifier will issue a telnet command which makes the device search every attached storage device for the directory with the name 'nmj_database'. It will stop on the first occurence and the full path to the database is returned and saved.
Next the required API calls are made and the A300/C300 will update it's NMJ database.


racquemis:
pull request to wrong branch

PR #453, opened 2012-08-10
Added NMJv2 Notifier

racquemis:
The already present NMJ notifier doesn't work correctly on the popcorn hour 300 series. 
On the 300 series the device no longer returns the NMJ location from the telnet commands the notifier issues.
By manually entering the URL for the API call it was able to start the scanner on the device. However after recent firmware upgrade the device now requires a second API call to specify the scan directory.

At setup the NMJv2 notifier will issue a telnet command which makes the device search every attached storage device for the directory with the name 'nmj_database'. It will stop on the first occurence and the full path to the database is returned and saved.
Next the required API calls are made and the A300/C300 will update it's NMJ database.

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -258,6 +258,57 @@
                     </fieldset>
                 </div><!-- /nmj component-group //-->
 
+				<div class="component-group clearfix">
+                    <div class="component-group-desc">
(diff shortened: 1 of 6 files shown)


racquemis:
Currently there is no API command to fetch the location of the database on the 300 series.Also the telnet method used in nmj.py was removed on the 300 series and therefore i have to search for it. 

The code around line 104 has already been updated, see the commit. I still need to search for a specific string since  XML parsing returns a syntax error on the response.(however syntax looks correct) therefore i just look for a specific string for the time being.

I added the 300 ms delay between api calls in order to fix an issue i was having with the database scanner on the NMJ getting stuck.

racquemis:
I checked again and XML parsing does seem to be working. I don't understand why i constantly got a syntax error on it yesterday. New commit is up. Added XML parsing and error messages for the API returnvalues.

PR #454, opened 2012-08-15
Make the DB test actually delete the test database

depassp:
This test fails on Windows. We really should close the connection before deleting the db


thezoggy:
we use travis-ci, where the env is destroyed after running... so there really is no cleanup needed. but i'll bring this up to lad to see if there is a cleaner way to go about destroying the env

midgetspy:
Please submit this against the development branch and I'll pull it, thanks.

depassp:
OK, done
https://github.com/midgetspy/Sick-Beard/pull/455

PR #456, opened 2012-08-20
Fix unc path creation

EchelonFour:
If you use UNC paths then sickbeard fails when trying to create the missing folders. This patch fixes that.


ghost:
Did the pull request: https://github.com/midgetspy/Sick-Beard/pull/459

PR #458, opened 2012-08-21
Issue-1591 When loading from TVDB mark episodes with an invalid date as ...

thampton:
If we have no aired date mark it as unaired instead of skipped. The only time an aired date should be datetime.date.fromordinal(1) is when there was no date in TVDB. Malformed dates will be caught higher up in the script by a ValueError exception.

### DIFF
--- a/sickbeard/search_queue.py
+++ b/sickbeard/search_queue.py
@@ -145,7 +145,7 @@ class RSSSearchQueueItem(generic_queue.QueueItem):
         curDate = datetime.date.today().toordinal()
 
         myDB = db.DBConnection()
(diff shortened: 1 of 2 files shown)


thampton:
This worked for simple test cases I did. I will do some more testing.

thampton:
Itofzo let me know if you see any other issues.

thampton:
Closed this because after running this for awhile. I realized this is to much of a hack to push live. A complete fix to this needs to update the places stats are collected.  As well as changing the way unaired shows work in general. This is not as trival as it first appeared.

PR #466, opened 2012-09-04
Add "Archival Delay" configuration option.

mrmachine:
Don't download episodes in "initial" quality within a configurable number of days from the air date, IF "archival" quality is specified for a show AND other episodes have already been downloaded in archival quality.

This prevents multiple downloads when the archival quality is frequently released a short time after the initial quality. It won't prevent immediate download of episodes in initial quality for shows that are not released in archival quality.

You should now be able to specify a single custom quality setting (e.g. initial: SD DVD, HDTV; archival: 720p WEB-DL) that works for all shows without frequent duplicate downloads, and without having to know in advance which shows are going to be released in archival quality.

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -68,6 +68,17 @@
                             </label>
                         </div>
 
+                        <div class="field-pair">
+                            <label class="nocheck clearfix">
+                                <span class="component-title">Archival Delay</span>
+                                <input type="text" name="archival_delay" value="$sickbeard.ARCHIVAL_DELAY" size="5" />
+                            </label>
+                            <label class="nocheck clearfix">
+                                <span class="component-title">&nbsp;</span>
+                                <span class="component-desc">Number of days (after air date) to wait for archival quality before downloading initial quality. (eg. 2)</span>
+                            </label>
+                        </div>
+
                         <div class="clearfix"></div>
                         <input type="submit" class="btn config_submitter" value="Save Changes" /><br/>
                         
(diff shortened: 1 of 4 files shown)


id628:
midgetspy, any chance of approving this request?

I've recently been warned that I'm using more bandwidth than I'm allowed and I tracked it down to having sickbeard set to download HDTV but archive WEB-DL because I keep finding bad HDTV downloads.  So I basically download everything twice, once on release day, then again when the WEB-DL version comes out a day or two later.

I'd really rather wait 2-3 days for a known-good WEB-DL download, but sometimes that just doesn't come out.  So if the WEB-DL isn't available within 2-3 days, I'm willing to live with the regular or even SD file.

JoeyJoeJo:
This would be super awesome.
There are many more, and more importantly, _quicker_ WEB-DL releases than before.
Use case used to be: got the HDTV for speed of release... watch the HDTV... if you archive, get the WEB-DL days, weeks or even months later.

Nowadays, WEB-DL can be within hours of HDTV. From what I read, a lot of folks would give up a few hours or a day, or maybe two, to get a WEB-DL and not double up.

My prefs for most shows would be:
Web-DL at broadcast date
Web-DL at broadcast date +1
HDTV at broadcast date +2
SDTV at broadcast date +3
Web-DL as Archive if it ever comes up

mstrelan:
Wouldn't it make more sense to call this "Initial Delay"? To me, "Archival Delay" suggests it will wait a certain period before it archives an episode, regardless of whether it has been download initially.

mrmachine:
Rebased onto latest development branch... Any chance of a review?

mrmachine:
Rebased onto latest development branch again... Anyone want to review? It's still working well for me :)

kiddsupreme:
I know this will make me look like an idiot, but how do I pull these changes into my install of Sickbeard? Sorry if this has been answered already.

PR #468, opened 2012-09-07
Fix db upgrade when files are missing

twoolie:
DB upgrade crashes when trying to update filesize if file does not exist on disk.


thezoggy:
thank you for submitting this, however we have already resolved the issue in master.

PR #473, opened 2012-09-12
force the output of git to english

SyedAmerGilani:
without this change the output of git can not be parsed in an non english os

### DIFF
--- a/sickbeard/versionChecker.py
+++ b/sickbeard/versionChecker.py
@@ -234,13 +234,16 @@ class GitUpdateManager(UpdateManager):
 
         output = err = None
 
+        git_env = os.environ
+        git_env['LANG'] = 'en'
+
         for cur_git in git_locations:
 
             cmd = cur_git+' '+args
         
             try:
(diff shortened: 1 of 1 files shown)


sickbeard/versionChecker.py
```diff
@@ -234,13 +234,16 @@ def _run_git(self, args):
 
         output = err = None
 
+        git_env = os.environ
```
halorgium, line 237: This will change the `os.environ` for subsequent processes. 

``` python
git_env = os.environ
git_env['LANG'] = 'de_DE'
os.environ['LANG']
# 'de_DE'
```

Doing `git_env = os.environ.copy()` will not modify the existing values.

thezoggy:
i think this was officially taken care of in https://github.com/midgetspy/Sick-Beard/issues/511 -- can you please confirm?

SyedAmerGilani:
yes this fixes also the problem.

thezoggy:
glad to hear it

PR #476, opened 2012-09-17
added hdbits support

tsubus:
Support for the private tracker hdbits (based on tvtorrents code). Tested and seems to be working. All that the user needs to do is to enter the passkey in the settings and enable the tracker.


tsubus:
issues with repo, will reopen it

PR #480, opened 2012-09-19
Change delete_empty_folders add ek.ek, keep_dir, ignore_items

ghost:
Add ek.ek
Add keep_dir, don't delete show.location during postprocessing
Add ignore_items : treat folder empty when it only contains files, folders in ignore_items
(doesn't ignore anything yet)


kierr:
Is this merged into master yet? It's driving me nuts, shows getting deleted every week.

thezoggy:
yes

PR #481, opened 2012-09-19
minor cleanups

thezoggy:


SickBeard.py
```diff
@@ -1,328 +1,341 @@
-#!/usr/bin/env python
```
midgetspy: Endlines? This file needs to be \n so it can be executed directly on linux boxes.

thezoggy: yeah, looks like pyDev did that.. gotta go fix that. sorry

PR #484, opened 2012-09-23
On home page show number of days remaining till next episode airs

lloesche:
Changed the "Next Episode" column to display the number of days until the next episode, in addition to the date itself.

Code might require cleanup as this is the first time I ever do anything in Python.
Esp. the "if $airDateDelta.days > 1" part, which I would have done inline as (if foo > 1 ? 's' : '') in any other language, but found out Python doesn't support that Syntax.

Also the string formatting is debatable as it's no longer nicely centered.

But the intention of this patch is that the number of days remaining until the next episode airs is more meaningful than the air date itself where the brain has to do the calculation of the delta.

### DIFF
--- a/data/interfaces/default/home.tmpl
+++ b/data/interfaces/default/home.tmpl
@@ -144,10 +144,21 @@ $myShowList.sort(lambda x, y: cmp(x.name, y.name))
   #set $nom = 0
   #set $den = 1
 #end if
+#if len($curEp) != 0:
+  #set $airDateDelta = $curEp[0].airdate - datetime.date.today()
+  #if $airDateDelta.days > 1
+    #set $daysStr = 's'
+  #else
+    #set $daysStr = ''
+  #end if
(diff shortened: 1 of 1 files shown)


thezoggy:
if you are going to submit changes, use the development branch. and honestly i'd just say use a badge to show how many days.. `<span class="badge">1</span>`

thezoggy:
then also you have to realize if you change the airdate to something else.. you need to update the sorting algorithm and header.. plus i think a lot of people like seeing a date.. makes it easy to sort...

lloesche:
Oh, yes makes total sense not to work on the master branch. Sorry, was my first time working with git and github, still learning.

I saw that the timedelta Objects provide that str() output, but since the airdate only contains the date, not the time, the time display would always be '0:00:00'. So it was between cutting that substr off, or using the .days accessor and constructing my own 'day[s]' string.

Regarding the sorting, I didn't think about that, I'm gonna implement it.

What is a badge? I found the class definition of the CSS, but where is it being used? What does it look like?

Thanks!

PR #487, opened 2012-09-25
Minor naming tweaks

ghost:
Remove '.nzb' extension from release_name

Don't sanitize the parts, but the 'end results' 
http://code.google.com/p/sickbeard/issues/detail?id=2088

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -692,52 +692,52 @@ class PostProcessor(object):
             return True 
         
         return False
-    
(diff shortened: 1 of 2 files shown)


thezoggy:
should we worry about .zip, .rar, .gz, .nzb, or .torrent?

ghost:
Judging on this https://github.com/midgetspy/Sick-Beard/blob/master/sickbeard/postProcessor.py#L783
I don't think so.

PR #490, opened 2012-09-26
Option to prefer single episode releases over season releases

johnronk:
This option is for those who prefer to get each episode nzb added to sabnzbd instead of a complete season nzb. There are multiple reasons why one would want to use this option: Getting a whole season nzb might result in one downloading 100Gb of data only to have it fail in (par) verification. The impact is much less severe when each episode is downloaded as singles. Another reason is if you, like me, prefer the original scene releases with .nfo files over some random season nzb.

As I figure this is not a main stream option it has not been added to the user interface and is currently only available in "config.ini". The default value is "False". 

Commit log:
- Added option "prefer_episode_releases" to config.ini. If a complete season of a show is added to the backlog this option will make sickbeard prefer single episode releases for that season over a comlete season release. I.e. if the complete season 1 of the show "Tv Show" is requested sickbeard will prefer "Tv.Show.S01E01...", "Tv.Show.S01E02...", e. over "Tv.Show.S01...". If not all episodes are available as singles the missing episodes will be picked individually from the season release. Though, "Proper"/"Repack" from season release will be preferred over non-"Proper"/"Repack" singles.
- Added some logic to "pickBestResult" so that if the current "bestResult" already is a "Proper"/"Repack" it will not be overridden by another "Proper"/"Repack" further down the list. This also ensures that "Proper"/"Repack" single episodes always will be preferred over "Proper"/"Repack" season releases (if the "prefer_episode_releases" option is activated).

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -144,6 +144,7 @@ NZB_METHOD = None
 NZB_DIR = None
 USENET_RETENTION = None
 DOWNLOAD_PROPERS = None
+PREFER_EPISODE_RELEASES = None
 
 SEARCH_FREQUENCY = None
 BACKLOG_SEARCH_FREQUENCY = 21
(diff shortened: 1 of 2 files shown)


thezoggy:
sb only grabs a season rip if you asked for a season... or if you asked for an episode and the only thing it could find is a season rip.. (then it grabs the season rip and tries to part out the nzb it needs --only supported for newznab providers).

thezoggy:
also, this pull request would need to be based off development not master branch.

johnronk:
Yes, I know that sb only grabs a season rip in the scenarios you describe. This option makes single episode nzbs preferred over a season nzb for those use cases (and possible missing singles will be retrieved from the season rip - but only with newznab providers). As I wrote this is not a main stream option, but I know multiple people who is very interested in this option.

I'll correct the pull request as you requested.

johnronk:
Closed in favor of pull request #491 which is based of the development branch.

PR #491, opened 2012-09-27
Option to prefer single episode releases over season releases

johnronk:
This option is for those who prefer to get each episode nzb added to sabnzbd instead of a complete season nzb (when all episodes in a season are wanted). There are multiple reasons why one would want to use this option: Getting a whole season nzb might result in one downloading 100Gb of data only to have it fail in (par) verification. The impact is much less severe when each episode is downloaded as singles. Another reason is if you, like me, prefer the original scene releases with .nfo files over some random season nzb.

As I figure this is not a main stream option it has not been added to the user interface and is currently only available in "config.ini". The default value is "False".

Commit log:
- Added option "prefer_episode_releases" to config.ini. If a complete season of a show is added to the backlog this option will make sickbeard prefer single episode releases for that season over a comlete season release. I.e. if the complete season 1 of the show "Tv Show" is requested sickbeard will prefer "Tv.Show.S01E01...", "Tv.Show.S01E02...", over "Tv.Show.S01...". If not all episodes are available as singles the missing episodes will be picked individually from the season release (only supported for newznab providers). "Proper"/"Repack" from season release will be preferred over non-"Proper"/"Repack" singles (newznab).
- Changed logic in "pickBestResult" so that if the current "bestResult" already is a "Proper"/"Repack" it will not be overridden by another "Proper"/"Repack" further down the list. This also ensures that "Proper"/"Repack" single episodes always will be preferred over "Proper"/"Repack" season releases (if the "prefer_episode_releases" option is activated).

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -144,6 +144,7 @@ NZB_METHOD = None
 NZB_DIR = None
 USENET_RETENTION = None
 DOWNLOAD_PROPERS = None
+PREFER_EPISODE_RELEASES = None
 
 SEARCH_FREQUENCY = None
 BACKLOG_SEARCH_FREQUENCY = 21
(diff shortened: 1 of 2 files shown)


gaetan-petit:
Maybe SickBeard should also fallback to this behavior when no season nzb is found.

johnronk:
@gaetan-petit: I'm not sure what you mean. This patch makes SickBeard behave as if no season nzb is found (except for the cherry picking part if not all episodes are found as single releases).

gaetan-petit:
@johnronk Yeah I get this, what I said is that this behvior should be a default fallback in "season mode" when no season nzb could be found.

johnronk:
Hmm, this might have to do with what kind of  search provider you're using? I'm using a newznab provider, and for me this is currently the default fallback. Example: If a complete season is in the backlog, SickBeard checks to see if there exists a nzb for the complete season with the required quality. If this is the case then it downloads this nzb. If not it searches for single release nzbs for all episodes in the season and adds them to the newsreader.

an0Nym0us63:
is this working for multiple episodes nzbs. such as for example ; bones 2X12-2X15

will be sent to sab as bones 2X12  
bones 2X13
bones 2X14
bones 2X15 ??

mano3m:
Would love this option. I actually always do the same thing manually, by first setting half of the episodes as wanted, then the other half...

simonk83:
My only issue with it grabbing full seasons is that it still doesn't clean up the folder after processing.

junalmeida:
Will test this and merge to my torrent branch

NewYears1978:
What is the setting?  Also when using sickBeard on unraid server....where is the config file as I can only locate "sickbeard.cfg" not "config.ini"

Every time SB tries to DL a full season it fails so I really need this option.

PR #492, opened 2012-09-29
Add update once option to xbmc notifier

EchelonFour:
Adds option to stop sending update commands to xbmc hosts once it has successfully updated once. If using a mysql database that has several clients, this is really useful.

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -66,6 +66,13 @@
                                 </label>
                             </div>
                             <div class="field-pair">
+                                <input type="checkbox" name="xbmc_update_once" id="xbmc_update_once" #if $sickbeard.XBMC_UPDATE_ONCE then "checked=\"checked\"" else ""# />
+                                <label class="clearfix" for="xbmc_update_once">
+                                    <span class="component-title">Update Only One Host</span>
+                                    <span class="component-desc">Update only one host in the list? (Useful for shared databases)</span>
+                                </label>
+                            </div>
+                            <div class="field-pair">
                                 <label class="nocheck clearfix">
                                     <span class="component-title">XBMC IP:Port</span>
                                     <input type="text" name="xbmc_host" id="xbmc_host" value="$sickbeard.XBMC_HOST" size="35" />
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -203,6 +203,7 @@ XBMC_NOTIFY_ONSNATCH = False
 XBMC_NOTIFY_ONDOWNLOAD = False
 XBMC_UPDATE_LIBRARY = False
 XBMC_UPDATE_FULL = False
+XBMC_UPDATE_ONCE = False
 XBMC_HOST = ''
 XBMC_USERNAME = None
 XBMC_PASSWORD = None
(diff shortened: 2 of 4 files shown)


thezoggy:
not understanding the need for this.. why not just put one host in the xbmc notify box?

EchelonFour:
Because it might not be turned on, but another xbmc front-end might be.
I have a couple of xbmc front-ends connected to the same server, sometimes 2 are turned on at the same time, and sometimes only one of the two are turned on. If they both update at the same time then I get duplicates in the database, and if only one is listed in sickbeard then the other wont update if it is the only one turned on.

thezoggy:
honestly i'd rather just see a rewrite of how we store xbmc hosts.. 
store the hostname and its associated params.. kinda like how we do the metadata. 
so that way we could parse that hostname and know that it is not a primary server, wants to be included on update, dont include on whatever, yes for whatever. 
hostname|0|1|0|1

i would think this would be much more flexible than people keep trying to extend and hack the current method. 
and then for presentation purposes in the config ui.. we'd show a drop down of saved servers.. using opt-group to divide stuff into master/slave servers.. 

but anyways.. this is all over kill for most as most of the time.. they only have 1 server. and the sql method of xbmc honestly isnt that common for most..

thezoggy:
http://wiki.xbmc.org/index.php?title=HOW-TO:Sync_multiple_libraries/Setting_up_XBMC

EchelonFour:
The problem is that there aren't really any master or slave servers. There is one master mysql server and several xbmc slaves, but only xbmc can do library updates, so sickbeard can do nothing to the server.

thezoggy:
im working on a xbmc notifier rewrite.. to add the json support. for now im just going to have it only send the update command to the first host. the notify functions will send to all hosts. later on we can come back to adding/changing the options and how we store the data so that hosts can have different username/pw/options.

thezoggy:
when i rewrote the xbmc notifier for frodo support i made it so that the library update commands only go to the first host. (this is currently in the development branch if you wanna test)

i do still plan to re-visit how we store/do everything with regards to multi host setup.. but for now this should cover most people. so this pull request would not only have a bit of conflicts due to all the code changes.. but really isnt needed any longer

EchelonFour:
ok, but I still think the functionality of trying to update them in order until successful would be good (if the first host is off for instance)

thezoggy:
while yes i would agree that feature is nice.. but wouldnt that cause quite unpredictable results if one does a master/slave setup where the slave should never be updating anything..  you use sql backend with xbmc?

EchelonFour:
yes i have a mysql backend. however the server itself is incapable of updating the library, it's just a data store. so each of the xbmc frontends connect to the server and can update the library. There is no "slave" that is incapable of updating the server. This may change in the future with the upnp library sharing stuff in the latest xbmc builds, but it is not finished yet

Skindred:
I can see the following scenario where having a list of 'Library updaters' might be usefull.

Say, I have laptop and a Raspberry Pi, both running XBMC and my Synology happily serving the MySQL library database.
Updating the library takes quite some resources from the RPi, but if the laptop were on,
and XBMC were running, it would be faster and handier to have the laptop perform the update,
instead of the RPi.

This way, I can enjoy watching my content on my RPi, and not having it bogged down (or even crashing) by library updates.

PR #494, opened 2012-10-02
Roksbox (Roku) metadata support

fhriley:
This adds metadata support for the Roksbox channel on the Roku.


thezoggy:
this should be based off development as all work goes into that first.

thezoggy:
this looks like its no longer needed due to all the metadata changes made recently. can you confirm that one of the other metadata targets works for this purpose just fine? (i suspect Mede8er structure is what you need)

PR #496, opened 2012-10-03
Fix files with mixed line endings

mjc:
These files have both CRLF and LF line endings.
Mixed line-endings confuse merge tools and text editors.

This commit converts them to CRLF only.

I did this via:

find . -type f -print0 | xargs -0 file | grep ", LF" | cut -d : -f 1 | xargs unix2dos

file(1) output on my system for files witih both CRLF and LF looks like this:

./sickbeard/postProcessor.py:                                           Python script, ASCII text executable, with CRLF, LF line terminators


thezoggy:
many IDE can auto correct the line ending for the whole repo. takes little effort.. but creates so much noise.. and loses git blame history

mjc:
You still end up with incorrect git blame by doing it over time, and it's a bigger problem because you can lose the useful parts of the commit if you don't realize you need to give diff the "ignore whitespace" option.  

You can make a tag for the commit prior to the line ending conversion and do "git blame pre-lf", if need be.

The IDE thing is relatively minor but the fact that diff tools get confused by mixed endings at times isn't.

I'll poke around in #git to see if they've any idea how to preserve blame history, or ideas on how to begin to whip up a script to map it or something.

thezoggy:
i think you are missing the point about rewriting the repo.. people have tried (i've even tried myself at one point.. with pep8/pyflakes fixes) and it got rejected by midget.

PR #501, opened 2012-10-05
Fix header from displaying funny on iOS (only browser that had a problem)

thezoggy:


piejanssens:
Nice to see it's fixed.

PR #506, opened 2012-10-12
Fixed the Trakt notifier to use basic authentication

jaredhocutt:
The Trakt notifier was using POST authentication which does not appear to work even though the Trakt API documentation says it is supported. I modified the code to use basic authentication instead so that the Trakt notifier will authenticate and work.

### DIFF
--- a/sickbeard/notifiers/trakt.py
+++ b/sickbeard/notifiers/trakt.py
@@ -17,7 +17,7 @@
 # along with Sick Beard.  If not, see <http://www.gnu.org/licenses/>.
 
 
-import urllib2
+import urllib2, base64
 
 from hashlib import sha1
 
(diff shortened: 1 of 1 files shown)


fernandog:
"The Trakt notifier was using POST authentication which does not appear to work even though the Trakt API documentation says it is supported."

still don't working?

thezoggy:
new code goes into development branch not master. plz update pull request to reflect that

jaredhocutt:
Just checked and it appears to work again. Seems that Trakt may have been having issues with accepting username/password as POST parameters at the time I made this pull request. It still probably wouldn't be a bad idea to use basic authentication. Would you like for me to still update the pull request for the development branch or go ahead and close this pull request?

PR #509, opened 2012-10-17
rewrite of saveSearch()

bradallred:
I work on a fork of SickBeard that adds a lot of search options (~40 with 4 additional download methods) and these are a couple of changes I have made to prevent saveSearch() from becoming unwieldy.

### DIFF
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -94,32 +94,32 @@ def change_NZB_DIR(nzb_dir):
 
     if nzb_dir == '':
         sickbeard.NZB_DIR = ''
-        return True
+        return True, u"Removed NZB directory"
 
     if os.path.normpath(sickbeard.NZB_DIR) != os.path.normpath(nzb_dir):
         if helpers.makeDir(nzb_dir):
             sickbeard.NZB_DIR = os.path.normpath(nzb_dir)
             logger.log(u"Changed NZB folder to " + nzb_dir)
         else:
-            return False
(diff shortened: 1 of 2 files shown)


midgetspy:
Great submission, thanks.

thezoggy:
also this change breaks the page as you cant uncheck and save... as something that isnt checked off isnt passed. thus this needs to be reverted

thezoggy:
https://github.com/midgetspy/Sick-Beard/pull/536 reverts pull #509

bradallred:
what exactly is the problem with saving? it is possible that i mucked something up when backporting, but everything behaves as expected for me. (except for the regex which as noted doesnt work with a host that starts with 's' if the protocol is omitted)

what is the expected result of saveSearch() so that i can resubmit something that works for you? did you even test this?

thezoggy:
go tell sb to use torrents save check config and you see `use_torrents = 1` now go uncheck it and hit save. you will see it is still 1 not 0. this is because when you dont have a checkbox checked its omitted from post.. which is why we explicitly define what is expected to be passed and default to 0/none if its not.. also with your changes i noticed that the save/test buttons for sab resulted in my sab host turning into `http://http://http://http://http://http://localhost:8082/` overall there was just too many issues to deal with and we reverted the changes until it could be properly looked into

bradallred:
ah i think i see what happened. In my fork i think I had modified the html so ensure that a space was always submitted and i dealt with that using the trim call you see there. I obviously failed to realize that when i back-ported this.

PR #514, opened 2012-10-25
Update Xbmc notifier to use JSON for new Xbmc versions(>=frodo)

wsnipex:
Xbmc recently dropped the old and long deprecated HTTP API, so notifications and library updates don't work anymore in xbmc nightly builds.

this PR adds methods for the JSONRPC API and uses it if a recent Xbmc version is found.
For older Xbmc builds, nothing changes.
We cannot use json for older xbmc builds, because they don't support notifications over json.

### DIFF
--- a/sickbeard/notifiers/xbmc.py
+++ b/sickbeard/notifiers/xbmc.py
@@ -34,26 +34,55 @@ try:
 except ImportError:
     import xml.etree.ElementTree as etree
 
-class XBMCNotifier:
+try:
+    import lib.simplejson as json
+except:
+    import json
 
+class XBMCNotifier:
+    
+    apiversion = 0
+    
     def notify_snatch(self, ep_name):
         if sickbeard.XBMC_NOTIFY_ONSNATCH:
-            self._notifyXBMC(ep_name, common.notifyStrings[common.NOTIFY_SNATCH])
+            if (self._jsonapiversion() <= 4):
+                logger.log(u"Detected XBMC version <= 11, using XBMC HTTP API", logger.DEBUG)
+                self._notifyXBMC(ep_name, common.notifyStrings[common.NOTIFY_SNATCH])
+            else:
+                logger.log(u"Detected XBMC version >= 12, using XBMC JSON API", logger.DEBUG)
+                self._notifyXBMC_json(ep_name, common.notifyStrings[common.NOTIFY_SNATCH])
 
     def notify_download(self, ep_name):
         if sickbeard.XBMC_NOTIFY_ONDOWNLOAD:
-            self._notifyXBMC(ep_name, common.notifyStrings[common.NOTIFY_DOWNLOAD])
+            if (self._jsonapiversion() <= 4):
+                self._notifyXBMC(ep_name, common.notifyStrings[common.NOTIFY_DOWNLOAD])
+            else:
+                self._notifyXBMC_json(ep_name, common.notifyStrings[common.NOTIFY_DOWNLOAD])
 
     def test_notify(self, host, username, password):
-        return self._notifyXBMC("Testing XBMC notifications from Sick Beard", "Test Notification", host, username, password, force=True)
+        if (self._jsonapiversion() <= 4):
+            logger.log(u"Detected XBMC version <= 11, using XBMC HTTP API", logger.DEBUG)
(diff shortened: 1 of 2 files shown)


thezoggy:
while this looks good, it will pose a problem for plex as it subclasses the xbmc notifier...  need to see if plex does json as well now.. or its just going to be best to make plex notifier have its own code going forward. (and maybe boxee?)

wsnipex:
as talked on IRC

```
def notify_download(self, ep_name):
    if sickbeard.PLEX_NOTIFY_ONDOWNLOAD:
        self._notifyXBMC(ep_name, common.notifyStrings[common.NOTIFY_DOWNLOAD])
```

it calls the private method, which is unaltered and always uses the old http api. So it should still work.
That said, I cannot test it, because I refuse to install plex ;)

jonnywilliamson:
Thank you for doing this!

thezoggy:
just tried this out with latest nightly.. filled out my info and hit 'test'.

```
00:55:33 ERROR::CP Server Thread-4 :: Warning: Couldn't contact XBMC JSON api at http://192.168.0.4:8080/jsonrpc: [Errno 10061] No connection could be made because the target machine actively refused it
00:55:33 INFO::CP Server Thread-4 :: Warning: Couldn't contact XBMC HTTP server at 192.168.0.197:9090:
```

noticed that it says it tried what is actually in the box but what was actually sent is whats been saved. -- this should be addressed so people dont have to keep trying new values and submitting.. also we should change the default port from 8080 to 9090 going forward...
now that the correct info was saved.. i tried hitting test again..

```
00:55:33 INFO::CP Server Thread-4 :: Warning: Couldn't contact XBMC HTTP server at 192.168.0.197:9090:
00:56:05 ERROR::CP Server Thread-4 :: Warning: Couldn't contact XBMC JSON api at http://192.168.0.197:9090/jsonrpc:
```

Loaded up xbmc and checked it's debugging:

```
00:57:11 T:3860   DEBUG: JSONRPC Server: New connection detected
00:57:11 T:3860    INFO: JSONRPC Server: New connection added
00:57:11 T:3860    INFO: WebSocket: missing Sec-WebSocket-Version
00:57:11 T:3860    INFO: JSONRPC Server: Disconnection detected
```

Tried it manually.. looks like the json api is currently borked in the nightlies. :(
Will test again tomorrow.

wsnipex:
Are you trying to use the event server? this code still uses HTTP, so you need to configure the webserver port.
I don't think websockets have anything to do with this.

thezoggy:
So appears i was confused on my part, as I was testing with a fresh install I saw 'HTTP' and knew this was the old http api method.. so i didnt enable that.. which turns out is now the new http json method. 

After enabling the webserver,
HTTP: In System/Settings/Network/Services activate Allow control of XBMC via HTTP
Also change port to 8080 or something higher than 1024 and not 80.

This pull request does work. Some minor logging issues (missing spaces / reporting back what its using a little over verbose at times / not using unicode for username/pw) to be addressed before being pulled. Also I need to merge this with my xbmc cleanup.

thezoggy:
ok after going through and cleaning up things i realize that the logic to know which branch of code to run (http api vs json api) is done by checking the host-string.. this works if theres only one host.. but not if there are multiple or if a mix match of xbmc versions. then there's also a lot of duplicate code.. and unneeded complexity. going to have to spend some time to move things around and not buffer apiversion.. could just try both.. but really this needs to be cleaned up greatly... also thinking about just only doing the update library for the first host.. that would solve multi xbmc setups (sql) with a slight hack as they could have their 'server' first and then their clients afterwards.

wsnipex:
it does check the JSON API Version string. This works as long as all hosts are >=frodo. As you pointed out, it doesn't with mixed eden and frodo setups, because its cached. Its easy to just disable the caching, although I think that those mixed multi-host setups are rare. Most ppl run mysql anyway, if they have multiple hosts.

I duplicated some code on purpose, because I didn't want to mess with the existing http code. Those methods can probably be dropped in a year or so, when ppl have migrated to newer versions(e.g. update library over json also works in Eden, but dunno about Plex).

Anyway, feel free to change anything to your liking, I just wanted to make sure the functionality is there when Frodo gets released.

wsnipex:
superseded by https://github.com/midgetspy/Sick-Beard/pull/524
thx @thezoggy

PR #515, opened 2012-10-27
Change how Change Show cycles through shows

flippingtables:
Small change to displayShow.js that makes it possible to cycle through
shows by clicking Previous/Next. Automatically jumps to the First or
Last show in the list if it reaches the end.


midgetspy:
Good idea, thanks!

PR #517, opened 2012-10-28
Added JSON fallback functions for XBMC, in case HTTP API fails or is non...

Polydwarf:
...-existent

This is a new set of functions in the XBMC notification code for using the XBMC JSON API to do communications.  Recently, XBMC has pulled out the HTTP API in their development code, so anyone running nightlies/.git builds of XBMC hasn't been able to get SickBeard to automatically update libraries.

The code is written so that it will try the HTTP API first, then go to JSON if the HTTP API fails.  This way, people that aren't on XBMC nightlies will continue to use the same code they always have.

There was one tweak I made that affected the HTTP API code... The XBMC Test Notification code didn't return a True/False like the rest of the notiifiers; it returned the http that came back, and then the calling code parsed it out.  I just moved the parsing of the return string into the notifier code so it could return True/False, and updated webserve.py to expect a True/False out of it.

In doing all this, I did find a bug in the development code of XBMC that actually makes this somewhat useless at the moment... http://forum.xbmc.org/showthread.php?tid=143740 which led to http://trac.xbmc.org/ticket/13462 ... Until they get this bug resolved, you still need to run the auto-clean.  Oh well, not much I can do about that.

All I can say is that the JSON code works on the ubuntu nightly ppa XBMC.  Anything else probably shouldn't matter, due to JSON being used only if HTTP API fails.

### DIFF
--- a/sickbeard/notifiers/xbmc.py
+++ b/sickbeard/notifiers/xbmc.py
@@ -22,6 +22,10 @@ import socket
 import base64
 import time, struct
 
+try: import simplejson as json
+except ImportError: import json
+
+
 import sickbeard
 
 from sickbeard import logger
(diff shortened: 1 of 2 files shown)


Polydwarf:
I see someone beat me to the punch and is working with Zoggy... I'll leave this here, just in case.  Otherwise, close away.

thezoggy:
thanks for the submission, slightly different approach than the branching code tactic wsnipez used. i've found issues in his code and will spend some time tonight rewriting things.. i'll reference your submission as well.

Polydwarf:
FYI, there are a couple of annoyances in my code..

Those of us on bleeding edge (ie, we need the JSON interface) get double hits to XBMC (first hit for HTTP API, which fails, then the JSON calls).  On a local network, I pretty much thought "eh, who cares", but it does show up in the XBMC debug logs as a hit to the http server.  I always pictured that once the official release of XBMC goes out with the HTTP API pulled, the order of functions would get reversed so that JSON is attempted first, and then down the line, the HTTP functions would get removed entirely.

Due to that double hit, some extra logging happens in SickBeard due to the HTTP API library update function throwing off a logging.ERROR message in the event of a failure.  It's something I had meant to refactor a little bit (moving the Error logging up to the externally called function), but just plain forgot about.

ghost:
Now that I've started using XBMC Frodo with building OpenELEC [1] from source, I would really love to see this merged sooner, rather than later. But I understand that stuff needs to pass the inspection. Friendly reminder, XBMC Frodo has release 11/24/12 [2]. So maybe a merge before it hits stable ;)

Good luck!

[1] http://openelec.tv/
[2] http://trac.xbmc.org/milestone/12.0

thezoggy:
the dates on releases generally are nothing but a guide line.. not firm. if you look at past releases they never hit them :)
but the xbmc/plex notifier rewrite is pretty much done, still got a little more testing to do on it before i'll submit it

ghost:
Have noticed that a couple of times ;) Rules are ment du be broken, right :P

Good :) Thanks!

rjanja:
What's the status on the XBMC notification changes? I see @Polydwarf put a lot of effort into this, so did @thezoggy using wsnipex's work as a base, but there's no pull request for the latter. Sickbeard hasn't been notifying XBMC for over a month now, and I'm getting a little antsy. Is there anything I can do to help with this effort?

thezoggy:
thats because this pull request was closed in lieu of mine which was referenced earlier, #524

rjanja:
@thezoggy: Oh jeez, I don't know how I missed that. Sorry.

And thanks for all your work on SickBeard!

thezoggy:
np, the frodo related work is in development branch and should be pushed out to the master branch when we are ready to make a new build. frodo is still a few weeks from being released and they are still making changes (so hopefully none of the code changes i've made will break before they release it)

PR #521, opened 2012-10-30
Update sickbeard/scene_exceptions.py

tmiland:
Changed url to exeptions.txt
Error in sickbeard: 2012-10-30 09:24:54.293085 CHECKVERSION :: Check scene exceptions update failed. Unable to get URL: http://midgetspy.github.com/sb_tvdb_scene_exceptions/exceptions.txt

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -68,7 +68,7 @@ def retrieve_exceptions():
     exception_dict = {}
 
     # exceptions are stored on github pages
-    url = 'http://midgetspy.github.com/sb_tvdb_scene_exceptions/exceptions.txt'
+    url = 'http://github.com/midgetspy/sb_tvdb_scene_exceptions/blob/gh-pages/exceptions.txt'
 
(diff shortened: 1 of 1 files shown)


tmiland:
Sorry about that, my bad. It's just that i get this error all the time, even now.

thezoggy:
yeah, ive started to see it pop up myself.. its wierd too because in the error log we see a doctype being passed which doesnt exist on the actual page.. so it makes me think github hiccups and serves up some page in the middle of it loading which we puke on trying to parse.

thezoggy:
did some research on this tonight:

There is a long-standing problem in httplib where it fails to handle a particular HTTP 1.1 protocol error. The symptom is a transfer that fails with the error  ValueError: invalid literal for int()  The problem is discussed here: http://bugs.python.org/issue1486335. Apparently it is fixed for Python 2.6: http://bugs.python.org/issue900744. Meanwhile a workaround is to use urllib.urlopen() which does not have the same problem.

PR #522, opened 2012-10-31
Update sickbeard/helpers.py

srdjanrosic:
- Simplify fileBitFilter
- If computed child mode is sticky, remove it since it doesn't work on files anyway, and remove the other writable bit since chances are people will be running with the setgid bit on their media directory anyway.


srdjanrosic:
same as before, just rebased on top of current master.

thezoggy:
this needs to be based off the development branch as all new code goes there first

PR #524, opened 2012-11-07
Notifiers: XBMC / Plex - cleanup and update

thezoggy:
- XBMC notifier was completely redone to support both the legacy http api and the upcoming xbmc12 (frodo) json-rpc method. Using the public calls as wrappers with the detection to know which internal methods to call.
- Due to the XBMC rewrite, the Plex notifier no longer inherits xbmc's notifier as the code has diverged too much -- plus we were wrongly showing 'xbmc' log messages.
- Cleaned up many small bugs I found in the process as well as updated the config ui notification page.
- Cleaned up the notifiers page to match the web ui, which made it easier to test and for others to add notifiers in the future.


aaearon:
With XBMC the 20121108 3b7a2ab nightly, was able to send a test notification to XBMC successfully.

ZestyChicken:
This works great. Well done. When will this be added to the master branch?

thezoggy:
when we are ready to push out a new build..  not too much longer.. but we have some bugs to fix by other recent code submissions first

PR #530, opened 2012-11-17
Fixed issue with UNKNOWN status episodes being ignored

CaptainRewind:
While this doesn't address the root cause of episode statuses showing up as UNKNOWN, it handles the condition should it arise by checking episodes of UNKNOWN status that have an airdate of today or in the past.  Please pull in these changes.
Thanks!

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -861,15 +861,19 @@ class TVShow(object):
             return False
 
         myDB = db.DBConnection()
-        sqlResults = myDB.select("SELECT status FROM tv_episodes WHERE showid = ? AND season = ? AND episode = ?", [self.tvdbid, season, episode])
+        sqlResults = myDB.select("SELECT status,airdate FROM tv_episodes WHERE showid = ? AND season = ? AND episode = ?", [self.tvdbid, season, episode])
 
         if not sqlResults or not len(sqlResults):
             logger.log(u"Unable to find the episode", logger.DEBUG)
             return False
 
         epStatus = int(sqlResults[0]["status"])
+        epAirDate = int(sqlResults[0]["airdate"])
+        todayDate = datetime.date.today().toordinal()
 
         logger.log(u"current episode status: "+str(epStatus), logger.DEBUG)
+        logger.log(u"current episode airdate: "+str(epAirDate), logger.DEBUG)
+        logger.log(u"current date: "+str(todayDate), logger.DEBUG)
 
         # if we know we don't want it then just say no
         if epStatus in (SKIPPED, IGNORED, ARCHIVED) and not manualSearch:
@@ -880,6 +884,9 @@ class TVShow(object):
         if quality in anyQualities + bestQualities:
             if epStatus in (WANTED, UNAIRED, SKIPPED):
                 logger.log(u"Ep is wanted/unaired/skipped, definitely get it", logger.DEBUG)
+                return True
+            elif epStatus == UNKNOWN and epAirDate <= todayDate:
+                logger.log(u"Still in UNKNOWN status for some reason, but it should have aired, so getting it", logger.DEBUG)
                 return True
             elif manualSearch:
                 logger.log(u"Usually I would ignore this ep but because you forced the search I'm overriding the default and allowing the quality", logger.DEBUG)


thezoggy:
i think the proper thing to tackle is why something is unknown status in the first place. not much in searching for an episode that never aired/malformed date. we want to reduce the chances of false positives.. dont want sb grabbing episodes that arent schedule to come out years in advance or if someone does something really stupid on tvdb

also, all new code goes into the development branch first, not master.

CaptainRewind:
My changes do check the air date against the current date, of course as 
you mentioned that relies on the air date being correct in TVDB. The 
issue has been affecting me so thought I'd try to fix it.

I'm guessing I should instead try to figure out why that's happening? 
Once the episodes get set as UNKNOWN it seems there's no going back, and 
I'm getting it on several shows.

thezoggy:
in my tvdb fixes branch ( https://github.com/midgetspy/Sick-Beard/pull/527 ) we now treat 0000-00-00 the same thing as if no date was set. As this is a malformed date to try and parse, which caused issues on during the tvdb-tvrage sync process. not sure if that would change anything for your issue though

CaptainRewind:
I'll keep digging into the issue I'm seeing, and remember to push to development branch next time when I find more.

thezoggy:
is your clock set wrong?

CaptainRewind:
No. I'm running Ubuntu 10.04, and using ntp. The clock is always right.  I have this problem with several shows, not just one or two. 
Basically, the unaired episodes start showing up as Unknown in the UI and sickbeard.db (instead of Unaired), and as such SB never runs a search for them.  The air date shows up fine for any shows having this condition (which is why my code changes check that the air date is in the past or today).  Also, it seems to be "show-wide," meaning that if a show has the problem, then all future and current episodes for that show have the problem (as opposed to one or two).  Even after deleting and re-adding a particular show with this condition, the same issue occurs.
The changes I submitted are working fine when I run them locally.  I would agree that root cause analysis would be needed to address why this is happening.  Before I changed this code, any Unaired episodes that appeared as Unknown I was manually changing to "Wanted". This manual process was getting cumbersome so this is what I came up with as a fix.
It seems I'll need to learn more about SB's behavior when it sets the status to begin with.  Perhaps I could write a method to go through and periodically clean up shows in this status (though if it doesn't run frequently, things would be missed until it runs).  That still doesn't address why this happens in the first place, and since it's so random tracking that down may prove difficult. 
Any ideas are welcome... I'm no Python expert, but this was/is a good learning experience.

thezoggy:
i dont have the unknown problem myself.. have you checked sb logs?

CaptainRewind:
Yes, and in looking through them again after this thread, I think the root problem has to do with having the config value of "add_shows_wo_dir = 1".  Setting to 0 gets the next-aired show to show up as "Unaired".  However, even when I have it set to 0, all future episodes that TVDB shows as airdate of "never" get status in SB set to "Archived". I believe this has to do with having "Set the initial status of missing episodes" set to "Archived" when first adding the show.

I'm still playing around with these settings, but I bet that if I have sb create the folder and mark initial statuses as Skipped, things should get much better for my situation.  This seems like a good short-term workaround.

Unfortunately, this is not the desired use-case for me.  I don't want a folder created or used, and want all PAST episodes to be Archived when I add a new show, but all FUTURE episodes to be marked as SKIPPED until the TVDB airdate is updated properly (e.g. has a real date, not "never" (which I'm guessing is the 0000-00-00 case that you mentioned above)).  Marking the future episodes (those without a TVDB airdate (or one of 0000-00-00)) as SKIPPED is necessary for SB to pick them up once the airdate is updated on TVDB.  Because mine aren't in the show's folder (because I don't have or want one), I think this is the very heart of my issue.

I'm still brainstorming how this will be possible.

thezoggy:
yes thats a known issue with that option.. theres also another bug with it.. ill have to dig in my notes... but main thing is that sb internally relied on having a valid show folder before it would download something as it didnt make sense to d/l something if the folder didnt exist as it would fail pp.

thezoggy:
```
0106-16-2012@[06:04am] <@_zoggy_> create_missing_show_dirs = creates shows dir if its missing
0106-16-2012@[06:04am] <@_zoggy_> add_shows_wo_dir = makes it so that sb does not create a show dir when adding 
0106-16-2012@[06:04am] <@_zoggy_> so with both those set to 1 in the ini. 
0106-16-2012@[06:04am] <@_zoggy_> add a show. nothing happens (just info stored in db.) when you go to pp a file.. it will create the missing dir.
0106-16-2012@[06:05am] <@_zoggy_> if you have one set to 0 and the other set to 1. some stuff works.. some stuff doenst.. 
0106-16-2012@[06:05am] <@_zoggy_> just depending on what your trying to do
0106-16-2012@[06:05am] <@_zoggy_> why i said.. it makes no sense to have two seperate options.. since they go hand in hand
```

```
0105-06-2012@[06:17am] <_zoggy_> so, if you have create_missing_show_dirs = 1 set and on the displayShow page if the folder doesnt exist.. you dont see any warning
0105-06-2012@[06:23am] <_zoggy_> ok shutting sb down, putting that back to 0, now trying out add_shows_wo_dir = 1
0105-06-2012@[06:24am] <_zoggy_> hmm wonder what happen if you add a show with no root dir set...
0105-06-2012@[06:25am] <_zoggy_> ok well add_shows_wo_dir =1 works, you get a shitload of log spam because the dir doesnt exist though 
0105-06-2012@[06:25am] <_zoggy_> 06:24:45 INFO::SHOWQUEUE-ADD :: The show dir is missing, not bothering to change the episode statuses since it'd probably be invalid
0105-06-2012@[06:27am] <_zoggy_> so with 'add_shows_wo_dir = 1' on displayShow you do get the msg about it missing
0105-06-2012@[06:27am] <_zoggy_> Location:     D:\Sickbeard_TV\Ugly Americans (dir is missing)

0105-06-2012@[06:28am] <_zoggy_> got the banner/poster cache images
0105-06-2012@[06:28am] <_zoggy_> so far it seems to be working
0105-06-2012@[06:28am] <_zoggy_> now let me try to manually pp
0105-06-2012@[06:29am] <_zoggy_>     if not ek.ek(os.path.isdir, ep_obj.show.location) and sickbeard.CREATE_MISSING_SHOW_DIRS:
0105-06-2012@[06:29am] <_zoggy_>   File "C:\Program Files (x86)\Sick-Beard\sickbeard\tv.py", line 92, in _getLocation
0105-06-2012@[06:29am] <_zoggy_>     raise exceptions.ShowDirNotFoundException("Show folder doesn't exist, you shouldn't be using it")
0105-06-2012@[06:29am] <_zoggy_> ShowDirNotFoundException: Show folder doesn't exist, you shouldn't be using it
0105-06-2012@[06:29am] <_zoggy_> ok yeah, so wont do it if the other option isnt enabled
0105-06-2012@[06:29am] <_zoggy_> so not really sure why there are two options.. really should be one and the same.. i cant think of an application you'd only want one and not the other
0105-06-2012@[06:29am] <_zoggy_> create_missing_show_dirs /  add_shows_wo_dir
```

testing:

```
0105-06-2012@[04:22pm] <_zoggy_> so just noticed that if you add a show without it creating the folder... it doesnt marked the ep as skipped (my default)
0105-06-2012@[04:23pm] <_zoggy_> it did 'unknown'
0105-06-2012@[04:23pm] <_zoggy_> ok have create_missing_show_dirs & add_shows_wo_dir set to 1
0105-06-2012@[04:24pm] <_zoggy_> add new show. skipped w/season folder...
0105-06-2012@[04:25pm] <_zoggy_> ok that worked.. no folder created. all ep marked as skipped
0105-06-2012@[04:25pm] <_zoggy_> went and manually pp an ep..
0105-06-2012@[04:25pm] <_zoggy_> it created the show+season folder and pp without any problems
0105-06-2012@[04:28pm] <_zoggy_> ok deleted show. shutdown sb
0105-06-2012@[04:28pm] <_zoggy_> ok set: create_missing_show_dirs = 1 & add_shows_wo_dir = 0
0105-06-2012@[04:29pm] <_zoggy_> add new show. skipped w/season folder...
0105-06-2012@[04:31pm] <_zoggy_> ok, show folder created..  deleting show folder now
0105-06-2012@[04:31pm] <_zoggy_> went and manually pp an ep..
0105-06-2012@[04:32pm] <_zoggy_> it created the show+season folder and pp without any problems
0105-06-2012@[04:33pm] <_zoggy_> ok deleted show. left folders. went and added show back using 'add existing'
0105-06-2012@[04:33pm] <_zoggy_> that worked fine
0105-06-2012@[04:34pm] <_zoggy_> 16:34:30 INFO::SHOWQUEUE-FORCE-UPDATE :: Unable to load TVRage info: [Errno 10060] A connection attempt failed because the connected party did not properly respond after a period o
0105-06-2012@[04:34pm] <_zoggy_> f time, or established connection failed because connected host has failed to respond
0105-06-2012@[04:34pm] <_zoggy_> 16:34:30 WARNING::SHOWQUEUE-FORCE-UPDATE :: Unable to add TVRage info: urlopen call to http://services.tvrage.com/tools/quickinfo.php?show=Ugly+Americans failed
0105-06-2012@[04:35pm] <_zoggy_> ok deleted show. shutdown sb.
0105-06-2012@[04:35pm] <_zoggy_> ok set: create_missing_show_dirs = 0 & add_shows_wo_dir = 1
0105-06-2012@[04:36pm] <_zoggy_> add new show. skipped w/season folder...
0105-06-2012@[04:37pm] <_zoggy_> ok that worked.. no folder created. but all eps are marked as 'unkown' not skipped
0105-06-2012@[04:37pm] <_zoggy_> unknown
0105-06-2012@[04:37pm] <_zoggy_> i really cant see why there are two variables for this mod.. really they should just be one and the same
0105-06-2012@[04:37pm] <_zoggy_> but anyways.. so theres something to fix
0105-06-2012@[04:38pm] <_zoggy_> if a show is added and the show folder doesnt exist.. the initial status doesnt get set
0105-06-2012@[04:39pm] <_zoggy_> i can manually set the ep to skipped if i want though
```

thezoggy:
so as you can see, yes the default status does not get set when using the `add_shows_wo_dir` option. now why i never looked into it.. those options were supposed to be removed and never were documented... we were suppose to update our code to just handle it internally vs having special options..

CaptainRewind:
Yeah that makes sense. In my case, I don't use SB post-processing.  Instead, I use categories to defer the post-processing to sabnzbdplus.
Sorry to cross issues, then.  I didn't see it as a known issues because at the time I didn't know what I was looking for.
As for setting Archived for initial status, I've tested and I do see my issue go away when I have SB create the folder and set initial status to SKIPPED.  These two settings combined seem to be working in my case for now.

thezoggy:
no, i would say keep it open.. this deff needs to be addressed. just not sure how just yet.

CaptainRewind:
It would seem that the latest update e6dcfb307ba4ac3dc6388e9b0278497359109f40, or one recently prior, changed some db schema or dropped a table.  After updating, the band-aid I had originally proposed with my pull request is no longer working, and SB crashes.  I'll have to dig in a bit to figure out exactly why.  I didn't have much time when I last tried it, so it could be something as simple as the way Python handles whitespace (or lack thereof).  That's bitten me many times in the past, so maybe being in a hurry was a problem there.
The issue still exists where shows added while "add_shows_wo_dir = 1" is set all get set to "Unknown" status.  Even existing shows, whose folders have been removed, have the issue for new episodes showing as Unknown.
For now, I've reverted to 9a511dc9296dbc1292bd98c421cad9c85f416bdf until I have more time to look deeper into this.
Searching "Unknown" episodes whose TVDB date is current or in the past seems to be the only way to workaround the issue without a major change in the way new shows get added.  A long-term fix might be to fix the way initial status is set, and not have initial status set be dependent upon folder existence.  Perhaps a simple check for existence of folder, and then if folder doesn't exist, continue setting initial statuses as it would if no files exist in the folder?  That's certainly a more elegant way to handle it.  Of course, such a fix would also require a one-time db patch to change all previously-Unknown episodes to either skipped or Unaired (depending on the TVDB date)... at least for any existing DB's, any new installations wouldn't need it.
Let me dig in again and see what I can find.

dckone:
i have the same problem. my SB is running on linux, the TV folder is shared using samba, and i am watching shows from a DLNA device over the network.
since i delete shows after watching, i dont want subfolders for each show in the TV folder, only 1 folder for each episode. so i delete the show folder after adding a show.
now very often, especially with a new season, shows dont get downloaded, instead status of all episodes in that season is "unknown", even though the air date is just now or slightly in the future or the past.

please fix, because working around this behaviour requires constant attention.

dckone:
also, it would also be helpful to search for "unknown" status in the episode status management. this would make it at least easier to work around this problem

ghost:
I think the problem lies here: https://github.com/midgetspy/Sick-Beard/blob/development/sickbeard/tv.py#L1178

```
    #don't update show status if show dir is missing, unless missing show dirs are created during post-processing
    if not ek.ek(os.path.isdir, self.show._location) and not sickbeard.CREATE_MISSING_SHOW_DIRS:
        logger.log(u"The show dir is missing, not bothering to change the episode statuses since it'd probably be invalid")
        return
```

The UNKNOWN situation only happens here when 
CREATE_MISSING_SHOW_DIRS=0 and ADD_SHOWS_WO_DIR=1 right?

Suggested change #L1178 :

```
    if not ek.ek(os.path.isdir, self.show._location) and not sickbeard.CREATE_MISSING_SHOW_DIRS and not sickbeard.ADD_SHOWS_WO_DIR:
```

@CaptainRewind Can you check if that indeed solves your problem?

BTW Looks like if you set CREATE_MISSING_SHOW_DIRS=1 the show_dirs are only created when you postprocess. Guess, if you don't use postprocessing you can try setting CREATE_MISSING_SHOW_DIRS=1 too.

thezoggy:
@itofzo i discovered many issues with those options if you didnt use them together.. i can go dig up my irc logs about it...

conihorse:
What would it take to get this moved on? I've got this issue consistently, if it's any help I offer to test.

thezoggy:
@conihorse if you look at your config.ini, what is `CREATE_MISSING_SHOW_DIRS` and `ADD_SHOWS_WO_DIR` set to?

conihorse:
@thezoggy 
create_missing_show_dirs = 0
add_shows_wo_dir = 0
thanks

dgoodwin-:
I also have this issue on ALL my shows. 

create_missing_show_dirs = 0
add_shows_wo_dir = 0

I should add that I went and deleted all the show folders as I'm not using any post processing (it's a seedbox with limited storage space). This means that every show has this when you go to view it:

Location:   /pathToFolder/Family Guy (dir is missing)

Looks like this issue is caused if you go and manually delete the show directories. Another similar thread:

http://sickbeard.com/forums/viewtopic.php?t=5126&p=23647

My issue is that I do not do any post-processing so I cannot have it create the show directory with "create_missing_show_dirs = 1" . I think I will just manually create every folder, do a mass update, and see what the result is.

dgoodwin-:
Just an update:

Replacing those folders has solved everything. I was having some stability issues as well which I had not been able to pinpoint (thought the VPS might be flakey) and SickBeard was crashing a lot. It hasn't crashed once since replacing the folders. 

If you have deleted the show folders manually and are seeing stability issues and/or issues with "unknown" episodes - replace them! Clearly deleting these folders (while seemingly innocent) ends up breaking stuff.

PR #531, opened 2012-11-18
Add update once option to xbmc notifier

EchelonFour:
same as pull #492 but updated to work with the recent updates the notifier
Adds option to stop sending update commands to xbmc hosts once it has successfully updated once. If using a mysql database that has several clients, this is really useful as it stops duplicates.


thezoggy:
you would also need to modify code to make this work with testXBMC or updateXBMC functions (webserve.py).. and how you report back the results

thezoggy:
this can be closed, it was basically pulled just implemented differently.

PR #533, opened 2012-11-20
Renaming subtitles when renaming episodes. 

tomelfring:
This commit will rename all *_.srt files where *_ is a country code
specified in the ISO693-1 language codes.

Addresses:
http://code.google.com/p/sickbeard/issues/detail?id=1698
http://code.google.com/p/sickbeard/issues/detail?id=1805

This time for the development branche.

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -163,14 +163,22 @@ class PostProcessor(object):
         # don't confuse glob with chars we didn't mean to use
         base_name = re.sub(r'[\[\]\*\?]', r'[\g<0>]', base_name)
     
+        # list of ISO693-1 country codes
+        ISO693 = ["ab", "aa", "af", "ak", "sq", "am", "ar", "an", "hy", "as", "av", "ae", "ay", "az", "bm", "ba", "eu",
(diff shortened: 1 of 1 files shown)


jurrabi:
I see this is not merged in master. Any reason? Seems like a good idea for many people...

jurrabi:
I just tested tommy101 version and at least it renames my ep.es.srt files. It's doesn't keep the .es.srt part (renames to just .srt) and that's a shame since I usually have .es and .en files... I guess I'll take a look myself.

What I'm sure is that more people will benefit from this in master...

LordZepto:
:+1: I was trying to do the same. I usually have ep.es.srt and ep.en.srt

tomelfring:
So you say some-episode-name.es.srt is not renaming properly?

jurrabi:
I took a look 3 weeks ago and got partial success. Básically the issue (well, I wouldn't call it issue since it was designed that way) is split in several routines and I only caught one. But since I had already run my masive renaming I didn't kept looking.

I'm at work now and don't have access or remember what I found out, but I'l try to share later today.

thezoggy:
can you rebase your submission, we modified how sickbeard detects the extension (to support multi dot ones). I think it this may not even be needed now because of it..

PR #537, opened 2012-11-25
Newzbin code cleanup

thezoggy:
- Updated provider to set the `maxage` off the users `usenet retention` value rather than always be 0 (unlimited)
- Use **urllib2** and handle errors a little more gracefully, cleaned up logging messages and made them unicode safe
- Retry a max of **5** times when Newzbin is throttling API request (5 nzb in 60 secs max)
- Handle an odd bug where sometimes randomly we will try and fetch a http not an http**S** url which resulted in SB not being able to parse the reportid and failing (still can not figure out where this bug is coming from)
- Increased the Newzbin cache from `1` min to `15` min so it matches the rest of the providers (majority use 15min, while the actual min is 10mins)

### DIFF
--- a/sickbeard/providers/newzbin.py
+++ b/sickbeard/providers/newzbin.py
@@ -18,9 +18,10 @@
 
 import os
 import re
-import sys
 import time
 import urllib
+import urllib2
+import base64
 
 from xml.dom.minidom import parseString
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


thezoggy:
well go figure:
`It is with regret that we announce the closure of Newzbin2.`

thezoggy:
dunno if you wanna pull then we can do newzbin cleanup. that way this is not just lost (if for some reason newzbin is resurrected)

midgetspy:
Naw just leave it in a branch, same thing.

PR #539, opened 2012-11-26
Added the ability to override resizing the images served up from the cache if PIL is installed.

thezoggy:
- Add `--noresize` to command line argument so that the user could disable the use of PIL on resizing the poster/banner cached images. (ex: they need PIL for another python app but don't want PIL to be used for sb..)
- Corrected the etree import so it mimics how tvdb_api does (will try cElementTree first then fall back to the slower non-C ElementTree)

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -162,9 +162,9 @@ def main():
     threading.currentThread().name = "MAIN"
 
     try:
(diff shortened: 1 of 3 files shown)


danfozzy:
please pull this! :)

PR #542, opened 2012-11-30
Remove Newzbin as a provider

TvdW:
This is a pretty sad pull request, but someone has to do it. These commits remove Newzbin from Sick-Beard, while keeping the 'History' page intact.


thezoggy:
this is already done in another branch, there are certain things you cant do and requires config upgrade to handle legacy references. also note that all code submissions need to happen against the development branch, not master.

PR #543, opened 2012-12-02
AddNewShow tvdb_api languages select fixed (would only contain the fallback 'en')

SerhatG:
A previous change caused the tvdb languages select to not be populated
anymore. This happens when adding a new show from scratch. Reverting
back the culprit line.


thezoggy:
Thanks for catching this, working on too many branches :(

SerhatG:
No problem. Glad I could contribute one way or another.

PR #547, opened 2012-12-06
Plugin architecture

bcyrill:
This pull request proposes a plugin architecture using [Yapsy](http://yapsy.sourceforge.net/).

So far I have only implemented the notifiers but in the future the metadata and providers can also be converted. The 'Update Plex' and 'Update XBMC' buttons are still missing. Everything else should be working.

A plugin system allows easy extension of Sick-Beard without having to touch the core system which eases the introduction of custom providers that are only used by a few users.

For now I image a plugin configuration page where a user can add custom github repositories and then install, uninstall or update the additional plugins. Any further ideas?

And please, feel free to comment on the current implementation there's still room for improvements.


RuudBurger:
Just wanted to say something about this, as @aanderson posted it over to CouchPotato issues.
If you want to make this work, make sure you don't make so much duplicate code.. All the HTML files make my eyes hurt. And while you start over with some stuff, maybe don't keep the crappy (and ugly code) for accessing config parameters.
Screen Shot 2012-12-22 at 22 41 43
Really?

Maybe have a look at how providers are implemented in CP (which is kind of the same). And yes I'm saying this because I like my way better. But otherwise I wouldn't have done it this way ;)
Example: https://github.com/RuudBurger/CouchPotatoServer/tree/master/couchpotato/core/providers/nzb/nzbx

bcyrill:
Actually it doesn't make additional duplicate code, but also it doesn't reduce it. ;-)
This is just a restructuring of the existing code with a very lightweight plugin system that was done in a few hours and there is still a lot of room for improvements.

RuudBurger:
I understand, but maybe improve it. Instead of just restructuring it.

PR #551, opened 2012-12-16
Adds Bonjour Announce

jphastings:
I thought it'd be handy for Sick Beard to announce itself via DNSSD (bonjour) - particularly because [this guy](http://dnssd.me/) made an extension for Chrome & Firefox so you can see all the locally announced sites in your browser and I like that I don't have to remember my server's IP.

So, I added some code to do just that, it's not particularly pretty but it works and fails gracefully!

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -116,6 +116,12 @@ def daemonize():
         logger.log(u"Writing PID " + pid + " to " + str(sickbeard.PIDFILE))
         file(sickbeard.PIDFILE, 'w').write("%s\n" % pid)
 
+def bonjour_register_callback(sdRef, flags, errorCode, name, regtype, domain):
+    if errorCode == pybonjour.kDNSServiceErr_NoError:
+        logger.log(u"Announced Sick Beard via bonjour.")
+    else:
+        raise Exception("Bonjour announce failed. Code: %d" % errorCode)
+
 
 def main():
     """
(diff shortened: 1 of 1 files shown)


SickBeard.py
```diff
@@ -116,6 +116,12 @@ def daemonize():
         logger.log(u"Writing PID " + pid + " to " + str(sickbeard.PIDFILE))
         file(sickbeard.PIDFILE, 'w').write("%s\n" % pid)
 
+def bonjour_register_callback(sdRef, flags, errorCode, name, regtype, domain):
+    if errorCode == pybonjour.kDNSServiceErr_NoError:
+        logger.log(u"Announced Sick Beard via bonjour.")
+    else:
+        raise Exception("Bonjour announce failed. Code: %d" % errorCode)
```
midgetspy, line 123: This should be logger.log(...., logger.ERROR). Throwing an exception there will bring the entire process down which is definitely not warranted if we just failed to announce.

SickBeard.py
```diff
@@ -309,6 +315,12 @@ def main():
             sickbeard.launchBrowser(startPort)
         sys.exit()
 
+    try:
+        import pybonjour
+        sdRef = pybonjour.DNSServiceRegister(name = "Sick Beard", regtype = '_http._tcp.', port = startPort, callBack = bonjour_register_callback)
+    except Exception,e:
```
midgetspy, line 321: Never catch Exception - this should be ImportException :0)

jphastings, line 321: Does my almost total lack of Python experience show? :P Glad you've had the time to review it :)

PR #552, opened 2012-12-17
Upgraded jquery 1.7.2 -> 1.8.3.

thezoggy:
Upgraded jquery 1.7.2 -> 1.8.3.
http://blog.jquery.com/2012/11/13/jquery-1-8-3-released/

Just minor fixes, figured might as well get this updated (just in case it had any effect on the upcoming JS code review).


mtakacs:
1.9.0 is out now..

thezoggy:
Yes but it breaks code as it deprefs some functions, and some plugin refactors. Why there is jQuery migrate. Also there is 1.10.1 too btw (it drops legacy is). I'm on the jquery-UI dev group.

-----Original Message-----

mr-orange:
Hy zoggy I've upgraded my branch a while ago with jquery 1.8.3 and this broke something o gui so I have had to fix.
An example is test button for sabnzb and torrent clients (but this is only on my branch).
It seem that this type of jquery selector is not working anymore $("input=[name='sab_host']").val().
this is the fix i made 
https://github.com/mr-orange/Sick-Beard/commit/9b51fe3ccabbcc527a235f5e0cd58f48b32b13dd

or you can use this name selector $("input[name='sab_host']")  as reported http://api.jquery.com/attribute-equals-selector/

PR #555, opened 2012-12-26
XBMC 12 (Frodo) brings a new file naming convention

MoserB:
XBMC 12 (Frodo) brings a new file naming convention (see http://wiki.xbmc.org/index.php?title=XBMC_v12_(Frodo)_FAQ#Local_images).

This adds basic support.

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -117,6 +117,7 @@ ROOT_DIRS = None
 USE_BANNER = None
 USE_LISTVIEW = None
 METADATA_XBMC = None
+METADATA_XBMCFRODO = None
 METADATA_MEDIABROWSER = None
 METADATA_PS3 = None
 METADATA_WDTV = None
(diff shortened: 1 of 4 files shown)


MoserB:
Yes Frodo is backwards compatible and will still read those files, but the problems start when you export the database from XBMC; there only the new filenames are supported and this will create a mess.

Personally I think that the new filenames are superior because I never liked the .tbn extension and now you can have both posters and banners on the season level.
Also, whats wrong with giving users a choice?

palli81:
Is it possible to extend this to download the poster _and_ the banner and then save them as poster.jpg and banner.jpg?

Clearmist:
The new XBMC naming conventions are still not fully implemented in SickBeard. Episode thumbnails should be "[episodename]-thumb.(png/jpg)" and not "[episodename].tbn".

Cranial:
I believe it's creating both actually. Well, it is for me.

I've been running with this commit for quite some time now, it's been stable. Any reason why it's not being pulled to mainline?

thezoggy:
just havent had a chance to review, kinda waiting on 12.3 to drop (should be any day now)

Cranial:
No worries. Before merging, as brought up by Clearmist, I think Frodo option should have all ".tbn" creation removed, to keep in line with current naming convention.

Clearmist:
The new naming conventions were committed in March 2013, and I am running a July release ( fb37d332b333b43e7bcae46b619ae6b7ef94e266). Why is SickBeard still creating .tbn files and not allowing me to select the Frodo naming scheme?

sick beard - alpha master - config - post processing_2013-07-21_13-56-26

vasc0x:
I've been dealing with this issue for some time also, so I decide to manually edit my SB source files to get the results I want. After the changes I'm actually getting filename-thumb.jpg instead of filename.tbn. If anybody is interested, let me know and I'll upload the files right here. I'm too busy now to fork anything.

mr-orange:
I've imported the work in my dev branch and will test it for few days... It should work but it's really missing proper check on saving (You can't enable at the same time xbmc and xbmc frodo)

Cranial:
vasc0x , I'd like to grab those files if possible? Sorry for the delayed response, been a bit busy.

MilhouseVH:
Won't calling this new metadata selection "XBMC (Frodo)" be confusing when users are running Gotham, or (God forbid) Eden? I think we should call this new selection "XBMC v12+" to avoid confusion.

If it must be called "XBMC (Frodo)", at least include "+" on the end to give some indication that the selection doesn't only apply to Frodo but also versions released after Frodo (some users _may_ realise Gotham > Frodo).

Ideally the old selection should be renamed from "XBMC" to "XBMC Pre-v12" but that might break existing user settings, I guess.

thezoggy:
using internal code names is not advisable. using the version is the route to go.. XBMC v12+ would be fine as most people would understand it.

cdloh:
Anything else happening on this front? It's pretty handy as the new naming convention is nice for the webui and the remote apps.

MilhouseVH:
XBMC 12+ support was added to mr_orange's The Pirate Bay fork of SickBeard back in August ([here](https://github.com/mr-orange/Sick-Beard/commit/1a72134f22e0ab9bd1d46c58a7a27c1fc367d428) with subsequent update changing [Frodo -> v12](https://github.com/mr-orange/Sick-Beard/commit/ffbf0b407da5a70283805657406a3402f667ee98) per discussion above), and it works well, without any problems for pre-12 users.

I'm really surprised to see this still hasn't been integrated into the midgetspy repo, is there any specific reason for not doing so?

thezoggy:
Note that there is still some more work to be done on the XBMC12+ metadata... 
https://github.com/midgetspy/Sick-Beard/wiki/ChangeLog

PR #557, opened 2012-12-30
add omgwtfnzbs support...

imajes:
Re-created commit to add omgwtfnzbs support. This is an early patch, and i'm not awesome with python -- so please feel free to refactor.


imajes:
Oh, i fixed the line endings for generic.py. it was in dos format, and in the process of trying to use it, i had to convert it :)

thezoggy:
dunno why line endings needed to be changed :x but anyways, your image you included for the provider is .gif not .png like we use. also make some other minor comments

thezoggy:
This should be closed as jrdnull has an updated one hes submitting tomorrow to replace this... will update once its submitted

PR #562, opened 2013-01-02
Support defining TVDB API Base URL in config.ini

Thermionix:
Allow's setting the TVDB API Base URL in the config.ini - It isn't exposed in gui

This won't change the workflow for anyone who won't use the option, and abstracts the setting outside of hardcoded-inside-source.

Many user's are using a proxy in front of the TVDB API to improve matching data

PR #563, opened 2013-01-02
Support defining TVDB API Base URL in config.ini

Thermionix:
Allow's setting the TVDB API Base URL in the config.ini - It isn't exposed in gui

This won't change the workflow for anyone who won't use the option, and abstracts the setting outside of hardcoded-inside-source.

Many user's are using a proxy in front of the TVDB API to improve matching data

### DIFF
--- a/lib/tvdb_api/tvdb_api.py
+++ b/lib/tvdb_api/tvdb_api.py
@@ -308,7 +308,8 @@ class Tvdb:
                 search_all_languages = False,
                 apikey = None,
                 forceConnect=False,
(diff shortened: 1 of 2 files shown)


Thermionix:
Changed branch from https://github.com/midgetspy/Sick-Beard/pull/562

joeserg:
This would be insanely useful, especially as TVDB URL returns an intermittent 501 error at the moment for some users.

thezoggy:
its 503. and its an issue with their cdn, cloudflare.

PR #566, opened 2013-01-19
Updated NZBs'R'US provider and added backlog searching.

jrdnull:
I used the other providers for an example and have overridden __get_title_and_url__ due to differences in the XML returned between the RSS feed and the API, if anything needs changing let me know as I am not that familiar with Python.

Changes:
- Updated RSS feed to include the SD-x264 and misc groups.
- Updated labels for credentials to reflect naming on NZBs'R'US website.
- Added backlog searching through the API documented [here.](http://www.nzbsrus.com/NzbsrusAPI.pdf)

### DIFF
--- a/data/interfaces/default/config_providers.tmpl
+++ b/data/interfaces/default/config_providers.tmpl
@@ -130,13 +130,13 @@
                     <div class="providerDiv" id="nzbs_r_usDiv">
                         <div class="field-pair">
                             <label class="clearfix">
-                                <span class="component-title">NZBs'R'US UID</span>
+                                <span class="component-title">NZBs'R'US User ID</span>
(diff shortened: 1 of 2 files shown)


jrdnull:
I've implemented those changes zoggy, thanks for the pointers. If everyone is ok with the usage of ElementTree I can clear up some of the providers when I get some free time.

Let me know if anything else requires changing.

thezoggy:
should rebase/squash commits to cleanup the history. force push to update this pull request.

jrdnull:
Zoggy, I've changed the formatting as you suggested, checked the retention and given it a default value.

Regarding the credentials check, I think it will be enough that the user has entered them, if they are incorrect they will get an error in the log. I'm unsure if all keys are 6 digits or if that may change, its not documented so I don't think we should check.

Also I'll see if I can find the language id's on the website.

thezoggy:
This is good to go.

PR #568, opened 2013-01-25
Use of website dereferer.org to hide source address

Kwebbel:
Prevent the destination url to trace back location of sickbeard website with HTTP_REFERER headers. 
With this change the website thetvdb will not be able to trace back the sickbeard location because HTTP_REFERER will be set to dereferer.org.

### DIFF
--- a/data/interfaces/default/comingEpisodes.tmpl
+++ b/data/interfaces/default/comingEpisodes.tmpl
@@ -132,7 +132,7 @@
     <span class="quality Custom">Custom</span>
 #end if
         </td>
-        <td align="center"><a href="http://www.thetvdb.com/?tab=series&amp;id=${cur_result["showid"]}" onclick="window.open(this.href, '_blank'); return false;" title="http://www.thetvdb.com/?tab=series&amp;id=${cur_result["showid"]}"><img alt="[info]" height="16" width="16" src="$sbRoot/images/thetvdb16.png" /></a></td>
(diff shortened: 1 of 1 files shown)


thezoggy:
dunno why you think this is needed though? run your sb through a reverse proxy if your paranoid...?

Kwebbel:
I personally find it a clean way to "protect" the users that have a closed sickbeard installation and dont want their private domains be looked up by some US based company/organization.

ppslim:
This is a great idea.

@thezoggy I think you have the wrong idea. A reverse proxy offers the same protection as running it direct. Yes, the proxy can offer additional authentication mechanisms and protection depending on how you configure it.

Accessing the link still has a referrer header though. Be it a public IP or internal IP, regardless of weather you hold it in public/private such as on a DSL static IP, hosted VM, internal to your network or access via VPN/SSH tunnel.

This feature is not going to offer one single bit of security to your setup and if people think it should reduce their paranoia, I want to be there to film the reaction when they realise the horse bolted a long time ago.

Kwebbel:
Additional example: if you click on a search result on google.com, which redirects you to www.imdb.com, the HTTP_REFERER parameter, within your request to imdb.com, is set to www.google.com. This way imdb.com can see that you came from google.com. 

Now; lets say someone has a SB installation on his private domain named sb.example.com. If he clicks on any link within SB that goes to an external website, that specific external website will see sb.example.com as HTTP_REFERER. By using the dereferer.org services the HTTP_REFERER header will be set to dereferer.org. This way the destination url will never see the origin (sb.example.com).

Also be aware that some domain providers show the address of their customers in the whois database which is accessable for anyone.

midgetspy:
Thanks. Might as well add it to the paypal link and the providers icon links too :0)

thezoggy:
and also the tvdb links on the coming ep page

thezoggy:
went ahead and just did this myself (excluding the provider links) - note that i used the base64 method and went with another site..  https://github.com/midgetspy/Sick-Beard/pull/698

PR #570, opened 2013-01-27
Search episodes between two dates.

wimhaanstra:
Added a method to search for planned or already broadcast episodes between two dates.

Not sure the code is up to standards for the Sick-Beard repo, but I tried to maintain the same code style.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -681,6 +681,68 @@ class CMD_Help(ApiCall):
         else:
             out = _responds(RESULT_FAILURE, msg="No such cmd")
         return out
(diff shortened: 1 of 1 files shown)


thezoggy:
all new code should go in development not master. the name of the function is misleading.. as episode.search is the same thing as clicking the manual search. this code looks just to return some sql data.. i would suggest a different name...

wimhaanstra:
I agree. Also, my python coding is not up to par. The pull-request was actually done to show that I would like to add this feature added to the API, but I am not sure where/how, but adding the code might speed things up a bit.

PR #575, opened 2013-02-03
First step of converting db use to go through the PeeWee ORM library.

hobeone:
This is the first incremental step of converting SickBeard to use an ORM for it's database.  When finished it would allow mysql or postgres (in addition to sqlite) to be used as the DB.

This pull request adds the base models, tests and converts tv.py to use the new models.

This is my first time using github so let me know if I'm doing anything wrong.

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -49,6 +49,7 @@ import getopt
 import sickbeard
 
 from sickbeard import db
+from sickbeard import db_peewee
 from sickbeard.tv import TVShow
 from sickbeard import logger
 from sickbeard.version import SICKBEARD_VERSION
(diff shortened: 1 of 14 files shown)


thezoggy:
new code would go in development not master. why peewee vs sqlalchemy or something else?

thezoggy:
http://charlesleifer.com/blog/peewee-a-lightweight-python-orm/

hobeone:
Can I change the target branch in this pull request or should I create a new one?

Looking over all of the Python ORMs I actually liked the Django one the most but I think you can't use that without using the whole framework.  SQLAlchemy seemed really heavyweight and required a bunch of setup where peewee took about 10 minutes to get going.

The Sickbeard use case is super simple so I don't think we'd need to use any of the advanced features of SQLAlchemy.

Also Peewee had this section on multi threaded apps:
https://peewee.readthedocs.org/en/latest/peewee/cookbook.html#multi-threaded-applications

I found a bunch of warnings about using SQLAlchemy with multithreaded cherrypy apps but I didn't dig into that too much.

thezoggy:
you would have to submit a new branch.. as theres different code/changes in development not yet in master

PR #582, opened 2013-02-11
Adds next episode season and number to home page

keltroth:
Almost everything is in the title.

I was oftenly looking for the last episode number for some reasons so I add it to the home page. What to you guys think ?

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ ehthumbs.db
 Thumbs.db
 .directory
 *~
+
+nohup.out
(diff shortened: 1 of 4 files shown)


thezoggy:
what about air-by-date shows?

thezoggy:
also, all new code goes into development not master

keltroth:
Hi,

First, sorry for pushing on master...
air-by-date is great too. But most of time, I'm looking for new episode manually and having the episode number here is very usefull.

PR #585, opened 2013-02-13
Image optimization

douglasdollars:
I noticed there were some pretty large assets in the repo. I ran them through [ImageOptim](http://imageoptim.com/), which has knocked out a lot of the metadata so they can load up a little speedier, but not changed them visually in any way.


war59312:
Convert to WebP for even faster! :dancers: 

For example poster.png.

Original Size: 696 KB (713,511 bytes)

WebP Size: 105 KB (107,900 bytes)

BTW poster.png is your optimized one from https://github.com/theDoug/Sick-Beard/blob/06ec33bdb262df1a9702286dbd854f606fbcf1aa/data/images/poster.png

That's a reduction of 6.6x. WebP is awesome!

douglasdollars:
I agree that it tends to convert to a smaller file size (and quite impressively!), but it's not yet natively supported in Internet Explorer or Safari and [requires a plugin](http://en.wikipedia.org/wiki/WebP#Support) to be visible. This seems like a bad burden to offset to the user currently.

war59312:
Should not be too difficult to take advantage of WebPJS @ http://webpjs.appspot.com/ ;)

thezoggy:
extra work for little gains and adding more complexity to update/maintain is not ideal. if anything we should switch to a sprite for all our images instead of them being separate.. which is one of the many things i have planned but just never enough time :( one of the big ones was getting over to using bootstrap for the menu.. i left all the old images still in the repo in case we need them.. its prob safe to delete them and clean up some of the other crud (lots of crud in there)... but anyways the banner/poster is prob the only thing thats worth optimizing as when i made those i was in a hurry and really didnt try to optimize them for mobile users as most people browse locally on a network.. and a few extra kb goes un-noticed to most

war59312:
Indeed. Love sprites.

simonk83:
I prefer 7-Ups.

(sorry)

drewfreyling:
So does that mean @thezoggy this pull request needs to be updated to just optimize the banner/poster?

thezoggy:
for starters this pull request needs to be based off dev not master, since all new code goes to dev first. and yes honstly the banner/poster is prob the only thing really worth optimizing at this point. all the other stuff will prob be deleted/redone.

thezoggy:
and no dont use some proprietary format to try and save some kb. personal stance is that we use png first, and jpg/gif for last resort.

douglasdollars:
Corrected/moved to midgetspy/Sick-Beard#640

PR #586, opened 2013-02-13
1080p updates

thezoggy:
- Added `webrip` string to the applicable categories
- Added 1080p related qualities to the **common_tests** routine, noticed that we forgot to update our tests for custom_naming changes. Fixed some of that.


mr-orange:
Good to know. I can filter out easily this group. However you are telling me that every web.dl is HD? right?

PR #587, opened 2013-02-16
Season and episode offsets

pieterwasalreadytaken:
Implemented per-show season offset and episode offset for current search queries for differences between the 'scene' and thetvdb.org. Must be manually and personally maintained per show.

I tested with the SickBeard index and the post processor. When one does use sabnzb+ to postprocess, things could get ugly.

Furthermore, any future providers should be using the sceneseason and sceneepisode fields in the TVEpisode object in stead of the regular season and episode fields.

More information on google code:
http://code.google.com/p/sickbeard/issues/detail?id=1352

and the forum:
http://sickbeard.com/forums/viewtopic.php?f=8&t=6600


pieterwasalreadytaken:
Discussed with thezoggy on IRC.. Too much of a delicate solution which could easily break things.

PR #588, opened 2013-02-17
Bugfix :: Postprocessing: nzbName and multiple videos in process folder caused files to be lost

Conjuro:
Multiple videofiles in same folder and nzbName variable set, would lead to loss of files. All videofiles would be processed using the same parsing from the nzbName, leading to each videofile overwriting the previous one.

### DIFF
--- a/sickbeard/processTV.py
+++ b/sickbeard/processTV.py
@@ -94,6 +94,10 @@ def processDir (dirName, nzbName=None, recurse=False):
 
     remainingFolders = filter(lambda x: ek.ek(os.path.isdir, ek.ek(os.path.join, dirName, x)), fileList)
 
+    # If nzbName is set and there's more than one videofile in the folder, files will be lost (overwritten).
+    if nzbName != None and len(videoFiles) >= 2:
+        nzbName = None
+
     # process any files in the dir
     for cur_video_file_path in videoFiles:
 


thezoggy:
what if one is a sample?

Conjuro:
Processing sample files remains unchanged.

Line 88: `videoFiles = filter(helpers.isMediaFile, fileList)` filters out files with the word sample `('(^|[\W_])sample\d*[\W_]')`

And process() in postProcessor.py still uses _checkForExistingFile() to check for larger/smaller file, then using priority_download decides on replacing or not.

The only thing that changes with this patch is that _find_info in postProcess.py looses its two first methods (self._history_lookup and self._analyze_name) to determine showid, season and episode. Since these two methods were the cause of files getting wrongly overwritten, this is a good thing.

Conjuro:
I realized I haven't given a good explanation of what goes wrong and why the patch is needed. Here's an example:

You have a download folder containing several episodes that have for whatever reason not yet been processed. At some point Sick Beard is called to process a download and `nzbName` is set (e.g. `nzbName='AwesomeShow.S01E16.RLG.nzb'`).

`processDir()` lists all video files in the folder and calls `postProcessor.process()` on each one. First one in this example is 'AwesomeShow.S01E01.RLG.avi'.

`process()` calls `_find_info()` to determine showid, season and episode. It has several methods to accomplish this, It checks snatch history, it parses nzbName, file name, folder name and so forth. The two first methods are snatch history and nzbName parsing, both of which rely on nzbName and either will return showid for 'AwesomeShow', season 1 and episode 16.

First file finishes process/move and the database is updated to point episode 16 at file S01E01.

Next file in folder (AwesomeShow.S01E04.RLG.avi) begins process, same procedure as above and nzbName stil parses to season 1 and episode 16. If 'AwesomeShow.S01E16.RLG.nzb' download is set as a priority or 'AwesomeShow.S01E04.RLG.avi' is larger than 'AwesomeShow.S01E01.RLG.avi' then E01 is getting overwritten by E04 and db is updated to point to E04.

This repeats for every video file in the folder until all except the last file is wiped out.

My patch nulls out nzbName if there are more than one video file in the folder, which basically means `_get_info()` can't use snatch history or nzbName parsing to determine showid, season and episode, and will move on to the next methods.

PR #590, opened 2013-02-17
Also index the folder when using the synology notifier

hrmohr:
When a new season begins the new folder is not indexed using the current method of just indexing the file. This hides the new episodes in this season when accessing them using Samsung AllShare. I've just added an extra call to the notifier that also indexes the folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ tests/cache.db
 *.tmproj
 *.tmproject
 *.sw?
+.idea
 
 # OS generated files #
 ######################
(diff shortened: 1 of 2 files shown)


thezoggy:
please dont 'fix' line-endings. it rewrites the whole file and makes it hard to know what was changed.. also it rewrites git history. and finally. new code goes into development not master. also dont submit pulls until they are ready/done/tested.

mreid-tt:
hey, with all the lines you changed it's difficult to see what code you added... perhaps you can take a look at a fix I just proposed under pull request #591 and if they achieve the same thing then maybe one of these can be withdrawn...

hrmohr:
Thank you for your replies. I will be more careful the next time.

It's hard to know what was changed if any code was changed when 'fixing' line endings/tabs. But that is why I committed the 'fix' and my addition in separat commits.

#591 seems to be a more general fix and not just an add folder.

PR #593, opened 2013-02-19
NMJv2 Notifier - Improved

racquemis:
Notifier for the Networked Media Jukebox for the 300 and 400 series.
Now uses the NMJ API on the PCH to locate the database. No need to search the contents of the shares anymore.

Users need to select the DB location (local harddrive or network share), the Find Database button does the rest. For users with more then one Database: If the wrong database is selected the 'Database Instance' dropdown menu can be used to skip the first Nth DB occurences.

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -262,6 +262,92 @@
 
                     </fieldset>
                 </div><!-- /nmj component-group //-->
+				
+				<div class="component-group clearfix">
+                    <div class="component-group-desc">
+						<img class="notifier-icon" src="$sbRoot/images/notifiers/nmj.png" alt="" title="Networked Media Jukebox V2" />
+                        <h3><a href="http://www.popcornhour.com/" onclick="window.open(this.href, '_blank'); return false;">NMJv2</a></h3>
+                        <p>The Networked Media Jukebox, or NMJv2, is the official media jukebox interface made available for the Popcorn Hour 300 & 400-series.</p>
+                    </div>
+                    <fieldset class="component-group-list">
+                        <div class="field-pair">
+                            <input type="checkbox" class="enabler" name="use_nmjv2" id="use_nmjv2" #if $sickbeard.USE_NMJv2 then "checked=\"checked\"" else ""# />
+                            <label class="clearfix" for="use_nmjv2">
+                                <span class="component-title">Enable</span>
+                                <span class="component-desc">Should Sick Beard send update commands to NMJv2?</span>
+                            </label>
+                        </div>
+
+                        <div id="content_use_nmjv2">
+                            <div class="field-pair">
+                                <label class="nocheck clearfix">
+                                    <span class="component-title">Popcorn IP address</span>
+                                    <input type="text" name="nmjv2_host" id="nmjv2_host" value="$sickbeard.NMJv2_HOST" size="35" />
+                                </label>
+                                <label class="nocheck clearfix">
+                                    <span class="component-title">&nbsp;</span>
+                                    <span class="component-desc">IP of Popcorn 300/400-series (eg. 192.168.1.100)</span>
+                                </label>
+                            </div>
+							<div class="field-pair">
+                                <label class="nocheck clearfix">
+                                    <span class="component-title">Database Location</span>
+									<span class="component-desc">
+										<INPUT TYPE="radio" NAME="nmjv2_dbloc" VALUE="local" class="radio" id="NMJV2_DBLOC_A" #if $sickbeard.NMJv2_DBLOC=="local" then "checked=\"checked\"" else ""# />PCH Local Media<br />
+									</span>
+                                </label>
+								    <label class="nocheck clearfix">
+                                    <span class="component-title">&nbsp;</span>
+                                    <span class="component-desc">
+										<INPUT TYPE="radio" NAME="nmjv2_dbloc" VALUE="network" class="radio" id="NMJV2_DBLOC_B" #if $sickbeard.NMJv2_DBLOC=="network" then "checked=\"checked\"" else ""#/>PCH Network Media
+									</span>
+                                </label>
+								</label>
+								<label class="nocheck clearfix">
+                                    <span class="component-title">Database Instance</span>
+                                    <span class="component-desc">
+									<select id="NMJv2db_instance">
+										<option value="0">#1 </option>
+										<option value="1">#2 </option>
+										<option value="2">#3 </option>
+										<option value="3">#4 </option>
+										<option value="4">#5 </option>
+										<option value="5">#6 </option>
+										<option value="6">#7 </option>
+									</select>
+									</span>
+                                </label>
+								<label class="nocheck clearfix">
+                                    <span class="component-title">&nbsp;</span>
+                                    <span class="component-desc">Adjust this value if the wrong database is selected.</span>
+                                </label>
+                            </div>
+                            <div class="field-pair">
+                                <label class="nocheck clearfix">
+                                    <span class="component-title">Find Database</span>
+									<input type="button" value="Find Database" id="settingsNMJv2" />
+                                </label>
+                                <label class="nocheck clearfix">
+                                    <span class="component-title">&nbsp;</span>
+                                    <span class="component-desc">The Popcorn Hour device must be powered on.</span>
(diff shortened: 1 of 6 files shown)


thezoggy:
can you squash and rebase your commits. also could you not change the line endings so we can see what you actually modified in the file? so does this extend the existing nmj for more devices while still working with older versions? does this require users to upgrade to a certain version to have the nmj api support the pch location?

racquemis:
For some reason my github client was changing CRLF to LF. Fixed.

This notifier isn't backwards compatible with the 100 & 200 series. Should be easy enough for someone to merge the NMJv1 and this notifier. However i'm not sure if that is wise since the 200 series will in time be updated to NMJv2 anyway. 
I came across the API calls while analyzing the NMJ Navigator App hat was released by Syabas quite recently. while there has been firmware updates to enable support for that application the API calls have been present alot longer then that as the API documentation shows.I can't say for certain that it will work, it's hard to check when nearly everybody in the community updates their firmware as soon as a new one becomes available. I'm also on the latest firmware.

ppslim:
@racquemis take a look at
http://git-scm.com/book/ch7-1.html#Formatting-and-Whitespace

racquemis:
I tried the core.autocrlf approach but didn't seem to work. Solved it another way so all is good :)

Conjuro:
Port forwarding through a firewall would allow changing the port.

racquemis:
possible but improbable since NMJ and the media collection (and therefore SickBeard) will generally be located in the same network.

midgetspy:
I can't test it but it at least handles error gracefully and the config appears to work so I'll pull it, thanks!

kernow:
I've found bug in this where the database path is not stored when the returned samba network path does not contain the string "://". Please see the following:

Output form the first api call:
https://gist.github.com/kernow/5403414
path is: smb://[foo=bar]GIGANTO/Video

Output from the second api call:
https://gist.github.com/kernow/5403426
database_path is: /share/nmj_database_1/media.db

This line is the issue https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L79 as the second condition is never met. I have removed the second condition and tested using my own setup and it works. I'm not sure the reason behind checking for "://" shouldn't it be assumed the popcorn returns the correct network path rather than mistrusting the information returned?

In my copy I have line 79 reading as `if dbloc=="network"`

racquemis:
I don't fully understand the problem. Can there be samba shares without the "smb://" prefix? i was under the assumption that popcorn hour needed that prefix to know its a samba share.

the condition on "://" is met.
the api call returns both network and local shares (it's possible to list only network or local shares, but there a bug which can crash the entire API and GUI when a specific sequence is called). because it lists both shares i need to distingish between the two. i do that with the strings 'localhost' and '://' to select the correct one.

I could have missed something. i appreciate any extra information you can give to clarify the possible issue you noticed.

kernow:
@racquemis I believe (I could be wrong) you are being too defensive with your code. There is no need for the additional checks `DB_path.find("localhost")>-1` and `DB_path.find("://")>-1`. The result of a metadata_database API calls  `returnValue` is only 0 when a database exists in that location. For instance, my database is in a network location, when the metadata_database API call is made for the localhost database `returnValue` is not 0. Then when the API call for the network database is called the `returnValue` is 0. Therefor you only need to check the value of `returnValue` and use a case/switch statement with the value of `dbloc` for network or local database locations.

As you can see from my example API call responses the database_path is `/share/nmj_database_1/media.db` and uses samba, there is no smb:// prefix. I've tested this and it works with my popcorn A-300 connected to a synology NAS.

racquemis:
I think you misunderstand the code.
DB_path.find("localhost")>-1 and DB_path.find("://")>-1 are not used to check if the database is valid but to select the db location from a list of shares. only when the share the selected a api call is made to see if a db exists in that location.

kernow:
As far as my understanding of the code goes, I'm a ruby programmer not a python programmer so could be missing something.

https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L60
Calls the file_operation api method which returns a list of paths, these are both network and local paths.

https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L64
the for loop iterates over each path

https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L69
Calls the meta_database api method for a path within the for loop

https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L73
`returnValue` is checked to see if it's 0

https://github.com/midgetspy/Sick-Beard/blob/5dd33643a22d3e9437ba88625ec49f9135cdbd9d/sickbeard/notifiers/nmjv2.py#L74
The database_path value is parsed from the xml and then used in the two conditionals in question

So in operation using my example XML responses whats happening is this.
1. The file_operation api call is made and two paths are returned, local path /opt/sybhttpd/localhost.drives/USB_DRIVE and network path smb://[foo=bar]GIGANTO/Video
2. The two paths are looped through, the local path is first in the list
3. A metadata_database api call is made for the local path, which responds with a return value of 57, nothing else happens in this iteration of the loop as the conditional evaluates to false due to returnvalue not being equal to 0
4. A metadata_database api call is made for the network path which responds with a return value of 0
5. As the conditional evaluates to true the database_path is parsed from the XML and the two conditionals in question are evaluated, both of which evaluate to false as it's a network path with no "smb://" prefix.

I can't see what use the additional check for the existence of "://" (and "localhost") plays here, we have already checked the returnValue provided by the api call is "0" so we know it's a valid path, we also know if it's a network or a local path as this is passed into the notify_settings method as a parameter.

racquemis:
Your wrong on your last sentence. you don't know if the share is local or path. dbloc is a parameter passed from the webpage and it only lets the function know which kind of share it needs to select.
in case dblloc="local" > check if share is local >> search for "localhost" in path
in case dbloc="network"> check if share is on network >> search for "://" prefix in path

but you may be right about one thing, if network paths can indeed lack a prefix ";//" the script needs to be updated.

However i propose to change:
if dbloc=="network" and DB_path.find("://")>-1:
to
if dbloc=="network" and DB_path.find("localhost")==-1:

kernow:
I've been thinking about this some more, can't we just get rid of dbloc entirely? I can only go by what my popcorn box has been returning. In my situation (and I suspect most other situations) there is no need have a local/network field in the interface.

TheDavidBox api returns a returnValue of 57 "No database is found" for each api call to metadata_database where no database was found at that path. The dbloc variable could be removed and all that need be checked is the returnValue. e.g.

```
if returnvalue=="0":
  sickbeard.NMJv2_HOST=host
  sickbeard.NMJv2_DATABASE=DB_path
  return True
```

It's actually irrelevant to sickbeard if the database is stored locally or in a network location, the only thing that is relevant is the path.

The only situation I can see this not working in is if a database exists in multiple locations, however the existing code doesn't fully cover this situation either, lets say there are multiple local databases, one on an internal hard drive and one on a usb drive. the api would return both the internal drive and usb drive paths, the notify_settings method would then return True for both possible alternatives, however it would return True (and store the settings) for whichever path came first in the list which may or may not be the correct one. In this situation it's impossible to know which is the correct database path, the only way to know would be to return the list of valid database paths to the user and let them select the correct one. however I suspect this is an edge case and likely not worth the time it would take to implement.

racquemis:
For people with a single there is indeed no need for a local/network field in the interface. i however have multiple databases (i suspect more people do) local and network and therefore i added that option. Your correct that in case of the multiple local or network shares only the first one is selected. i have however a item on the interface that allows you to skip the first x entries. now i could just delete the dbloc thing and just use the dropdown menu to skip entries, but that isn't really a user friendly solution.

kernow:
In that case maybe the best solution is to gather info on all available database paths. If there is only 1 valid path simply store that. If there are multiple valid paths return the list of paths to the UI and let the user select the correct path. It would then be possible to get rid of the local/network field and the option to skip entries.

PR #594, opened 2013-02-20
Make episodes which have no airdate set to UNAIRED.

Zariel:
Currently shows which have no air date can be set to wanted or such, fix it to make the appear as if they have not aired.

### DIFF
--- a/data/interfaces/default/displayShow.tmpl
+++ b/data/interfaces/default/displayShow.tmpl
@@ -127,6 +127,12 @@ Change selected episodes to
     
 #for $epResult in $sqlResults:
 
+#if int($epResult["airdate"]) == 1
+    #set $status = $UNAIRED
+#else
+    #set $status = $epResult["status"]
+#end if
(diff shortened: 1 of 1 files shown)


Smenus:
Doesn't this prevent unaired specials from being set to wanted?

I admit I haven't tried your changes, but it seems to me that this would prevent things like DVD specials (which sometimes do get added to TheTVDB, Adventure Time unaired pilot for example) from being manipulated. While I agree that it's annoying that I can set episodes that are going to air, but haven't been given a date yet to wanted, I'm not sure that this is the best way to go about it.

thezoggy:
episodes that have no airdate shoulndt be unaired. this is by design. if they did air..they should have an airdate. populate tvdb with the data.

Zariel:
This only changes the display in the displayShow template, I changed it now to show specials how they were shown before, ie they will never be set to UNAIRED if they have an airdate of never.

PR #597, opened 2013-02-24
Change hashbang lines to use /usr/bin/env python2

pums974:
This makes it compatible with systems with python3 by default

### DIFF
--- a/autoProcessTV/hellaToSickBeard.py
+++ b/autoProcessTV/hellaToSickBeard.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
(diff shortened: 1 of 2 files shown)


thezoggy:
new code goes into development first not master

pums974:
> new code goes into development first not master

Sorry for that

> sb's code isnt written in the syntax of python3. only python 2.x.. so this seems a bit pointless of a submission

I understand that, but on my system, the default python is python3 so in order to use sb, I have to call it with python2 which is not a problem.
The problem appears when sabnzb try to execute the sabToSickBeard.py script, problem which can easily be addressed by modifiying the hashbang.

thezoggy:
sure thing, and notice i deleted the comment right after it posted as i investigated the actual change.

PR #598, opened 2013-02-24
Added New Provider for DailyTvTorrents

dzerovibe:
I've added full support for the DailyTvTorrents.org RSS feeds.
- All the major RSS Feed parameters are supported and configurable (eg. quality preferences, minage, wait, norar, single)
- The Scheduled Search uses the default feed (combined feed of new episodes)
- Backlog searches work using the per-show feeds
- Includes the ability to override the show name used for the queries (on the show Add and Edit pages), since DailyTvTorrents doesn't currently support querying by tvdb_id


thezoggy:
new code goes into development not master. also you should squash your commits to have a cleaner history.

thezoggy:
we use .png for provider images not .gif

data/interfaces/default/displayShow.tmpl
```diff
@@ -86,6 +86,16 @@ replace with: <b><%=", ".join([Quality.qualityStrings[x] for x in bestQualities]
 #end if
     </td></tr>
     <tr><td class="showLegend">Language:</td><td><img src="$sbRoot/images/flags/${show.lang}.png" width="16" height="11" alt="" /> $show.lang</td></tr>
+
+#if $sickbeard.DAILYTVTORRENTS:
```
thezoggy, line 90: why this hack?

data/interfaces/default/editShow.tmpl
```diff
@@ -51,6 +51,20 @@
 
 <form action="editShow" method="post">
 <input type="hidden" name="show" value="$show.tvdbid" />
+
+#if $sickbeard.DAILYTVTORRENTS:
```
thezoggy, line 55: you should handle this on the backend.. not force users to figure it out themselves..

thezoggy:
overall way too many 'hacks' for just a rss feed. i say no to pulling this.

sickbeard/tv.py
```diff
@@ -67,6 +67,7 @@ def __init__ (self, tvdbid, lang=""):
         self.paused = 0
         self.air_by_date = 0
         self.lang = lang
+        self.dailytvtorrents_show_name = ""
```
thezoggy, line 70: providers shouldnt have to have their own hacked name. this is a horrible approach

junalmeida, line 70: i agree with @thezoggy

dzerovibe:
Firstly, sorry for requesting the pull into the wrong place, I haven't done any collaborative work on projects like this before.

As for the hackiness, I agree, it's not ideal, but I did what I felt I had to in order to make this work with their feeds. I find their feed system to be much better than others, and I really wanted to get this working with SB.

As it is now, it works exactly as it should, and I don't think there are many shows where a user would have to override the show's name for the feeds (only 1 of my shows so far), but I'm guessing you probably don't like this approach because it intrudes to much into the core (eg. a new field in the DB). I'll keep looking for other ways to accomplish this. It would be better if DailyTvTorrents supported querying by TVDB ID, and I plan to ask them if that would be possible.

dzerovibe:
BTW, would this have a good chance of being pulled in if I took out all the stuff that handles the special show names? If so, I'll create a new pull request without that.

thezoggy:
providers dont all agree on show names.. this is nothing new. you can see how we used to deal with it for newzbin/nzbmatrix and how we deal with it for eztv. your solution is way to evasive and would make for crappy code to maintain in the long run. not saying that the need isnt there.. just the tactic you are taking is wrong.

dzerovibe:
Yeah, fair enough. I'll take another stab at it. Just had a quick look, and I assume you're referring to using the Scene Exceptions list? Looks like I can work with that to get the job done.

PR #600, opened 2013-02-24
Notifiers: XBMC update

thezoggy:
- Added GUI/Config option to update only the first host in xbmc host list. This should appease everyone now.
- Added SB logo to notifications sent via the JSON-RPC (xbmc12)
- Fallback to testing the legacy http-api before just giving up on the host:ip -- should fix those running xbmc on xbox/non offical builds
- Fix the json update_library routine on updating shows with `,` in the path
- Cleaned up configNotifications.js to make jsLint/Hint happy as well as update code for jquery 1.9x as `.complete` is depreciated in favor of `.done` for the $.get calls.
- Fix merge conflicts that the NMJv2 pull created -- and cleaned up misc things with it in the GUI


thezoggy:
@midgetspy be sure to remove the logo image reference to my site and just use the github one in xbmc.py (only left mine in there so you could test with)

PR #599, opened 2013-02-24
Add notifier that executes custom command when episode is added to library

Remboooo:
This plugin enables the execution of a custom command (as the sickbeard user) when an episode file is added to the library, to allow for easy scripting of custom actions (I use it for notifications on IRC and notifying a custom Android app, for example).

There are ofcourse security risks involved; that's why this plugin won't work without the user explicitly editing a line in config.ini. This is clearly stated, including the risks involved, on the configuration page.

On the to-do list would be adding the ability to execute custom commands on other types of events, as well as customizing the (order of) the command-line arguments.

PR #603, opened 2013-02-26
Notifier that executes a custom command when an episode is added to library

Remboooo:
Cleaned up pull request https://github.com/midgetspy/Sick-Beard/pull/599.

Similar to extra_scripts, I developed this plugin while unaware of that configuration option. The problem with extra_scripts is mostly that it is invisible to the user unless he knows what he is looking for.

The best solution might not be this pull request per se, but maybe combining extra_scripts with this frontend to provide visibility to the user looking for such an option.


Remboooo:
Messed up again.. argh. Ofcourse this pull request should be for the development branch.

PR #608, opened 2013-02-28
Add priority support for nzbget 9.0+

Prinz23:
I added priority support for nzbget 9.0 and higher, old version will use the old command.

I also added exception handling when connecting to nzbget to prevent the module from crashing if something goes wrong when trying to connect to nzbget.


thezoggy:
guess this has issues?

PR #526, opened 2012-11-11
Made checks for x264, h264, h.264 more liberally matching

sirchia:
I required these modifications to the quality checkers to be able to download 'The legend of Korra' using NZBMatrix as a search provider.

PR #611, opened 2013-03-01
Made checks for x264, h264, h.264 more liberally matching

sirchia:
Another attempt since the last request became too messy.

Please refer to closed pull request for details.
https://github.com/midgetspy/Sick-Beard/pull/526

Hope you will incorporate these minor improvements on the regular expression matching for 
quality profiles as they tolerate more diverse NZB naming styles without increasing the probability for false positives.

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -137,17 +137,17 @@ class Quality:
 
         checkName = lambda list, func: func([re.search(x, name, re.I) for x in list])
 
-        if checkName(["(pdtv|hdtv|dsr|tvrip).(xvid|x264)"], all) and not checkName(["(720|1080)[pi]"], all):
+        if checkName(["(pdtv|hdtv|dsr|tvrip).(xvid|[xh]\.?264)"], all) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDTV
-        elif checkName(["(dvdrip|bdrip)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
+        elif checkName(["(dvdrip|bdrip)(.ws)?.(xvid|divx|[xh]\.?264)"], any) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDDVD
-        elif checkName(["720p", "hdtv", "x264"], all) or checkName(["hr.ws.pdtv.x264"], any):
+        elif checkName(["720p", "hdtv", "[xh]\.?264"], all) or checkName(["hr.ws.pdtv.x264"], any):
             return Quality.HDTV
-        elif checkName(["720p", "web.dl"], all) or checkName(["720p", "itunes", "h.?264"], all):
+        elif checkName(["720p", "web"], all) or checkName(["720p", "itunes", "h.?264"], all):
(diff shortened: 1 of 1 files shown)


thezoggy:
last time someone proposed this.. it was matching too much spam/bad releases...  do you have a reason/example that this needs to be modified?

sirchia:
Yes, as mentioned in the discussion of the previous pull-request, I needed it to get a complete download of 'The legend of Korra'. Unfortunately I was using mostly NZBMatrix at the time, which as you may know is down now so it may be harder now to verify. But in a nutshell, these are the relaxations that I have implemented:
Change 1:
Pre-change:
  only x264 or either h264 and h.264 when its from itunes
Post change:
  x264, x.264, h264, h.264 from any source
Rationale:
  Inlcuding a 'dot' is more a manner of taste and x264 is the open source library name implementing the h.264 -> no indication for a less trustworthy source 

Change 2:

Pre-change:
  web.dl, web-dl or anything with one character between web and dl
Post change:
  web.dl, webdl, web-dl or anything else with web in it
Rationale:
  I don't know for sure anymore, but I believe there was one episode missing after the 1st change and a perfectly good version only included WEB as a source indicator (which was indeed a WEB-DL quality upload)

Hope you can agree to this motivation :)  

Edit: advertised markdown did not show table, restyled for readability

thezoggy:
you do realize that a . isnt a `.` in regex. a . would be a `.`. the . means ﻿﻿any character except a newline

thezoggy:
please show me specific release names that need this change.

sirchia:
Yes, I know what a . means in a regex :)
I had one example in the previous pull request:
"The Legend Of Korra S01E07 The Aftermath 720p HDTV h264 OOO"

While this is a perfectly valid HDTV release, it was not found because your regex did not expect h264 to be used outside of very specific itunes releases.

As NZBMatrix is down, I cant verify the other names of releases that gave me problems but would be something like:

"The Legend Of Korra S01E10 Some title 720p WEB h264 OOO"

bruvv:
any update on this one?

sirchia:
I think the proposed change has been well explained, is quite minor and does not allow much room for false positives/spam/bad releases. I hope the devs will incorporate them soon.

thezoggy:
last time we looked into doing this change it introduced bad releases.. why i asked for some valid things that sb is missing because of this. (if we did foreign content..yes i would agree it is needed)

bruvv:
I have my own build so that is why I asked, I use it now and it's working better then the default SB settings (I was missing x264 releases)

depassp:
I like this commit.  Here are a few samples for which this change would be needed:

Criminal Minds S08E19 720p WEB-DL DD5 1 H264-NFHD
Top Gear US S03E15 1080p WEB-DL H264-NTS
Top Gear US S03E15 720p WEB-DL H264-NTS
mad men s06e02 1080p web-dl h264-nts
Glee S04E18 1080p WEB-DL DD5 1 H 264-ECI
Legit S01E13 720p WEB-DL DD5 1 H 264-BS

thezoggy:
@depassp those are already caught by sb... no change is needed

PR #612, opened 2013-03-01
Enhancement :: Config - Notifier - Plex : Expanded test functionality.

Conjuro:
If someone wishes to update the Plex library, but don't want notifications, they only need server host; In which case the test button will error since it only tests client host connection.

This patch adds test function for server host and allows for empty client host value (or server host).

I've tested all failed connection scenarios, but not successful ones since I don't have a Plex installation to test with.


Conjuro:
Rebased and squashed for current development branch.

PR #613, opened 2013-03-01
Fix update via tarball for GITHUB recent url changes

mr-orange:
Github have changed something in tarball url adding some parameter like expiration date. The changes broke updating features via Tarball. This pull request fix the error during update process.


midgetspy:
Thanks for the contribution.

PR #615, opened 2013-03-01
Metadata change

luxmoggy:
The code for the new version of the metadata. Can be used to generate the XMBC in both the new and old format.

Only the changes in the following files are needed:
sickbeard/**init**.py
sickbeard/tv.py
sickbeard/webserve.py
sickbeard/image_cache.py
sickbeard/metadata/**init**.py
sickbeard/metadata/generic.py
sickbeard/metadata/mediabrowser.py
sickbeard/metadata/ps3.py
sickbeard/metadata/synology.py
sickbeard/metadata/tivo.py
sickbeard/metadata/wdtv.py
sickbeard/metadata/xbmc.py
sickbeard/metadata/xbmcorig.py
data/interfaces/default/config_postProcessing.tmpl
data/js/configPostProcessing.js

tvdb_api & checkVersion (have changes in them that are nothing to do with this change - don't know how to exclude them) or squash commits sorry

Have tested the generation of metadata providers. config.ini is updating correctly.
The PostProcessing screen is showing the new options.

The code for Season Fanart isn't working and is skipped in all Metadata providers

### DIFF
--- a/data/interfaces/default/config_postProcessing.tmpl
+++ b/data/interfaces/default/config_postProcessing.tmpl
@@ -469,7 +469,6 @@
 #set $cur_id = $GenericMetadata.makeID($cur_name)
 <div class="metadataDiv clearfix" id="$cur_id">
     <div class="metadata-options-wrapper">
-                <!-- IRC: TO BE DONE: Edit new metadata opsions here-->
         <h4>Create:</h4>
         <div class="metadata-options">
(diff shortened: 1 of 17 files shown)


thezoggy:
not sure why you need to do all these changes and not just add another provider or extend an existing one with with overrides.. then you also have tvdb hacks in this and other misc things.. along with dirty commit history. this will not be pulled

luxmoggy:
Trying to remove commits from the branch and redo the commits in one. So will close this and open it again.

The reason that I change all of the files is that generic is based on the old XBMC format and I wanted to extend this to the new XBMC format and have the same ability to change configuration that is currently there. e.g. turn on and off season banners or season posters. Which can't be done with a new provider as the is only the option season thumbnails yes or no (4 possible setting vs 2).

PR #620, opened 2013-03-04
fix bug where xbmc update only first config did not save

EchelonFour:


thezoggy:
this is already part of the code, what branch are you using ?

EchelonFour:
That's weird. Never mind.

PR #621, opened 2013-03-04
update for new metadata using new XBMC as base for generic.py

luxmoggy:
The code for the new version of the metadata. Can be used to generate the XMBC in both the new and old format.

Have tested the generation of all metadata providers.
config.ini is updating correctly.
The PostProcessing screen is showing the new options.

The code for Season Fanart isn't working and is skipped in all Metadata providers

The reason that I change all of the files is that generic was based on the old XBMC format and I wanted to extend this to the new XBMC format and have the same ability to change configuration that is currently there. e.g. turn on and off season banners or season posters. Which can't be done with a new provider as the is only the option season thumbnails yes or no (4 possible setting vs 2).

hopefully better this time. git is proving hard to use...


thezoggy:
this can be closed as https://github.com/midgetspy/Sick-Beard/pull/789 handles this and more

PR #622, opened 2013-03-05
Updated init.freebsd (Shutdown didn't work)

nickbock:
Updated verify_sickbeard_pid() to include "--quiet --nolaunch" to the ps | grep.
Updated sickbeard_stop() wget called a incomplete url and kept waiting on de shutdown until it timed out.


thezoggy:
can you submit this against development branch so we can pull. all new code goes there. i'd rather you re-submit so you get credit vs us just modifying the code ourselves

nickbock:
Sure thing. I'm on it.

PR #624, opened 2013-03-06
Better handling Show Update Status "Ended"

Prinz23:
Since the "Ended" Status is set on the tvdb many times as soon as the TV Shows is canceled, there are still Episodes left to air. So stopping to do Updates for "Ended" Shows is not a good idea. I changed that.

Now the last known Airdate and the the Airdate of the last Episode is read from the db. If a Show is marked "Ended" the Show is still updated if:
1. The last Episodes Airdate is known = up to 90 days after the last known Airdate a normal Update is done.
2. The last Episodes Airdate is unknown = up to 365 days after the last known Airdate a normal Update is done.

Both is necessary to avoid the need of a manual Forced Update for Shows that are marked "Ended", but not all Episodes have been aired yet.

### DIFF
--- a/sickbeard/showUpdater.py
+++ b/sickbeard/showUpdater.py
@@ -23,6 +23,7 @@ import sickbeard
 from sickbeard import logger
 from sickbeard import exceptions
 from sickbeard import ui
+from sickbeard import db
 from sickbeard.exceptions import ex
 
 class ShowUpdater():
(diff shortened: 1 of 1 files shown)


thezoggy:
funny, i was just talking to midget last week about changing this. my solution is to do a force update on the show on each backlog. that way if a show got marked ended accidentally we would catch it (granted upto 3 week grace period)

thezoggy:
ill have to go dig up my notes and posted it back here, but its a simple one line fix

Prinz23:
But that solution would only update the show every 3 weeks and that can be to late in the day of dcma. And the postprocessing for new episodes wouldn't work correctly within the 3 weeks.

And the Shows aren't marked accidentally ended, the show is ended (this status won't change because of the update). Only not all episodes are aired yet in some cases.

thezoggy:
The show status is not used during pp. 

-----Original Message-----

Prinz23:
The Episodes aren't updated without this patch. So it's a unknown Episode and it doesn't know the Name of the Episode. 

So yes the status is not used, but the new Episodes aren't updated or even known without this patch.

thezoggy:
the episodes shouldnt be updated until they have the required data filled in..  this is by design to stop populating episodes that are wrongly added to tvdb/duplicate/errors/etc.

Prinz23:
So what you are saying basically is:

It's better to don't handle a Show as normal update just because the Field Status has the value "Ended", even that in many cases the Show still airs.

I don't see any logic in just stopping the update, because the status Field that doesn't say more then that the Show is Canceled (not that all Episodes have aired) to stop the update.

What is the harm in doing what this patch does? 
It does what Sickbeard would do if the Show would only be changed to Ended after all Episodes have aired.
The current way does harm, because it assumes the Ended Status means that all Episodes have aired and that is simply not true in many (most) cases.
Not to mention that Shows sometimes aren't changed to Status Ended for months or even years after the show has ended. So according to your logic this would mean that Sickbeard does currently harm (something wrong) with such Shows.

But in the end it's not my decision how logical or unlogical Sickbeard handles things. So will wait what's decided.

thezoggy:
no, im saying that we dont need to hammer tvdb on updating a show if there 'shouldnt' be any updates. that was the whole point of not updating on ended. the problem is that this gets set early / show gets resurrected.. and we would never know (unless someone force full update). my solution was just to do the force full update when we do the actual backlog (3 weeks out) or if someone uses -f startup argument. once the show is updated.. then we would see the correct status/airdates/etc.

Prinz23:
> updating a show if there 'shouldnt' be any updates.

And this is a wrong assumption. Because "Ended" doesn't mean that, it only means that the show is canceled, nothing more. Everything else is a baseless assumption.

thezoggy:
taking your ball home and not playing with others is quite childish way to deal with a conversation that doesn't go your way.

Prinz23:
It would have been closed anyways. So what is the point not closing it myself?

thezoggy:
because you assume that it wouldnt have been pulled? i dont make that call.. midget does.

PR #625, opened 2013-03-06
Add missing runtime from thetvdb

Prinz23:
The runtime wasn't imported from the tvdb.

Is Field will be needed for new timezone sensitive Coming Episodes Page. If the time (localized start time + runtime) is later as now the TV Show will be shown as Missed.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -654,6 +654,9 @@ class TVShow(object):
         if myEp["status"] != None:
             self.status = myEp["status"]
 
+        if myEp["runtime"] != None:
+            self.runtime = int(myEp["runtime"])
+
         if self.status == None:
             self.status = ""
 


thezoggy:
can you check to see if this handles bad data from tvdb (where the runtime isnt an int) cant recall if the tvdb_api handles it or we need to catch for it

Prinz23:
You can only enter a number at the tvdb website. It's a drop down field with predefined numbers at the thetvdb.

Prinz23:
It's not the only int convert without a exception handling in the code. The best would be the use the function that i included in the newznab.py pull request and put that function in the helpers.py:

``` python
    def tryInt(self, s):
        try: return int(s)
        except: return 0
```

and replace all the int(...) with helpers.tryInt(...)

PR #628, opened 2013-03-12
Enable users to change the frequency for Scan and Process

pyrocumulus:
The default interval for auto process is 10 minutes. This high frequency causes problems for users with Sick-Beard on NAS devices that want to use deep sleep, like Synology. Deep sleep is only possible when there is no hard drive activity, which will reset the inactivity counter. So every time SB does a scan on the hard drive, the device will wake up from deep sleep if it can go to sleep at all.

Being able to lower the frequency to something like 120 minutes or even less will solve this issue.

### DIFF
--- a/data/interfaces/default/config_postProcessing.tmpl
+++ b/data/interfaces/default/config_postProcessing.tmpl
@@ -92,6 +92,17 @@
                             </label>
                         </div>
 
+                        <div class="field-pair">
+                            <label class="nocheck clearfix">
+                                <span class="component-title">Scan and Process Frequency</span>
+                                <input type="number" name="auto_post_process_frequency" value="$sickbeard.AUTO_POST_PROCESS_FREQUENCY" size="5" min="1" class="input-small" />
+                            </label>
+                            <label class="nocheck clearfix">
+                                <span class="component-title">&nbsp;</span>
+                                <span class="component-desc">Time in minutes between scans (eg. 10)</span>
+                            </label>
+                        </div>
+
                         <div class="clearfix"></div>
                         <input type="submit" class="btn config_submitter" value="Save Changes" /><br/>
 
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -170,6 +170,7 @@ ADD_SHOWS_WO_DIR = None
 CREATE_MISSING_SHOW_DIRS = None
 RENAME_EPISODES = False
 PROCESS_AUTOMATICALLY = False
+AUTO_POST_PROCESS_FREQUENCY = 10
 KEEP_PROCESSED_DIR = False
 MOVE_ASSOCIATED_FILES = False
 TV_DOWNLOAD_DIR = None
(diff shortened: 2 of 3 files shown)


thezoggy:
ty for the submission, i can see how some people would want this. what if someone tried to set the value to 0 (or -# or not even an int)? i dont see any 'dummy' proofing to catch things.

in the future you should just rebase and push, as it would keep the commit clean. now if we pull it will have two commits vs just the one we really need. not a huge deal currently. but if your pull sat there for 10 weeks and you were pushing a merge sync to it every day it would result in quite a bit of useless noise

pyrocumulus:
@thezoggy thanks for your feedback. Well for as far as I saw, the sanity checking is done by the template (Cheetah?). At first I had 'min="10"' in the template and I couldn't set any value lower than 10; therefore I set 1. I think sanity checking is already functional. However, I'll double check this when I get home.

About the rebase and push; I'm terribly sorry about that. I haven't thought of that. Truth be told, this is actually my first encounter with Git and GitHub. Lot of software development experience, but zero Git. I'll do some reading on rebase. But I understand that doing rebase after making the pull request is not done. Do you want me to Close this pull and make a new request with rebase?

thezoggy:
I'll have to check on the .tmpl page myself since I was only glancing at the submission. The rebase/push was just for future reference or if there are commits you need to do. No need to close this (so far)

-----Original Message-----

pyrocumulus:
Okay :) Well I did some extra testing (sanity wise) and it appears to almost be covered. I tried 0, alphabetic and weird characters, all is well.

I did however find that leaving the box completely empty results in an exception. But I found that leaving the Retention box empty for example, appears to have the same problem (although that keeps on loading, instead of showing an error). I'll fix that in a separate branch.

For now; I can commit an updated version of this change that will also handle an empty input box. 

edit: I have another commit ready for the above problem. You want me to push that commit as well? Or can I somehow merge all those commits into a single new commit?

thezoggy:
you can squash and commit locally. then rebase if you need to want to rewrite history then force push to update this pull request

thezoggy:
note that solely relying on html5 input is not ideal. it is extremely easy to bypass this. plus you have to be mindful that not all browsers do things the same.

http://caniuse.com/#feat=input-number

thezoggy:
about the usenet retention, we pass the 'default' if nothing was entered. 

from last nights conversation when i took a glance into this:

```
[11:17pm] <@_zoggy_> so noticed today in init we do
[11:17pm] <@_zoggy_>         USENET_RETENTION = check_setting_int(CFG, 'General', 'usenet_retention', 500)
[11:18pm] <@_zoggy_> but on the savesarch screen we do
[11:18pm] <@_zoggy_>         if usenet_retention == None:
[11:18pm] <@_zoggy_>             usenet_retention = 200
[11:18pm] <@_zoggy_> prob should make those match?
```

so either need to make it pass the default if none is set or catch for none and imply the default

pyrocumulus:
I will try to do so :-) thanks for the tips.. getting started with Git
isn't as easy as I expected. But I plan on doing a lot more work for SB.

About relying on html5: I'll read up on that and see how sanity checking
can be improved for all the inputs.

pyrocumulus:
Okay, rebase worked! This commit is now fully up to date and includes protection against an empty field.

@thezoggy 
Yes if nothing has been entered in the retention you try and save the default. However, when you press Save on the page something else goes wrong. It doesn't refresh and the loading indicators keep spinning. This is probably because:

```
if usenet_retention == None:
  usenet_retention = 200
...
sickbeard.USENET_RETENTION = int(usenet_retention)
```

The if-statement doesn't catch an empty string, so the default won't be set. So when the second line is hit, it fails on converting an empty string to an integer. This is probably why the page stops processing further and the configuration isn't actually saved.

But those are worries for another branch :) The html5 thing is also broader than this push, but I have to do some more reading up on that. And testing on how it can fail on other browsers.

thezoggy:
your commit history looks odd.. did you pull and push from dev?

pyrocumulus:
I don't know what I did wrong, but when I look at it it does look odd.
Somehow your commit to midget's repo appears extra :o just when I thought I
had it.

I did pull from development but I guess I did something wrong with rebase.
Only my own commits are merged apparently. I'll see if I can fix it
tonight.

pyrocumulus:
Well I fixed it using rebase, again. I think it's good now, right? Only my commit and the diffs look good. I still do not exactly know what I did wrong.

Could you do me a small favor and tell me how you keep your topic branches up to date with the upstream?

pyrocumulus:
Very much outdated

PR #631, opened 2013-03-15
Enhancement :: Logging : Compression (gzip) for archived log files

Conjuro:
Adds compression to rotated log files.

Primary log file remains clear text, archived/rotated (.#) logs are stored in compressed form (.gz).

When log rotation is called, the newest archive log (.1.gz) is compared to LOG_SIZE variable. If size not yet reached then primary log is appended, else logs get rotated.

One note of clarification. A few decompression tools do not support multiple gzip blocks in same file, and will only decompress first block. Most (common) tools do support it. The alternative would be to decompress, append and then compress again, which involves more disk trashing, temporary files, CPU and memory usage at no significant (in my opinion) gains.

On another note. I'd like some feedback on whether I should make compression a config option? And if so, just config.ini or web GUI also?

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -23,6 +23,8 @@ import threading
 
 import logging
 
+import gzip
+
 import sickbeard
 
 from sickbeard import classes
@@ -96,7 +98,7 @@ class SBRotatingLogHandler(object):
         
         i: Log number to ues
         """
-        return self.log_file + ('.' + str(i) if i else '')
+        return self.log_file + ('.' + str(i) + '.gz' if i else '')
     
     def _num_logs(self):
         """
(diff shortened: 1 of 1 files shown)


Conjuro:
Forgot to mention why I made this patch. Sick Beard is quite "chatty", which is a good thing for debugging reasons. However, it does mean it can run through log files quite fast and since Sick Beard is meant to be fairly automatic, problems might not be discovered before the relevant log entries are gone from the logs.

Since log entries compress quite well, one 10 MB file can be compressed down to about half a MB. Meaning we get 610 MB worth of log entries, rather than just 40 MB. (Assuming default LOG_SIZE=10 and NUM_LOGS=3).

thezoggy:
honestly the whole flat file approach of logs really needs to go..  the constant writing of a file.. the unicode concerns.. the horrible way to filter/pull any data out of it.. the whole thing about it seriously needs to be redone. ideally id like to do something similar to how unraid does their logs. more on this later

Conjuro:
`unraid` as in http://lime-technology.com/?

I'm not familiar with it. At a cursory glance, they seem to be using syslog? That would be a problem on windows machines since it's not native (though there are windows version available for install), or were you thinking of using syslog for *nix OS's and event handler (win32evtlogutil) on windows?

unRAID syslog seems to write to a flat file log also, so it might not have been what you were thinking of.

At any rate, you wrote you'd have more on it later, so I'll stop speculating for now and be patient. :)

Conjuro:
Rebased for current development branch. Been using for three months without issue.

thezoggy:
now that im back from vacation, ill take a look at this _soon_

PR #634, opened 2013-03-20
Updated init.freebsd

nickbock:
- Removed the use of daemon to daemonize Sickbeard and made more use of
  the default functions of FreeBSD and Sickbeard.
- Also removed the need to edit the init file. The file gets it's settings
  directly from the sickbeard config file. All configuration is done trough `/etc/rc.conf.local` or `/etc/rc.conf`. You only need to  add `sickbeard_enable=YES` and an optional user `sickbeard_user="someuser"` in `/etc/rc.conf.local` or `/etc/rc.conf`. For more configurable options check the file itself.
- Changed the `sickbeard_stop()` and `sickbeard_status()` functions to be a bit more resilient when
  faced with strange situations.

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -1,24 +1,28 @@
 #!/bin/sh
 #
 # PROVIDE: sickbeard
-# REQUIRE: DAEMON sabnzbd
+# REQUIRE: wget
 # KEYWORD: shutdown
 #
 # Add the following lines to /etc/rc.conf.local or /etc/rc.conf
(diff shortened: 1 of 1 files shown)


badboybeyer:
I like the improvements. Could you also add support for non-default web roots? The configuration option is:

```
web_root = /
```

thezoggy:
this also doesnt work for users using https

ghost:
superceded by #908

PR #635, opened 2013-03-21
Add timezone correction to the Coming Episodes Page

Prinz23:
It's automatically corrected via the dateutil lib. 

The local timezone is detected and the show timezone is detected from the network. The network timezone table is loaded and updated with every version check from https://github.com/Prinz23/sb_network_timezones

Also the zoneinfo of the dateutil is also automatically updated from that url if needed.

And I added the the possibility to set the missing days shown on the Coming Episodes Page in the Config -> Generals Menu. 2 day to 30 days allowed.

I added runtime import from the tvdb, because a show will be marked as missed after airtime + runtime is over. 

All int convert functions are replaced with a new helpers function helpers.tryInt.


kungktd:
I really hope this gets into SB. I requested this as soon as I started to use SB. 
Great work Prinz23!

simonk83:
Hallelujah !!

Prinz23:
> what happens if github is down? can we disable the use of the timezone?

It's displayed like now, without timezone correction. Same if the Network is not known or wrong or timezone for network is unknown, ... And network timezone info is saved in the cache.db like the scene exception in a new table. The table is only updated if needed.

Of course it's displayed on the correct day and many other fixes were currently the displaying is simply wrong.

And fixed offsets don't work. Timezones change multiple times a year!!! Most of the time DST dates change. And this code is only executed if a user clicks on the Coming Episodes link, so there is no load on the system if the user doesn't look at the Coming Episodes Page.

You can lookup how often timezones change:
ftp://ftp.iana.org/tz/releases/

...
2010: 15 times
2011: 14 times
2012: 10 times
2013: 2 times until now

drewfreyling:
+1
I also dislike the reliance on github here.

Prinz23:
Scene Exceptions relies on github for years now, the exact same way. And that is a much more important function, this only affects the displaying, nothing more.

What up with the double standards?

rolo20:
It works! Thanx

drewfreyling:
@Prinz23 Wasn't aware of the current reliance on it. I say we should go for it then and if we don't like the reliance on it we remove it everywhere not just in this instance.

Prinz23:
The data has to come from somewhere. github is as good as anywhere else.

So removing is impossible or one has to remove the functions completely. Which would mean that shows with scene exceptions wouldn't be downloaded at all anymore.

simonk83:
My most wanted feature for a very long time now.  Hoping for a quick pull.

sickbeard/helpers.py
```diff
@@ -24,6 +24,7 @@
 import shutil
 import traceback
 import time, sys
+import hashlib
```
thezoggy, line 27: we do this for when we generate the md5 apikey, try the newer hashlib but fallback to the legacy md5 (pre python 2.5)

```
        try:
            from hashlib import md5
        except ImportError:
            from md5 import md5
```

Prinz23, line 27: md5 is only needed for pre python 2.5. 

Since the minimum requirement for Sickbeard to run is python 2.5, this is useless

thezoggy, line 27: yes, noticed that i updated the comment from my original post as i researched it once I made the comment to see why we do it (as it was taken from sab).

greenythebeast:
Is it possible to get airtime added to the List layout of the coming episodes page?

Prinz23:
The Airtime is already displayed in the list layout as long the OS system setting is set to display it. It's displayed with the python datetime format function strftime in "Locale’s appropriate date and time representation."

mr-orange:
Really good work on this Prinz23. I've added this in my dev branch. Also insert 2 line of code for proper air time string formatting independent from locale setting.

Prinz23:
> Also insert 2 line of code for proper air time string formatting independent from locale setting.

I didn't add this on purpose, because I live in a 24 clock country like many people and don't like (find it even confusing) 12 clocks.

I'm thinking about adding a option in config to set the format for date and time. If people want it.

sickbeard/network_timezones.py
```diff
@@ -0,0 +1,166 @@
...
+
+# update the network timezone table
+def update_network_dict():
+
+    _update_zoneinfo()
+
+    d = {}
+
+    # network timezones are stored on github pages
+    url = 'http://github.com/Prinz23/sb_network_timezones/raw/master/network_timezones.txt'
```
arogl, line 115: Any reason you have chosen Perth for the Australian times? Sydney would be better.

mr-orange, line 115: I do not choose anything about timezone. You have to ask to Prinz23

Prinz23, line 115: I had to guess the correct timezone for country's with multiple timezones. Only for US and Canada I knew the correct timezone.

I have to open a Thread in the Sickbeard Forum where people can post and correct wrong timezones and request new networks too. Just like the scene exception Thread in the Forum.

Prinz23:
I opened a Thread in the Forum where new or incorrect Timezones for Networks can be reported/requested. Just like the ADD Scene Exception Thread.

http://sickbeard.com/forums/viewtopic.php?f=4&t=7007

ChrislyBear-GH:
Best fork I've used so far! Please pull it soon, cause this feature is awesome.

A note to Prinz23:
It's very nice, that you offer so many different date formats. Since I'm missing the ISO 8601 date format (yyyy-mm-dd) I have to ask if it would be possible to offer an option to directly enter a format string? 

Chris

mattymess:
I cloned this repo, but don't see where these changes are in 'Config > General.' Do I even need to do anything in the Sickbeard admin?

ChrislyBear-GH:
I believe you didn't checkout the right branch.
1. Terminate SickBeard
2. Execute `git checkout timezones-coming-eps` inside the directory in which the cloned repo resides.
3. Start SickBeard again.

Chris

mattymess:
I cloned Prinz23's fork so I didn't think I'd need specifically check out that branch, but I will try and let you know. Thanks for the reply, KurrKurr.

mattymess:
That worked like a charm. Thanks @KurrKurr!

threesquared:
Why was this request closed? I would really like to see this in the master branch.

simonk83:
Absolutely

ChrislyBear-GH:
Yes, why the closing? It hasn't been merged yet!?

mr-orange:
@Prinz23 
I really liked your work on timezone integration. And it's really a pity you have closed this.
Please do not delete the branch and if you want I and other would really happy to have a pull request on my branch.
Thanks.
Mr_orange.

Prinz23:
I will not delete it.

I simply will not work on anything thetvdb (for example: Sickbeard) related anymore.

Stanzilla:
@itofzo can this get merged, please?

Prinz23:
The Pull-Request is very old and many bugfixes are missing in this very old request.

Stanzilla:
@Prinz23 care enough to re-do it?

Prinz23:
There are many reasons why I don't bother.

And additionally: When was the last time that any pull request was merged in this Repo? (Not counting the main dev's, thezoggy)

thezoggy:
Stuff gets pulled, usually we pull it into our own branch to fix issues first before pushing to the repo.

-----Original Message-----

ppslim:
Granted, but often attribution / credit is lost in the process, although the licensing used does obviously support this :(

thezoggy:
Incorrect. If the pull is pulled into a branch and pushed.. The history and 'credit' is maintained.. There are things I've ported over from another branch (for example prinz nzbget changes) which we credit even though I end up rewriting or changing due to the branch differences... The credit is still there. 

-----Original Message-----

Stanzilla:
Anyway, would be nice if one of you guys can redo this patch against master.

Prinz23:
If at all, I would have to port the mr_orange branch (no longer existing) here, since I fixed a lot of things in that version.

Stanzilla:
It's kinda sad that there have to be so many branches and forks of the same thing. I guess it comes to what you said, not enough stuff getting merged from the pull requests.

thezoggy:
So you would rather SB be closed source? That way there wouldn't be forks or branches..

-----Original Message-----

Stanzilla:
Nope, just more working together with people, so more open with code audits etc.

PR #636, opened 2013-03-25
Update to fix paths no longer updating properly in Frodo.  Without this ...

mattsch:
...fix I

get the following error in xbmc:
WARNING: Process directory '"/path/to/show/"' does not exist -
skipping scan and clean.

Signed-off-by: Matthew Schick matt@excentral.org

### DIFF
--- a/sickbeard/notifiers/xbmc.py
+++ b/sickbeard/notifiers/xbmc.py
@@ -414,7 +414,7 @@ class XBMCNotifier:
                 return False
 
             logger.log(u"XBMC Updating " + showName + " on " + host + " at " + path, logger.DEBUG)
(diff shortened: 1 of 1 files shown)


mattsch:
Is it really necessary to support 12 when 12.1 is out for all platforms?  I can see backwards compat for older full versions but not for point releases.  I don't see an obvious way to pick out the version via json, IMO it's not worth the hack for an upgrade everyone should be doing anyway...

thezoggy:
last time i checked 12.1 wasnt out for all platforms.. will have to check since openlec and others are usually a bit behind

mattsch:
OpenElec is updated as are the official ubuntu ppa's.  Those are all that matters, right? ;)

thezoggy:
Those are the prob child ones (that people complained from earlier) so we are prob good to just revert the path escape hack. Just need to make sure it can still update shows with a coma or quotes in its name

-----Original Message-----

thezoggy:
ok we were in a time crunch so i just submitted a pull to revert what we had before, you had left in the appending of two empty strings.. which is uneeded/messy. ty for reporting though

PR #638, opened 2013-03-26
Added open-file by clicking on the filename in displayShow

rescribet:
Gives the user the possibility to choose an executable and enable the
function.
Once enabled the user can click on a filename of an episode in the
displayShow page to open the executable with the filename as a
parameter.
It should be extended to give the user the possibility to customize the
parameters given to the exec for broader compatibility.
Build on and tested with win8 and VLC, windows media player and mediaplayer classic.

### DIFF
--- a/data/interfaces/default/config_general.tmpl
+++ b/data/interfaces/default/config_general.tmpl
@@ -158,6 +158,32 @@
                     </fieldset>
                 </div><!-- /component-group2 //-->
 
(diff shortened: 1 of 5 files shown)


junalmeida:
Guess that most setups runs sb on different machines, I mean, the one sb is running is not the same I use to watch. I guess that this will not attend most of our users

PR #639, opened 2013-03-27
1080p migration tweak

thezoggy:
- Add some logging to each 1080p migration part so people see some sort of progress in their console
- Use show_id instead of tvdb_id for first part of migration, just in case there is a duplicate id or if show is missing tvdbid (never know)
- Optimize query by reducing the arithmetic performed each time and queuing the list of updates to perform using a transaction (thanks @Prinz23)

### DIFF
--- a/sickbeard/databases/mainDB.py
+++ b/sickbeard/databases/mainDB.py
@@ -27,6 +27,7 @@ from sickbeard.name_parser.parser import NameParser, InvalidNameException
 
 MAX_DB_VERSION = 12
 
+
 class MainSanityCheck(db.DBSanityCheck):
     def check(self):
         self.fix_duplicate_shows()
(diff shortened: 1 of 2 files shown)


thezoggy:
notes for future reference: 
- http://stackoverflow.com/questions/1711631/how-do-i-improve-the-performance-of-sqlite
- http://web.utk.edu/~jplyon/sqlite/SQLite_optimization_FAQ.html#transaction-wrappers

Prinz23:
BEGIN/END TRANSACTION won't work correctly with sqlite3 python. 

I use a mass_action funtcion (added to db.py) in my timezone coming episodes pull to fill the network table with all info i a second or so...

Prinz23:
My mass_action function does nothing else as use a transaction to execute any amount of sql queues  

I just updated a fix for the function in the timezone pull request. 

It's easy to use and to convert old calls to the action function to use this function. You just call that function with a python list with all the sql queues and the function executes it as one transaction. It's really fast in comparison.

thezoggy:
your `mass_action` function did exactly what i wanted and deff was a speed improvement. I've now included it as part of this pull. only using it for just this 1080p migration for right now..

Prinz23:
I improved the speed of the mass_action a bit for systems with low IO disk performance

PR #640, opened 2013-03-28
Optimize banner and poster

douglasdollars:
Following midgetspy/Sick-Beard#585, this commit is based off of the Development branch and only optimizes the poster and banner. As before, this only removes extra metadata and heft from the file, not actually affecting the image presentation at all.


douglasdollars:
Total savings about 37kb

Screen Shot 2013-03-27 at 9 45 01 PM

PR #647, opened 2013-04-02
Add uTorrent API support & ThePirateBay provider.

flex1986:
I add uTorrent API access similar as to SabNZBD, uTorrent also supports magnet links so we can now use thepiratebay provider from mr_orange.

Tested locally for thepiratebay and ezrss, I don't have access to the other torrent provider so I can't test those.


flex1986:
Moved my changes to the development branch, new pull request coming.

PR #649, opened 2013-04-04
Add SCC provider.

Poodlemastah:
Add a SCC provider.


Conjuro:
Pull requests are issued against the development branch, not the master, also you might wanna squash those two commits into one for the sake of a cleaner log.

PR #651, opened 2013-04-04
Add SCC provider

Poodlemastah:


wookienz:
any movement on this feature?

thezoggy:
parsing a webpage, https://github.com/midgetspy/Sick-Beard/pull/651/files#diff-e099a178555d7ca0f68f31a83ebbc42fR50
makes it very prone to breakage.. would make me question if this even works
as this is an old pull request and needs to be rebase off current development
also, we dont have a way to validate it works (no account).

PR #653, opened 2013-04-07
Added status command to ubuntu init file

LeonB:
Needed by puppet for example
- removed trailing space

### DIFF
--- a/init.ubuntu
+++ b/init.ubuntu
@@ -41,7 +41,7 @@ DESC=SickBeard
 ## otherwise default sickbeard is used
 
 ## The defaults
-# Run as username 
+# Run as username
 RUN_AS=${SB_USER-sickbeard}
 
 # Path to app SB_HOME=path_to_app_SickBeard.py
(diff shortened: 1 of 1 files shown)


LeonB:
No love? :(

thezoggy:
is this based off the development branch ?

init.ubuntu
```diff
@@ -106,9 +108,14 @@ case "$1" in
         start-stop-daemon --stop --pidfile $PID_FILE --retry 15
         start-stop-daemon -d $APP_PATH -c $RUN_AS $EXTRA_SSD_OPTS --start --pidfile $PID_FILE --exec $DAEMON -- $DAEMON_OPTS
         ;;
+
+  status)
+       status_of_proc -p $PIDFILE $NAME $NAME
```
ghost: This doesn't work because variable $PIDFILE doesn't exist, it's $PID_FILE

Anyway I am wondering if it's not safer to test: if [ -f /lib/lsb/init-functions ];
before importing and using it? For backward compatibility

thezoggy:
@LeonB can you correct this and force push

LeonB:
I've added the suggested changes. What did you mean with force push?

thezoggy:
yes looks better now. thanks

ghost:
There is no fallback (or check) in status if init-functions doesn't exist.
Thinking of something like this: 
https://github.com/itofzo/Sick-Beard/compare/midgetspy:development...add_status_init.ubuntu_init.style_dev

IMHO shorthand makes it harder to read (if you're not used to it.)
Force push: When you squash your commits and want to push again, you'll have to use force.

LeonB:
OK. Learned something :)

But why all the checks? This is the init file for Ubuntu and as far as I know `lib/lsb/init-functions` has been in Ubuntu since at least 10.04

ghost:
Probably isn't needed on most systems, but it doesn't hurt either :)
SickBeard still supports Python 2.5 too.

thezoggy:
@itofzo so this is good to pull now?

LeonB:
I think it is good to pull. I squashed all the commits into one commit.

If the extra lsb checks are desirable it can be pulled from [@itofzo's branch](https://github.com/itofzo/Sick-Beard/blob/add_status_init.ubuntu_init.style_dev/init.ubuntu)

PR #654, opened 2013-04-13
WindowsError is a subclass of OSError and need not be excepted

depassp:
This fixes bug 2254: https://code.google.com/p/sickbeard/issues/detail?id=2254

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -532,7 +532,7 @@ def delete_empty_folders(check_empty_dir, keep_dir=None):
                 ek.ek(os.rmdir, check_empty_dir)
                 # do the library update for synoindex
                 notifiers.synoindex_notifier.deleteFolder(check_empty_dir)
-            except (WindowsError, OSError), e:
+            except OSError, e:
                 logger.log(u"Unable to delete " + check_empty_dir + ": " + repr(e) + " / " + str(e), logger.WARNING)
                 break
             check_empty_dir = ek.ek(os.path.dirname, check_empty_dir)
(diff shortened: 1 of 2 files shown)


thezoggy:
While I can't find why WindowsError is thrown.. it is at times..
https://github.com/buildbot/buildbot/pull/540
and more info: http://www.mail-archive.com/python-list@python.org/msg266474.html

@itofzo added this code, maybe he knows more

depassp:
I suspect @itofzo added the WindowsError "just in case".

The point is, the bug reporter ran this on a Linux machine.  An exception was raised, probably OSError (probably from os.rmdir, although I'm not sure why).  Since WindowsError is not defined in Linux, Python raised a NameError.

Whether or not a non-Windows machine raises WindowsError is irrelevant since WindowsError is a subclass of OSError.  It is sufficient to catch OSError.

This fix means SickBeard will no longer crash completely.  Instead, users will get a log message "Unable to delete ...", at which point we might be able to better debug why OSError was raised.

ghost:
I didn't "introduce" this, so don't know why WindowsError was added.
AFAIK OSError should be enough.

thezoggy:
@itofzo git blame shows that you added it when you did your cleanup folder code.. why i assumed you added it

depassp:
You're right.  Looks like it was introduced by @midgetspy in https://github.com/midgetspy/Sick-Beard/blob/a8667b528e5181560cc4e788fd6134a90170170f/sickbeard/helpers.py#L469

thezoggy:
https://github.com/midgetspy/Sick-Beard/commit/a2d6e2f57007c93191b692eebc53a0282d12403d
is why git blame is showing itofzo, but digging back I see it was actually added under midgetspy's commit:
https://github.com/midgetspy/Sick-Beard/commit/a8667b528e5181560cc4e788fd6134a90170170f

so sorry about that itofzo :)

ghost:
No worries :)

PR #660, opened 2013-04-15
Issue 284: modify SickBeard to clean up and remove old episodes

kierse:
Modified SB to support defining episode manage settings per show. User can choose to keep all, the current season, or a range of 1-10 episodes. When an episode is identified as stale it's status is set to ignored and all associated media is removed from the filesystem.

This first pass solution does not:
- alert notifiers of removed media
- remove any files generated by the metadata providers
- clean up the edit show template

Note: I do plan on addressing these issues in the future


thezoggy:
1) no need to change line endings.. it will lose git history.. just tell your client to ignore them
2) new code goes into development not master first

kierse:
Removed line endings commit and opened new pull request against development [here](https://github.com/midgetspy/Sick-Beard/pull/661)

PR #663, opened 2013-04-17
Increased support for seeding torrents after rename.

nervling:
Instead of simply leaving an option to keep files when renaming, there are now four options.
Move - Do not keep just move.
Copy - Same as keep before, copy file.
Physical Link - Creates a physical link in the destination folder.
Symbolic Link - Moves the file and creates a link in its original location.

Combining post processing scripts with this will allow using to continue seeding torrents while still having the media files renamed/moved into their TV directories.

I've added linktastic to libs which is used by CouchPotato and nzbToMedia to create symbolic/hard links on both *Nix and Windows systems.

Also I realize that by using nzbToMedia we can create physical links and then just allow sickbeard to move the files. However, not all systems support physical links (unRaid to name one), so even if I can get nzbToMedia to include symbolic links, Sickbeard also will need to support symbolic links. Additionally, it seems better to me to just include this support in sickbeard itself rather than in the scripts.

### DIFF
--- a/data/interfaces/default/config_postProcessing.tmpl
+++ b/data/interfaces/default/config_postProcessing.tmpl
@@ -53,10 +53,25 @@
                         </div>
 
                         <div class="field-pair">
-                            <input type="checkbox" name="keep_processed_dir" id="keep_processed_dir" #if $sickbeard.KEEP_PROCESSED_DIR == True then "checked=\"checked\"" else ""# />
-                            <label class="clearfix" for="keep_processed_dir">
-                                <span class="component-title">Keep Original Files</span>
-                                <span class="component-desc">Keep original files after they've been processed?</span>
+                            <label class="nocheck clearfix" for="process_method">
+                                <span class="component-title">Process Episode Method:</span>
+                                <span class="component-desc">
+                                    <select name="process_method" id="process_method" class="input-medium" >
+                                        #set $process_method_text = {'copy': "Copy", 'move': "Move", 'hardlink': "Physical Link", 'symlink' : "Symbolic Link"}
+                                        #for $curAction in ('copy', 'move', 'hardlink', 'symlink'):
+                                          #if $sickbeard.PROCESS_METHOD == $curAction:
+                                            #set $process_method = "selected=\"selected\""
+                                          #else
+                                            #set $process_method = ""
+                                          #end if
+                                        <option value="$curAction" $process_method>$process_method_text[$curAction]</option>
+                                        #end for
+                                    </select>                                
+                                </span>
+                            </label>
+                            <label class="nocheck clearfix">
+                                <span class="component-title">&nbsp;</span>
+                                <span class="component-desc">What method should be used to put the renamed file in the TV directory?</span>
                             </label>
                         </div>
 
--- a/lib/linktastic/README.txt
+++ b/lib/linktastic/README.txt
@@ -1,0 +1,19 @@
+Linktastic
+
+Linktastic is an extension of the os.link and os.symlink functionality provided
+by the python language since version 2.  Python only supports file linking on
+*NIX-based systems, even though it is relatively simple to engineer a solution
+to utilize NTFS's built-in linking functionality.  Linktastic attempts to unify
+linking on the windows platform with linking on *NIX-based systems.
+
+Usage
+
+Linktastic is a single python module and can be imported as such.  Examples:
+
+# Hard linking src to dest
+import linktastic
+linktastic.link(src, dest)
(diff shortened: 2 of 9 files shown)


mr-orange:
if a release is archived (rarred/zipped) this mechanism is it working?

nervling:
How does sickbeard currently handle post processing zipped/rarred releases? I haven't been able to find anything for it in the source code during post processing, so if there is something please point me there.

The only potential issue is if the file is moved to the TV directory then unzipped. This would break a symbolic link (physical is fine). In this case, we'd need to just do a copy instead because we have to keep a copy of the zip file for seeding and an unzipped for the TV directory.

If sickbeard unzips before post processing that's fine.

mr-orange:
sickbeard does not unzip/unrar anything. For usenet user unrarring is done by sabnzb for torrent it can be done via client script (utorrent/transmission/deluge....). Anyway yuo have to check if the release is archived and do what you said.

nervling:
So you want it to just default to copy if it's a rar/zip? I assume sickbeard would actually post process a zip file successfully, meaning it would not complain that it's not a media file or something? Presumably, the file should be unzipped by the time it reaches sickbeard (by the userscript), but for the case that doesn't happen doing a copy seems reasonable to me.

mr-orange:
You are wrong. SB will not process Any archive. I agree with you that for archived release you have to move unarchived release + don't move archived content. This kind of mixed logic is present in your code?

nervling:
No my code does not currently do this, it just sym links all types of files during processing.

If sickbeard does not process any archive what's the point? This change only affects what sickbeard is post processing, so if it can't process an archive, that case would never come up no?

mr-orange:
Probably you are right but if a release it's archived you will find in the tv-download dir the archived release + unarchived one by the client script. In this case set a phisical/hard/link  or copy the release (unarchived) it's not the best choice.... the best would be move the release in the sickbeard directory overriding the post processing config

nervling:
I agree, that would be the best option, in that scenario. Do you really think this is worth coding into sickbeard though? As a client, if I am calling the API I would expect it to behave as expected instead of behaving differently when there is an archive of the same name in the processed directory. 

Also, the downside of creating the links when it is unnecessary is incredibly low. Hard links and symbolic links use almost no disk space, so the cost of creating these links is negligible. Copy would duplicate the file for no reason, but this is actually the current behavior of sickbeard. I did not add copy, it's just the previous Keep Files option.

I'm not opposed to doing the work if that's what the sickbeard devs think is best, but I personally wouldn't go that route.

mr-orange:
Yes probably code this feature is a bit overkill and as you said links
takes no time and space.
Good luck for this.. Surely it's a requested feature from a lot of torrent
guys.

nervling:
Thanks, hopefully @midgetspy can take a look at this soon :).

pisoj1:
This only works if they are on the same hard drive correct?

nervling:
Correct, though if the user did try and configure it that way incorrectly it will work by failing over to copy.

nervling:
link-ss

Here is a SS of what the feature looks like btw.

thezoggy:
would those physical / symlink options work right for os that cant do it? a lot of sb users run sb on a different os than where their content is..

nervling:
If their content is on a different machine it does not matter if the OS is different as a symlink/hardlink only works on the same disk (or raid array). 

Many of us run NAS machines capable of running sickbeard/couchpotato/sab/transmission etc., in which case this is helpful for us. People running separate machines/disks who want to continue seeding should use the copy option.

nervling:
Also, I was incorrect earlier, symbolic links will work between separate file systems, physical ones will not.

Ockingshay:
any traction on this? a much used feature of couchpotato and one greatly used by the torrent community.

crayment:
+1

nervling:
I'd suggest using the TPB branch if you really want this. I believe this is checked in to the dev branch there.

junalmeida:
@nervling At linux I can sucessfully make symlinks between different harddrives without raid.
I can do the same using different drive letters in Windows NTFS:

```
D:\>mklink test /D c:\Boot
Link simbólico criado para test <<===>> c:\Boot

D:\>dir test
 O volume na unidade D é Dados
 O Número de Série do Volume é D76C-EE20

 Pasta de D:\test

19/10/2013  10:52    <DIR>          bg-BG
19/10/2013  10:52    <DIR>          cs-CZ
```

junalmeida:
@VeNoMoUs I will give this a try.

PR #666, opened 2013-04-27
Better click-handling of main show listing's table

dbr:
> Mainly it handles cmd+click on Chrome/OS X properly - before, the link
> would be opened twice, once in a new tab (default link handling) and
> once in the current tab (because of the window.href change)
> 
> This should better emulate the clicking behaviour in "most" browsers,
> but I've only tested in Chrome 26.0 and Safari 6.0, on OS X. What
> could possibly go wrong?

### DIFF
--- a/data/js/tableClick.js
+++ b/data/js/tableClick.js
@@ -1,10 +1,23 @@
 $(document).ready(function () {
 
     $("table td.tvShow").live('click', function (e) {
-        if ((!$.browser.msie && e.button == 0) || ($.browser.msie && e.button == 1)) {
-            if (!e.shiftKey) {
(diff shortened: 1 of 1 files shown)


thezoggy:
all new code goes into development branch first. can you resubmit for that branch? (sadly it means submitting a new pull request -- github wont let you change the target branch after submitting)

dbr:
Ah, okay - new pull request at #685 

Might be worth having a `CONTRIBUTING` file with details like that? (Github [will point people towards](https://github.com/blog/1184-contributing-guidelines) it when making a pull request etc)

thezoggy:
Didn't know about this. Thanks will have to add :)

-----Original Message-----

PR #668, opened 2013-04-29
Fix hr.ws.pdtv.x264 parsing

georgewhewell:


Prinz23:
``` python
self.assertEqual(common.Quality.HDTV, common.Quality.nameQuality("Test.Show.S01E02.HR.WS.PDTV.x264-GROUP")) # why does this not pass?
```

To answer the original question: Because it's SD not HD... The resolution of such releases is only around 540 lines which is lower as even PAL with 576 lines. So it's clearly SD Quality.

So the part:

``` python
or checkName(["hr.ws.pdtv.x264"], any)
```

in HDTV section of the common.py is wrong. At least I don't want such SD releases as HD downloaded.

That is just my personal view of such releases.

thezoggy:
the reason it doesn't work is because we reversed the checking order, and the pdtv part results in it being matched as SD (since we stop at the last quality matched). left the bug in there since it really only affected one show. `Kitchen.Nightmares.US.S06E06.HR.WS.PDTV.X264-DIMENSION` and midget was suppose to look at making a proper fix for it. but if this fixes it, then might as well pull it

thezoggy:
@Prinz23 the hr.ws part shows that its hd not sd (theres not a true Hd for these type of shows since they way they are broadcasted.. so its not like HDTV).

HD release: Kitchen.Nightmares.US.S06E09.HR.WS.PDTV.X264-DIMENSION
SD release: Kitchen.Nightmares.US.S06E09.PDTV.x264-LOL

```
HR.PDTV.x264:                                                              │
│   - HR PDTV x264 encodes are allowed for non-HD shows airing fully          │
│     upscaled on HD channels. Windowboxed (with large black bars on all      │
│     sides) content lacks sufficient source resolution to warrant releasing  │
│     in HR format and is therefore banned.                                   │
│   - Video shall be cropped and resized to fit within 960x544 following      │
│     standard crop and resize rules listed above                             │
│   - Directory tags shall be HR.PDTV.x264 for 4:3 shows, and HR.WS.PDTV.x264 │
│     for 16:9 shows                                                          │
│   - All other standard rules apply          
```

and about source

```
ª   - Source Definitons:                                                      ª
ª     - 720p/1080i/1080p shall be tagged HDTV or AHDTV (depending on capture  ª
ª       method). In the case of upscaled content airing on HD channel, use    ª
ª       PDTV tag.                                                             ª
ª     - 576i/576p shall be tagged PDTV.                                       ª
ª     - 480i shall be tagged DSR.  
```

Prinz23:
First of all I don't care about scene rules. There are a few scene rules that are against any logic.

BTW. my postprocessing fix will refuse this as hd, because the resolution is SD and not HD. Such releases will fail the resolution check.

And even the scene rules state:

> HR PDTV x264 encodes are allowed for non-HD shows airing fully upscaled on HD channels.

It's non-HD, so it's SD just upscaled on the TV, but again downscaled to SD resolution by the release group. How is this HD?

Newznab sites also list hr.ws.pdtv.x264 releases correctly in TV-SD Category.

baz1860:
@Prinz23 Scene 'logic' while occasionally being slightly off-kilter (see the numbering of American Dad for details) is what you should be coding to - they release what you download. Code to your 'logic' and stuff is gonna fail
—

thezoggy:
yes nn sites classify anything with pdtv as SD, but as you know.. nn isnt always the rule to follow.
regardless. the 'hr' releases are going to be the bigger resolution between the two. these shows dont get released as typical hdtv or webdl.. so generally the hr.ws.pdtv is going to be the best aired you can do.. and theres only a handful of shows that get capped in it.. hence its a small bug we havent really cared about (no one has complained). and to recap, the hr.ws are released under mkv container.. like scene does for hd. and the non hr is mp4.. like scene's sd.

Kitchen.Nightmares.US.S06E14.HR.WS.PDTV.X264-2HD

```
Overall Container Format    Matroska
    Bitrate 7 138 bps
Video   Duration    43mn 2s

    Codec   x264
    Width   960
    Height  544
    Aspect  16:9
```

Kitchen.Nightmares.US.S06E14.Prohibition.Grille.PDTV.x264-2HD

```
Overall Container Format    MPEG-4
    Bitrate 4 758 bps
Video   Duration    43mn 2s

    Codec   h.264
    Width   720
    Height  404
    Aspect  16:9
```

Prinz23:
BTW. Nowhere in the Scene rules is it written that HR.PDTV.x264 is HDTV. That is just your interpretation of it.

The release (not source) video resolution decides what is SD and what is HD. If you would take the source resolution as the criteria you would have to say that practically all SD releases are HD, because they are almost always just downscaled HDTV recordings.

And you break nothing regardless how you categorize it: SD or HD... Just the User has set the correct one.

Putting something like that in HDTV would break any on known standard resolution based tests, like i want to add to the postprocessing. HD is defined as much higher resolution as SD (SD includes all up to 576 lines resolutions).

thezoggy:
source != quality. and you seem to be missing the point here. you say they are both SD. well how would a user pick between the two releases if they are both SD and both come out. fact of the matter is that both come out from the scene and one is higher quality than the other. I hope you would agree that one is better than the other.. one is hd while the other is sd (with regards to sb) and this patch is with regards on having the 'hd' version actually match as that rather than sd.

PR #669, opened 2013-04-30
Added proper-finding to BTN

genial123:
Implemented the previously non-existing findPropers() to BTN provider. Cannot use OR-operator in the API's release-field, so two API-calls are needed.

### DIFF
--- a/sickbeard/providers/btn.py
+++ b/sickbeard/providers/btn.py
@@ -23,12 +23,13 @@ import generic
 from sickbeard import scene_exceptions
 from sickbeard import logger
 from sickbeard import tvcache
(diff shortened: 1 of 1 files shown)


mr-orange:
The findPropers() is implemented in generic provider. Why are you redefing this?

genial123:
From what I could see that was how it was done in the other providers, so thought that was the preferable way of doing it. If not, where should I place it?

thezoggy:
you subclass the generic provided functions if the generic functions dont do what you want

thezoggy:
This can be closed, this was handled under Pull #713

PR #670, opened 2013-05-05
omgwtfnzbs tweaks

thezoggy:


thezoggy:
This turned out to be slightly more than omgwtfnzb tweaks as we now do auth check before trying to use the provider. I've tested this with multiple providers and everything appears to be in order and safe to pull.

thezoggy:
need to update this to migrate old OMG config values... since now its been pushed out into a build :(

PR #671, opened 2013-05-08
Update postProcessor.py

bobbintb:
The postProcessor.py file has been modified to now include the name of any episode that has been replaced with a better quality one as an extra parameter passed to the "extra scripts". This will allow the user more flexibility in their extra scripts ran after post processing by knowing what files have been replaced. Normally the following 6 parameters are passed:
1. final full path to the episode file
2. original name of the episode file
3. show tvdb id
4. season number
5. episode number
6. episode air date

The 7th is now the deleted file. If no file was replaced, the argument will be "None"

This is my first experience with git and a pull request so if I did something wrong, please be gentle. :)

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -202,6 +202,7 @@ class PostProcessor(object):
                 ek.ek(os.remove, cur_file)
                 # do the library update for synoindex
                 notifiers.synoindex_notifier.deleteFile(cur_file)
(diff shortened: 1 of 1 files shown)


Conjuro:
Pull requests are issued against the development branch and not master. Only way I know of to fix it, is to close this PR and reissue against right branch.

bobbintb:
oops, my bad. i knew that too. just wasnt paying attention.

PR #674, opened 2013-05-11
Enhancement :: Provider - Newznab : Increased search results

Conjuro:
Newznab providers only return the first 100 matches to searches. Added offset processing in order to retrieve "all" (offset hardcoded to max 1000) matches.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -176,6 +176,7 @@ class NewznabProvider(generic.NZBProvider):
         params = {"t": "tvsearch",
                   "maxage": sickbeard.USENET_RETENTION,
                   "limit": 100,
+                  "offset": 0,
                   "cat": '5030,5040'}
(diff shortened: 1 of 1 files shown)


Conjuro:
Rebased for current development branch. Been using for one month without issues.

Any comments on the sanity check? Smaller? Larger? Remove? I included it because initially I made a mistake with types and the check to break the loop never occurred causing a lot of grabs on the search provider, so I got a bit gun shy.

thezoggy:
ill have to check, this change fundamentally changes the need for the hack i did awhile ago.. and if we really do want to go this route.. it means more hits to the providers.. and we could do other categories as well.. need to review

koendelaat:
I requested another implementation of this problem under https://github.com/midgetspy/Sick-Beard/pull/838
In this solution it checks for the number of remaining items, so no risk of an infinite loop

PR #675, opened 2013-05-13
Twitter direct message

sirchia:
The changes in the pull request allow other users to be mentioned in the notification tweets, as well as open up the possibility to use twitter direct messages instead of publicly visibile tweets on the timeline.


Conjuro:
Pull requests should be issued against the development branch.

PR #676, opened 2013-05-13
Added support for direct messages and mentions in twitter notifications

sirchia:
Oops, wrong target branch...did base my changes off the correct branch though :) 

_Take 2_

The changes in the pull request allow other users to be mentioned in the notification tweets, as well as open up the possibility to use twitter direct messages instead of publicly visibile tweets on the timeline.


drizzt09:
I am using SB 502 on Windows 8, how do i get these updates?

PR #681, opened 2013-05-22
Bugfix :: loadFromTVDB : Prevents corrupted data from TheTVDB

Conjuro:
tv_shows.show_name in sickbeard.db can never be a NoneType. Patch prevents this from happening if someone has removed the seriesname attribute in the TVDB show entry.

http://www.sickbeard.com/forums/viewtopic.php?f=4&t=7249

### DIFF
--- a/sickbeard/show_queue.py
+++ b/sickbeard/show_queue.py
@@ -404,6 +404,9 @@ class QueueItemUpdate(ShowQueueItem):
         except tvdb_exceptions.tvdb_error, e:
             logger.log(u"Unable to contact TVDB, aborting: " + ex(e), logger.WARNING)
             return
+        except tvdb_exceptions.tvdb_attributenotfound, e:
+            logger.log(u"Data retrieved from TVDB was incomplete, aborting: " + ex(e), logger.ERROR)
+            return
 
         # get episode list from DB
         logger.log(u"Loading all episodes from the database", logger.DEBUG)
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -637,7 +637,10 @@ class TVShow(object):
 
         myEp = t[self.tvdbid]
 
-        self.name = myEp["seriesname"]
+        try:
+            self.name = myEp["seriesname"].strip()
+        except AttributeError:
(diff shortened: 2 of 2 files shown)


Conjuro:
Any suggestions for a good way to test it? I suppose I could stick a proxy between me and TVDB to alter the incoming data, just hoping for an easier/quicker solution.

Conjuro:
https://github.com/Conjuro/Sick-Beard/commit/a6775ad2ddc409b10c1e3f0ad94e12ced60f17e7#commitcomment-3272009

Thanks, @thezoggy. That does feel like a neater way of doing it. I also replaced the `else` with `except` since `else` can only follow an `except` and has the opposite effect.

Hmm. Squash and forced push removed the commit comment reference, any way of replacing commits but retaining the comment history?

thezoggy:
do you have a show/tvdbid that you use to test this code out with? / how did you find this issue?

Conjuro:
Came across it while doing support in the [forums](http://www.sickbeard.com/forums/viewtopic.php?f=4&t=7249). First one person, then another two reported the same issue. One using windows binary, one using git master also windows, and third using a different fork. Had them upload their databases and found it to be the same show (Person of Interest, tvdbid 248742) for all three. I checked the TVDB entry, but it's correct now and never went wrong in my own database, indicating it was only missing for a period less than 24 hours. I didn't question further, but I'd guess all three are in the same or at least neighboring time zones.

So far I've tested that the patch doesn't interfere with normal data, and using a custom python script that it behaves as expected. Unless someone can suggest a better way of testing, I'll try using a proxy to alter the data in real time while doing a "Force Full Update", when I get a chance.

Conjuro:
I've now tested it by using a [proxy to alter](http://conjuro.net/misc/user.filter) the [SeriesName tag](http://conjuro.net/misc/poi.en.xml) in the [xml from thetvdb.com](http://thetvdb.com/api/9DAF49C96CBF8DAC/series/248742/en.xml) while using "Force Full Update". Sick Beard aborted the update and returned the [proper error](http://conjuro.net/misc/sickbeard-seriesname.log).

VeNoMouS:
Nice conjuro :), re timezones, knightNZ (least i assume from his username) and myself are based in New Zealand

PR #683, opened 2013-05-31
Update newznab.py

ghost:
Added 5060 category (other) for NZBs(dot)ORG. This will make the RawHD TV quality work with NZBs(dot)ORG.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -184,7 +184,7 @@ class NewznabProvider(generic.NZBProvider):
 
         # hack this in for now
         if self.getID() == 'nzbs_org':
-            params['cat'] += ',5070,5090'
+            params['cat'] += ',5060,5070,5090'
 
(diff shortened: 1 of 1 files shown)


thezoggy:
new code goes into development not master.

many providers wont work well with rawhd due to how they classify it.. for example in sbi we throw it in tv>other just like most sites...  why when i added support i even made note of the providers that would and wouldnt work.. at that time nzbmatrix/newzbin needed changes for example.. but they are gone now. the problem with adding the 'other' category is that you introduce a bunch of crude into the results.. and then also with sites only showing back a limited amount of results.. you run the risk of wasting those precious spots for crap that doesnt matter. which means you could be missing valid results because they are pushed outside of the result amount. why we didnt want to add the cat.. my solution is just to allow people to configure the cats per nn indexers if they want.. but it requires some changes and not sure if we really want this to be a gui change since people could do much more harm than good.

PR #685, opened 2013-06-05
Better click-handling of main show listing's table (corrected branch)

dbr:
Exact same as #666 but in proper branch

> Mainly it handles cmd+click on Chrome/OS X properly - before, the link
> would be opened twice, once in a new tab (default link handling) and
> once in the current tab (because of the window.href change)
> 
> This should better emulate the clicking behaviour in "most" browsers,
> but I've only tested in Chrome 26.0 and Safari 6.0, on OS X. What
> could possibly go wrong?

### DIFF
--- a/data/js/tableClick.js
+++ b/data/js/tableClick.js
@@ -1,10 +1,23 @@
 $(document).ready(function () {
 
     $("table td.tvShow").live('click', function (e) {
-        if ((!$.browser.msie && e.button == 0) || ($.browser.msie && e.button == 1)) {
-            if (!e.shiftKey) {
-                var href = $(this).find("a").attr("href");
-                if (href) { window.location = href; }
+        // Check the primary click button is pressed, different in IE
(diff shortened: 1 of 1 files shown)


thezoggy:
Trying this out, FF 23.0.1 on OS X 10.6
with build500, cmd+click this results in the current tab going to the link and then a new tab to the page as well.
with this pull, cmd+click results in two tabs going to the link while the original click being prevented.

Trying this out Chrome 31.0.1650.26 beta on OS X 10.6
with build500, cmd+click this results in the current tab going to the link and then a new tab to the page as well.
with this pull, cmd+click results in one tab going to the link while the original click being prevented (just as stated).

so just need to tweak this to make firefox happy as well...

thezoggy:
just submitted new pr to resolve this. JS is not the way to go as you battle browser differences and os differences. its best to just leave this up to the browser to handle  and  correct the target area by using CSS.

PR #697, opened 2013-06-25
Fix show.stats -- exclude specials

thezoggy:
show.stats :: Exclude specials (season 0) from stats (match what we do for sb ui)


styxit:
This is what is was looking for, thank you.

PR #701, opened 2013-06-30
Update processTV.py

immure:
Currently if asked to process a directory without a video file then the script will silently fail


thezoggy:
new code goes into the development branch. please resubmit

PR #702, opened 2013-06-30
Update processTV.py

immure:
Currently if asked to process a directory without a video file then the script will silently fail


ghost:
When using Post Processing option "Scan and Process"  you log an error every 10 minutes.

PR #704, opened 2013-07-01
Added <br /> to manage_episodeStatuses.tmpl

ghost:


junalmeida:
According to W3C:

```
br elements must be used only for line breaks that are actually part of the content, as in poems or addresses.
br elements must not be used for separating thematic groups in a paragraph.
If a paragraph consists of nothing but a single br element, it represents a placeholder blank line (e.g. as in a template). Such blank lines must not be used for presentation purposes.
```

http://www.w3.org/TR/html5/text-level-semantics.html#the-br-element

thezoggy:
there are many things that you should or shouldnt do with html with regards to what the w3c says. either we add `<br>` or wrap the line in `<p>` or `<div>` or modify the previous table css and then go add in other conditional css. regardless this is the lesser of all evils and we've done it before.

PR #707, opened 2013-07-04
Issue 1027: Fanart.tv Support

dan-boehm:
This Pull request is for a solution to [Issue 1027](https://code.google.com/p/sickbeard/issues/detail?id=1027&q=fanart&colspec=ID%20Type%20Status%20Priority%20Milestone%20Owner%20Summary%20Stars)

These commits enable support for downloading/saving extra artwork from [Fanart.tv](http://fanart.tv/tv-fanart/).  This is in addition to the artwork already downloaded from the TVDB.

Newly downloaded artwork includes:
- Landscape Art (Thumb)
- ClearLogo
- ClearArt
- Character Art
- Extra Fanart (Backgrounds)

Additionally, Banner art can now be downloaded along with the typical poster art (as opposed to just in lieu of).  This is still downloaded from the TVDB.  Also, the use_banner checkbox still works as it did previously, so that support is still covered.

This solution can be slow if the extra fanart option is checked.  This is due to some choices that were made in the existing codebase, where the image data is downloaded before it is determined if it should even be saved.  This can be worked on in the future.  However, not using the new functionality shouldn't result in any noticeable performance hit.

Several choices were made for interfacing with Fanart.tv.

First, a new library to aid in interfacing with their API was added.  It is licensed under the Apache License, Version 2.0.  The original source repository is here: https://github.com/z4r/python-fanart.

Finally, Fanart.tv requires a free API key.  Instead of registering a key specifically for Sick-Beard, I included a config option to enter a key.  It can be entered from config->post-processing.  It would be possible to use a Sick-Beard specific key, but I like the idea of forcing the user to register a key on Fanart.tv since it supports the service a little better.  It is still young, and it could benefit from a larger more-involved community.

The following are some examples of how these extra images are used in XBMC:

Aeon Nox Logo view

Aeon Nox Paused with ClearArt


thezoggy:
new code goes into the development branch first. sickbeard still maintains python 2.5 compatibility.. adding a lib that requires 2.6+ isn't going to go over well with midgetspy..

dan-boehm:
Fair enough, I'll close this pull request and submit a new one when I address that.

PR #710, opened 2013-07-27
CMD_Shows should return both TVDBID and SHOW_NAME regardless of sort order

seanstar12:
When making an API call, CMD_Shows returns an object of objects thus making grabbing required information difficult to obtain. These modifications would make of the information available.


thezoggy:
new code needs to go against the development branch not master. not sure why this is needed though...

PR #711, opened 2013-07-27
Improve SDTV and SDDVD quality detection by matching source and encoding separately.

mrmachine:
This will allow filenames like `Doctor.Who.S02E01.New.Earth.DVDRip.DD2.0.x264 Z.mkv`, which include audio information between the source and encoding, to be matched correctly.


midgetspy:
As this is one of the few areas of SB that actually has tests can you please write tests for this code? And also confirm that you ran the existing QualityTests in common_tests.py? Thanks.

PR #715, opened 2013-08-04
New API call sb.analyzename

jayme-github:
I needed a way to figure tvdbid, season and episode(s) out of a name string.

As this is already done while postprocessing, this code is mostly cloned. There might be a better way to do this without code duplication. Maybe the process should be moved to a more general place and then could be used by `Postprocessor._analyze_name()` and the api?

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -30,7 +30,7 @@ import traceback
 import cherrypy
 import sickbeard
 import webserve
-from sickbeard import db, logger, exceptions, history, ui, helpers
+from sickbeard import db, logger, exceptions, history, ui, helpers, show_name_helpers, scene_exceptions
 from sickbeard.exceptions import ex
 from sickbeard import encodingKludge as ek
 from sickbeard import search_queue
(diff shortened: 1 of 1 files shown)


thezoggy:
sounds like you are trying to implement post processing code via the api.. 1) you would just call the actual post processor.. 2) not sure why you'd want to do this unless your just trying to use sb internals for your own app's gain.
i see no reason to pull this into sb trunk at this time.

jayme-github:
The reason for this is simple: I wanted to determine if I'm interested in some files/folders without having direct access to them.
To query the api I needed the tvdb id, series and episode number. The easiest way to get these was the reliable code already present in sickbeaed - so why don't expose this function via api?

If you're not interested in this please close this request. If the problem is the implementation I'm willing to change the code and call parts of the postprocessor directly if possible.

thezoggy:
I'll create a branch to get this documented on the api.. added to apibuilder.. and review

thezoggy:
This command doesn't handle air-by-date filenames.. `Jimmy.Fallon.2012.06.07.Chris.Rock.REPACK.720p.HDTV.x264-2HD` for example results in traceback.

ghost:
Changed _analyze_name() to _find_info() to get season episodes

PR #718, opened 2013-08-13
Add option to restrict the desired encoding when searching for episodes.

sammys:
I'm creating a pull to notify everyone that this is in progress. I welcome people to test this out (backup your database first) and also to suggest any improvements that can be made to this particular feature.

Due to the size of the patch I will limit the scope to restricting searches **only**. If you have other encodings that need to be supported along with file extensions then please notify me.

I have not tested mass updates nor saving the encoding when episodes are downloaded or postprocessed. That needs a bit more testing than I have time for right now. Please assist if you can.


thezoggy:
'encoding' is the wrong term for this. 'codec' is the correct one. and 'xvid' is no more as of the end of 2012. so i see little reason to add this.. it would only cause issues for people trying to backfill. there was only a brief period where you saw both codecs being used..

sammys:
Hi @thezoggy. Thanks for taking a look at this pull request. As you said, codecs only causes issues for people trying to backfill. Use case:

> Person 'A' solely uses iTunes for one's TV show playback so, for all shows, 'A' wants to automatically download x264 releases only. Once they are downloaded 'A' will switch the restriction to Any and have their postprocessor transcode them.

It's particularly handy for shows like Saturday Night Live, which have many seasons. Let's say that the person would like to only download x264 releases and ignore XviD. Another use case:

> A show has already been downloaded in XviD and transcoded. The codec can then be restricted to x264 and switched to Wanted. Any subsequent blu-ray releases will then be downloaded and overwritten.

Not sure if the overwriting part is implemented yet, hehe.

It may not be useful to you but there are people with these unique use cases. The code is not very complex and it won't add much maintenance overhead.

To meet your requirements I have modified all code to use the term codec instead of encoding.

Thanks again for considering my contribution to Sick Beard. Nice work so far btw.

PR #719, opened 2013-08-17
Separate WEBRip from WEB-DL

ghost:
WEBRip and WEB-DL should not be considered the same thing because WEBRip quality varies and isn't consistent.

Currently i have WEBRip in my ignore_words list as advised by the change log, but their are certain shows that are released as WEBRip only or long before WEB-DL so putting WEBRip in ignore words isn't an optimal solution.

I and probably others would prefer if we were able to initial as WEBRip and archive as WEB-DL therefore we need them to be listed under separate settings.

Really appreciate your hard work. I don't like to bribe but i pledge to donate $20US if this issue gets resolved.

Thank-you.


thezoggy:
this pull was submitted for the wrong branch, unable to tell what you are changing.. please resubmit. but honestly webrip+webdl being in the same class is fine.. you usually dont see both come out.. just one or the other. as webrip is the screen capture (hulu/netflix) and webdl is (itunes). so splitting these into their own quality just adds unneeded complexity to the already growing number of qualities

thezoggy:
looks like this was just a feature request submitted as a pull request. so please close this out.

PR #721, opened 2013-08-21
fixed key error in case the xbmc version returns a result that is missing keys

Velrok:


thezoggy:
i do believe this was already fixed in the development branch a few months ago...

Velrok:
Ok. Never mind then :).
This was a hotfix for me and I thought I just push it back.

ghost:
The other fix was for different missing keys.
On which XBMC version does this error happen?

PR #723, opened 2013-09-07
Option to suppress XBMC 'Failed to detect XBMC version' warning messages in log

amney:
Many XBMC users configure their devices to sleep/shutdown when not in use. This causes sickbeard to throw warning messages whenever an XBMC update/notification takes place and the XBMC is not reachable. This can quickly fill up the web interfaces log warning page, potentially concealing more important messages and causing the user to repeatedly have to clear their error log.

### DIFF
--- a/sickbeard/notifiers/xbmc.py
+++ b/sickbeard/notifiers/xbmc.py
@@ -144,7 +144,7 @@ class XBMCNotifier:
                     if notifyResult:
                         result += curHost + ':' + notifyResult["result"].decode(sickbeard.SYS_ENCODING)
             else:
-                logger.log(u"Failed to detect XBMC version for '" + curHost + "', check configuration and try again.", logger.ERROR)
+                logger.log(u"Failed to detect XBMC version for '" + curHost + "', check configuration and try again.", logger.MESSAGE)
                 result += curHost + ':False'
 
         return result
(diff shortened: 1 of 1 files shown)


amney:
That's a reasonable course of action, but would it not upset users that would like to know if their XBMC box is down? If you are ok with that I will add some more commits.

amney:
@thezoggy - I have gone ahead and reverted the previous commit and added a new commit that only changes the log level, without presenting any new user options.

ghost:
For a cleaner commit history:

git rebase -i HEAD~3

only pick last line, force push

amney:
Thanks for the tip, I'm relatively new to merging and was concerned rebasing would 'rewrite history'.

ghost:
Yes, you changed history :)

On shared branches you shouldn't do this.

piejanssens:
These messages are really annoying and not an error at all. Please merge asap.

thezoggy:
this is being addressed in another branch

thezoggy:
this can be closed, ended up just implementing your original plan to have an option to suppress the error. sorry it took so long 
https://github.com/midgetspy/Sick-Beard/commit/2732fb03e36abdcfdcf098dfce475c4b6f55a2d0

PR #724, opened 2013-09-09
Fix, add, change hdbits provider

ghost:
Fix json import and init HDBITS_USERNAME
Add credentials check
Add extra parameter in getUrl for POST data.
Change url variables
- pep8 + minor config provider page changes

Removed findEpisode, should use _get_episode_search_strings and _doSearch 
but don't have a hdbits account to make this change and test.


thezoggy:
why the close?

ghost:
To prevent accidental pull.
I think I make another PR with just the really needed fixes and reopen this
one when I have a change to test it.

PR #726, opened 2013-09-09
SB-API - Update

thezoggy:
Added JSONP callback to the api-builder.
Fix minor spelling mistake in the pp logger.
Remove sb.analyzename for now as it wasn't properly added, and after 'fixing' it (thanks itofzo) there appears to be some possible issues (messes with history) that could come up from misusing the pp for this purpose. Ideally this would need to be properly segmented and handled for the 'test' case scenario.. also it should add the quality for the release?


thezoggy:
Will let itofzo submit his pr to revert pull and ill re-submit for the minor changes so we keep history clean.

PR #736, opened 2013-09-24
404 fix when a custom web_root is used

jerbob92:
When SB reached an 404 page, and a custom web_root is used, it will always redirect to /, this commit will fix that and use the root from the options.


thezoggy:
needs to be against dev not master

PR #741, opened 2013-09-27
Fix trailing os dir separator

breunigs:
users are able to include a trailing slash/backslash. This causes `displayShow.tmpl` to eat the first character of the filename of a show. This patch ensures the os dir separator is removed if present.

STR the issue:
1. add an existing show. Note trailing os sep **isn't** present. Say path is `/some/show/test`.
2. Edit show
3. manually edit the location text field to read `/some/show/` (note trailing os sep **is** present)
4. Save

Path now contains a trailing os sep.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -2424,7 +2424,7 @@ class Home:
                     # change it
                     try:
                         if location.endswith(os.sep):
-                            location = location[:len(location) - 1]
+                            location = location[:-1]
                         showObj.location = location
                         try:
                             sickbeard.showQueueScheduler.action.refreshShow(showObj) #@UndefinedVariable
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -2423,6 +2423,8 @@ class Home:
                 elif not do_update:
                     # change it
                     try:
+                        if location.endswith(os.sep):
+                            location = location[:len(location) - 1]
                         showObj.location = location
                         try:
                             sickbeard.showQueueScheduler.action.refreshShow(showObj) #@UndefinedVariable


sickbeard/webserve.py
```diff
@@ -2423,6 +2423,8 @@ def editShow(self, show=None, location=None, anyQualities=[], bestQualities=[],
                 elif not do_update:
                     # change it
                     try:
+                        if location.endswith(os.sep):
+                            location = location[:len(location) - 1]
```
amney: Small change, this line could be written more efficiently as:
`location = location[:-1]`

breunigs: thanks, I’ve updated the PR accordingly

sickbeard/webserve.py
```diff
@@ -2423,6 +2423,8 @@ def editShow(self, show=None, location=None, anyQualities=[], bestQualities=[],
                 elif not do_update:
                     # change it
                     try:
+                        if location.endswith(os.sep):
+                            location = location[:-1]
                         showObj.location = location
```
ghost, line 2428:
```
    showObj.location = os.path.normpath(location)
```

should be enough.

breunigs, line 2428: `normpath` does more than simply removing the trailing slash. If I understood the docs correctly, it changes `/A/B/../C/` to `/A/C` by looking at the string only. If `B` is a symlink, this may give the wrong path. `realpath` appears to resolve symlinks correctly, but that removes any symlinks I might have (which makes it impossible to mirror the directories locally but store the files on external disks, for example).

I’m aware that the code above uses `normpath`-magic. I have no idea why `os.path.samefile` is only available on unix, otherwise I’d suggest switching to it.

PR #745, opened 2013-10-03
add release group important support.

rainlake:


thezoggy:
'best_releaser' is not a very good variable name. may want to rename that to something like 'preferred_releasegroups' so it indicates what it is.. and the fact it can hold multiple values.

PR #749, opened 2013-10-07
Remove duplicate scene exceptions (same name, different tvdb_id)

arucard21:
If duplicate scene exceptions exist, the first one found (= oldest) will be used. This removes the older duplicates and allows the latest retrieved scene exception to be used.

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -109,6 +109,10 @@ def retrieve_exceptions():
                 if cur_exception not in existing_exceptions:
                     myDB.action("INSERT INTO scene_exceptions (tvdb_id, show_name) VALUES (?,?)", [cur_tvdb_id, cur_exception])
                     changed_exceptions = True
+                # check if there are entries with the same show name but a different tvdb_id
+                # remove all duplicate entries, except the one with the tvdb_id you retrieved from the net
+                if myDB.select("SELECT COUNT(tvdb_id) FROM scene_exceptions WHERE show_name = ?", [cur_exception]) > 1:
+                    myDB.action("DELETE FROM scene_exceptions WHERE show_name = ? AND tvdb_id != ?", [cur_exception, cur_tvdb_id])
 
         # since this could invalidate the results of the cache we clear it out after updating
         if changed_exceptions:


thezoggy:
maybe its just your wording.. but `If duplicates exist, the first one found (= oldest) will be used.` makes it sound like that the newest exception (from online) is being discarded in pref over the old data in the db (which could be from the user or old online exception data). this would not be ideal.. in the situation where if we pushed out something wrong then fixed it.. users would be stuck with the old data.

arucard21:
Ah yeah, it's my wording. I was describing the old situation which was undesirable for the very reasons you mentioned. This commit fixes that old behaviour by removing the old, invalid data. 

So just to be clear: 
- Before this PR is applied: scene exceptions with the same name but a different tvdb_id get added to the database, resulting in 2 scene exceptions with the same name. When searching for scene exceptions, these are searched by name and the first, older, exception will be returned. This causes, as you described, the users to be stuck with their old exceptions as newer ones are added to the database but never actually found
- After this PR is applied: when adding scene exceptions, you check for those with the same name but a different tvdb_id (different from the one you are adding at that moment). If any are found, you remove all the older ones, leaving only the newer scene exception. This ensures that seaching for a scene exception by name will always return a single tvdb_id.

thezoggy:
took a look at this tonight, doesnt appear to do anything that i can tell

arucard21:
Yeah, it applies to a corner-case situation. It basically only occurs when you have an older scene exception for a show with a specific tvdb_id, but then you have another, newer scene exception for the same show but with a different tvdb_id. If this occurs, you want the more recent scene exception to be used since this is most likely the correct one (and if not, you can create another one to correct it). This patch will clean up the older scene exception and allow the new exception to be used. 

This happened to me with Hulk and the Agents of S.M.A.S.H.
Official tvdb site: http://www.thetvdb.com/?tab=series&id=267543&lid=7
Duplicate site: http://www.thetvdb.com/?tab=series&id=272497&lid=7 (almost entirely removed but still accessible by link as proof that this does occur)

So there was a scene exception for this show with tvdb_id 272497 which was always used instead of the newer one with tvdb_id 267543. Granted, it doesn't occur very often, but when it does it's not easy to figure out. 

It is easily reproduced by changing the tvdb_id of one of the scene_exceptions in the text file from which the scene exceptions are updated (in a test environment only, obviously). Of course, in that case you'll most likely be changing the tvdb_id to an incorrect one, but the older one will be removed anyway. You can correct it by changing it back to its original tvdb_id in the text file and updating again.

arucard21:
Closing this since my branch for this doesn't exist anymore (and the functionality is now already available in sickbeard, though implemented a bit differently)

PR #753, opened 2013-10-13
SB database cleanup

thezoggy:
- PEP8 fixes for db.py
- Cleaned up the db version check wording, raise SystemExit exception with message to match how we handle other startup issues.
- Added db version min check since we no longer are going to support migrating 2+ year old dbs.
- Updated initial db schema to latest (based off build 501).
- Trimmed down the SB db migration process to only support build 496+ (2012-06-28)
  
  > which covers the major migrations but allows us to eliminate a lot of legacy things that were not done correctly.

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -263,13 +263,15 @@ def main():
 
     sickbeard.CFG = ConfigObj(sickbeard.CONFIG_FILE)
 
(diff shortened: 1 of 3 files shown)


ghost:
Merged differently

PR #754, opened 2013-10-15
Add 10bit support

manusfreedom:
Be able to download 10bit video version.
Useful for anime.
I will try to add anime support without big change so it will can be integrated in main branch.


lad1337:
i am exited ! about the anime part ... not 10bit ... i dont like 10bit :P

PR #757, opened 2013-10-17
Speed up scrolling

timdorr:
I removed the fixed background, as it was causing a full page reflow under Chrome. In particular, this speeds up the scrolling on show pages with lots of episodes. I'm getting repaints of 0.3-0.5 ms instead of 200-300ms, so it's smooth as butter now :+1: 

Was there a reason the background was fixed? It's a very subtle pattern, so I didn't see any reason it should be.

### DIFF
--- a/data/interfaces/default/inc_top.tmpl
+++ b/data/interfaces/default/inc_top.tmpl
@@ -24,7 +24,7 @@
 
 <style type="text/css">
 <!--
-#contentWrapper { background: url("$sbRoot/images/bg.png") repeat fixed 0 0 transparent; }
+#contentWrapper { background: url("$sbRoot/images/bg.png") repeat 0 0 transparent; }
 
 [class^="icon-"], [class*=" icon-"] { background-image: url("$sbRoot/images/glyphicons-halflings.png"); }
 .icon-white { background-image: url("$sbRoot/images/glyphicons-halflings-white.png"); }


thezoggy:
what version of chrome? chrome v30 works fine as it is for me..  ill have to check this again at home on my sickbeard db with 2,000+ shows to really stress

timdorr:
I'm mostly noticing this on my Mac, which has a slower graphics card (Intel HD 3000) and is retina, so there are more pixels to push. My PC doesn't have it as bad, though you can notice the difference if you have two pages open side-by-side.

Here is one scroll from the SNL show page on my PC (Chrome 30.0.1599.101 on Win 8.1) with the fixed bg:
capture

And without the fixed bg:
capture2

Again, it's not a huge difference on my PC (6-5ms paints vs <1ms), but that also has much beefier hardware (Core i5-3570 and a GTX 680). On my laptop, the difference is more pronounced:

sick_beard_-_alpha_build_501_-_saturday_night_live_and_all_accounts__inbox__3_messages_

And without the fixed bg:
sick_beard_-_alpha_build_501_-_saturday_night_live_and_all_accounts__inbox__3_messages_

It appears to not even paint on my laptop because it can make use of the graphics card to directly composite the page to the display. 

Again, was there a reason for adding the fixed background? It seems like a relatively small design decision. I get that you're going for a look and respect that, but it seems like giving way a little bit on something I'm sure most people aren't going to notice would yield some pretty significant performance gains. Thanks for considering this!

thezoggy:
looking into this now, looks like its been a known 'bug' on chrome for awhile but its hit and miss on which os it shows up on.. http://code.google.com/p/chromium/issues/detail?id=18537
some notes here about it as well:
http://updates.html5rocks.com/2013/02/Profiling-Long-Paint-Times-with-DevTools-Continuous-Painting-Mode

guess i've never known/noticed this because all my machines are beefy. 
just tried it on a mac mini from a year ago.. no issues with the way it is on chrome/firefox.. so i cant really replicate this but i can atleast confirm that the change doesnt affect anything negatively (not sure about IE since I cant test that currently)

trying to recall why i did fixed instead of scroll (the default). could have sworn it was something to fix some issue in a browser.. but it could have been for the old bg image.. dunno but ill chime back on this in a few

thezoggy:
Fix the page bg from having funky issues on scrolling.  -- https://github.com/midgetspy/Sick-Beard/pull/529 / https://github.com/midgetspy/Sick-Beard/commit/7dcb5815b02f5717fb906ae18fd63587b926edce

now ill have to go dig on what browsers/what the issue was

update, the reason was:

```
2012-11-13 01:11:58 <jjaomni> do you get some black blocks in the really long web pages zoggy?
2012-11-13 01:11:54 <jjaomni> back groud color some times is black when i have reallly long pages
2012-11-13 01:11:39 <_zoggy_> jjaomni, that was a css issue i fixed awhile ago.. dont think midget's merged it
2012-11-13 01:11:50 <_zoggy_> the bg image for the div was set to scroll not fixed
```

so ill have to see if i can try and replicate that problem.. the change was made in build 499.. so ill just go load up build 498 real quick in a vm and see if i can cause it to happen / narrow down what is the true problem.

thezoggy:
just tested build 497 with IE8, FF 3.6.28, Chrome 5, Chrome 30, FF 24, IE11... no problems. so no idea why this was changed unless it was just something funky that I cant reproduce anymore. Safe to pull.

timdorr:
Thanks, man! I really appreciate you taking the time to look into this.

PR #759, opened 2013-10-19
Add rTorrent support

ccormier:
This commit adds full support for rTorrent as a target for torrents as nzbget and sab does for nzbs.

### DIFF
--- a/data/interfaces/default/config_search.tmpl
+++ b/data/interfaces/default/config_search.tmpl
@@ -100,7 +100,7 @@
                             </label>
                         </div>
 
-                        <div id="blackhole_settings">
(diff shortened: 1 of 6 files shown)


thezoggy:
the blackhole method isnt nzb specific...  know people use it with torrents/scripts.

ccormier:
Yes I tried torrent blackhole, it does work too, this it's an improvement to the torrents blackhole. Works similar to how "sab" and "nzbget" enhance the nzb blackhole.

-benefits are instant add to queue
-instant status on queue add in sickbeard
-no need to setup blackhole/watch directories
-sickbeard and rtorrent on different computers
-implemented same way as NZBget and SAB support

PR #760, opened 2013-10-19
Update ezrss.py

AndyMeusel:
Causing SSL error


thezoggy:
so you dropped the s and changed several lines line endings?

PR #765, opened 2013-10-22
Fix for log file handle remaining open on restarts

nate1280:
Close the log file prior to starting a new subprocess so the log file doesn't remain locked by previous instance.

I've had this issue for a long time in both source and exe builds with windows.  I'm not sure if other OS's have the same problem or not.

To reproduce, just start SB, and restart it from within the UI.  Using process explorer goto the python process for sickbeard and press Ctrl+L, you should see the log file listed twice, if its like this, the log file is locked still by the previous instance.


ghost:
Thanks. Rewrote it a bit and use the close function inside SBRotatingLogHandler too.

PR #767, opened 2013-10-29
Add grace-period to bypass the local tvdb show status 'ended' check.

thezoggy:


ppslim:
Why is graceperiod hard coded and not a hidden config item?

thezoggy:
because we dont want people to abuse things, and not everything needs to be a config item

ppslim:
Agree, but leads on to the inevitable, what is the thinking behind the values chosen?

thezoggy:
if you just analyzed how it all works you would see that you really dont need to tweak anything, we are going to hit tvdb more often for ended shows than ever before.. but we just dont want to hit them all the time. this is a pretty good balance of checking in for a period of time before/after the last ep airdate and then before the next ep air and then finally every 3 weeks. if you really want to debate this, come join the sickbeard irc channel #sickbeard on freenode for more.

thezoggy:
closing in lieu of upcoming submission from myself with itofzo changes

PR #770, opened 2013-11-09
Changed base_url from tvdb to thirdparty to fix wrong order for American...

xcorp:
... dad among others


xcorp:
Ignore this, merged to the wrong branch

PR #772, opened 2013-11-13
Update the FreeBSD init script to comply with the standards in the FreeBSD Porter's Handbook

mraspberry:
Updated the FreeBSD init script to comply with the rules set forth in the FreeBSD Porter's Handbook. This ensures that the init script complies with rc(8) if a FreeBSD port for Sick Beard is made.


thezoggy:
what if the user uses https for sickbeard..this wouldnt work

PR #775, opened 2013-12-01
Correct checking for TV Download Dir deletion

skardach:
I've noticed that when TV download dir set up in Post-Processing settings is a symbolic link and a single downloaded episode file is processed, my download directory gets deleted. Fix is rather straightforward and makes sure that real paths are compared, not only the canonical forms of directory paths.


thezoggy:
all new code goes into the development branch first. please submit it towards that.

PR #783, opened 2014-01-05
Small speed improvement by using 301 redirect where apropriate.

bfg1981:


thezoggy:
this can be closed out, unless you think the ,301 should be added somewhere appropriate? (the only place i can think of is the 404 handler?)

bfg1981:
I feel a little bit like we're not discussing the same issue, I don't understand what this has to do with 404?

This patch is extremely narrow.
Let's say your SB installation is available at http://sickbeard.example.com, when accessing this URL, you are automatically redirected to http://sickbeard.example.com/home, this is not changed at all.

The only difference is that this may be cached, thus avoiding unnecessary lookups on subsequent use of the site. I think the default is 302 (if I remember correctly).

Indeed, adding a '/' on the redirect would make it even faster, I think.

The only reason to not accept this patch, IMHO, is that you don't want to cache this, which is a perfectly reasonable position.

On the other hand, this is a minor issue, I discovered this, when trying to speed up the access to the front page, which is extremely slow, due to a large database. The major issue being the SQL query on the home page, which takes about 3 seconds. I have tried to move SB to a MySql server, which roughly cut's the time in half, but moving to a more flexible backend involves more than only making the front page work.

bfg1981:
A less intrusive approach though would be to continue using 302, and rather set cache-control, which could be done less specifically

thezoggy:
@bfg1981 our 404 handler redirects the user back to /home/ (it used to do / then that redirected to /home/ -- my recent update fixed that). 

Scenario, lets say the user sets up a reverse proxy incorrectly and keeps trying to go to the wrong sb page. The 404 page redirects them to the proper sb home page using javascript. I'm curious if using the 301 and just having cherrpy redirecting without the page would be better. They may never know they were ever going to a bad url though..

bfg1981:
So I stripped of my patches, and retested, and hopefully I got a little bit wiser.

I understand what you fixed, which is the 404. I tried to provoke a 404, and traced it down. I never intended to address 404's at all, optimizing 404s would be a sideeffect, and never what I intended. IMHO, I would add a small delay before redirecting in the 404's though, to stress that there was an error, and a more informative error message. Eg. this URL does not exist, in 5 seconds you'll be redirected, if this doesn't work, please press this link. I would under no circumstance do any 300 stuff if you don't explicitly know the location. The redirect is IMHO, a very elegant solution, except that it goes to fast.

Let's say you type in "/home/displayShov" where you obviously just have misplaced a character, and you are immediately redirected, you are not given the opportunity to correct yourself, and are immediately redirected, this is not desirable behavior

There is however no 404 on accessing the base URL "/", which is the very narrow issue I was trying to fix with this patch. Today, when accessing the base URL, which is how I, and I suppose many others are accessing SB, in place of writing of the whole URL, with "/home/" on it, you get a 303.

There are a various ways of doing this with 30x status codes.
Among these:
301 Moved Permanently - This is often permanently cached, and should be used if "/" will NEVER be reclaimed
302 Found - This should not be cached, unless an expires header is set.
303 See other - This should never be cached.

I suggest 301, if "/" will never be reclaimed, or alternatively 302, with expiry time of one month (30days=2592000s), if it should be left open for reclamation.

Again, this is a really minor issue, but I'd like to see it closed on it's merits, and not on a misunderstanding.

PR #784, opened 2014-01-05
Make Sickbeard work properly behind a proxy.

bfg1981:
Adds a configuration option BEHIND_PROXY, without exposing it in the UI.

This is only a halfway implementation, it could be argued to leave this always on, or it could be argued to check for this option before headers in PageTemplate, please advise on how to proceed.

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -112,6 +112,7 @@ ANON_REDIRECT = None
 USE_API = False
 API_KEY = None
 
+BEHIND_PROXY = False
 ENABLE_HTTPS = False
 HTTPS_CERT = None
 HTTPS_KEY = None
@@ -313,7 +314,7 @@ def initialize(consoleLogging=True):
 
     with INIT_LOCK:
 
-        global ACTUAL_LOG_DIR, LOG_DIR, WEB_PORT, WEB_LOG, WEB_ROOT, WEB_USERNAME, WEB_PASSWORD, WEB_HOST, WEB_IPV6, USE_API, API_KEY, ENABLE_HTTPS, HTTPS_CERT, HTTPS_KEY, \
+        global ACTUAL_LOG_DIR, LOG_DIR, WEB_PORT, WEB_LOG, WEB_ROOT, WEB_USERNAME, WEB_PASSWORD, WEB_HOST, WEB_IPV6, USE_API, API_KEY, BEHIND_PROXY, ENABLE_HTTPS, HTTPS_CERT, HTTPS_KEY, \
                 USE_NZBS, USE_TORRENTS, NZB_METHOD, NZB_DIR, DOWNLOAD_PROPERS, \
                 SAB_USERNAME, SAB_PASSWORD, SAB_APIKEY, SAB_CATEGORY, SAB_HOST, \
                 NZBGET_PASSWORD, NZBGET_CATEGORY, NZBGET_HOST, currentSearchScheduler, backlogSearchScheduler, \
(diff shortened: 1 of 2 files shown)


thezoggy:
isnt the proxy stuff already there in sb.. by using `web_root` set in the config.ini.

bfg1981:
No, web_root is of of significance for proxying, but not for this particular issue.

However there is a lot of code in the PageTemplate that checks for X-Forwarded headers, but that is not sufficient. I suspect that code might be simplified or made redundant, since this approach offloads those checks for CherryPy. With the old approach, it seems almost everything worked, except for redirects.

thezoggy:
there is already proxy support in sb..

```
        proxies = urllib.getproxies()
        proxy_url = None # @UnusedVariable
        if 'http' in proxies:
            proxy_url = proxies['http'] # @UnusedVariable
        elif 'ftp' in proxies:
            proxy_url = proxies['ftp'] # @UnusedVariable
```

bfg1981:
No, that would be the other way around. If SB requires a proxy to connect outside, that is relevant.

My patch, addresses the issues when SB webpages are served from behind a proxy. This is complementary to the code already in the PageTemplate.**init** .

A more interesting question though is if this should be configurable, or always on, as it is in the PageTemplate code today. If it should be always on, the whole thing collapses to the patch in webServeInit.py

bfg1981:
If needed, I could set up a test environment, to show the issues. Just send a mail to bfg@blenning.no, and I can provide you with a test server login.

bfg1981:
Just wondered what I should do to get this included. This is a blocker when deploying behind a https proxy.

IMHO, I should either integrate all the proxy stuff with the BEHING_PROXY option, or remove the BEHIND_PROXY option, and always assume true.

thezoggy:
i setup a test socks proxy last night, noticed that sb does not send any data through it.. and the code i mention above was for a really old legacy tvdb hack (which actually included anymore.. thus it does nothing).

tried this out with it on or off with a socks proxy configured.. not seeing a difference.

bfg1981:
I need to get better at explaining background.

So I have an application server running SB in http mode on a machine with access to most of the internal network resources.
This sits behind an Apache installation running mod_proxy, which provides sickbeard as an https service facing the internet.

If this patch was not applied, I would press a link, and the link would be to a resource with scheme http rather than https. Using this patch, I set in the Apache configuration 'RequestHeader set X-Forwarded-Proto "https"' and the SB produces apropriate links. This is even on a standard track, so eventually this will most likely be changed to simply Forwarded-Proto, once the standard has been approved. But now I'm just letting CherryPy handle it, and they'll just have to update once the standard is approved.

thezoggy:
so it sounds like your just running sb through a reverse proxy...

bfg1981:
Perhaps awkwardly explained, but absolutely

xombiemp:
I'm confused at what is currently broken when running Sickbeard behind a reverse proxy.  I'm running a similar setup that you described, with Sickbeard behind an Apache reverse proxy, and I have no problems currently.
In my Apache config I have the following settings:

```
ProxyPass http://localhost:8081/sickbeard/
ProxyPassReverse http://localhost:8081/sickbeard/
RequestHeader set X-Forwarded-Proto "https"
RequestHeader set X-Forwarded-Port "443"
```

With this setup, it even redirects back to the right place after performing an upgrade.  So it seems like reverse proxy already works.

bfg1981:
Thanks for the feedback.
You are certain you don't have a redirect from http to https that makes it work?
I haven't upgraded to latest for many months now, awaiting this patch, so it might be a good idea to see if it's suddenly fixed.

xombiemp:
I do have this in my Apache config:

```
<VirtualHost *:80>
    ServerName mydomain.com
    RedirectPermanent / https://mydomain.com/
</VirtualHost>
```

bfg1981:
Upgraded to latest in development branch.
Problem seems to have vanished.
Thanks

PR #785, opened 2014-01-05
Sets a processtitle for Sickbeard, if module is available.

bfg1981:
This is useful to distinguish Sickbeard from other processes running python, where several such are available.

Implemented to only give a warning if the appropriate module is not available.

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -204,6 +204,15 @@ def main():
     # Need console logging for SickBeard.py and SickBeard-console.exe
     consoleLogging = (not hasattr(sys, "frozen")) or (sickbeard.MY_NAME.lower().find('-console') > 0)
 
+    try:
+        from setproctitle import setproctitle
+    except ImportError:
+        if consoleLogging:
+            sys.stderr.write(u"setproctitle module is not available.\n")
+        setproctitle = lambda t: None
+
+    setproctitle(sickbeard.MY_NAME)
+
     # Rename the main thread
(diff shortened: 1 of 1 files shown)


thezoggy:
tried it out.. ide thowing out code warning.. bad indentation
instead of std.stderr shouldnt you use std.stdout - or even show it? could see this confusing people.. could just include it with sb..  but looks like you cant just install it.. but have to compile it.. so going to be a pita on binary builds.

anyways, couldnt get this working.. need to figure out how to get setproctitle to install on windows

bfg1981:
I don't have a lot of python coding experience on windows, seems the right way to do it is with easy_install, which from a quick google search may have some issues on Windows. I'll have to set up a python dev-environment on a Windows machine to see what is going on.

I inspected the code visually, I can't figure out why it should be badly indented. I saw I used a unicode string where it is perfect latin-1, might have something to do with that.

As for using stderr over stdout, it's a warning level message. I'd prefer to log it, but there's no logging facility available yet. I'd argue it belongs on stderr, but the rest of the code seems not to be very consistent on what belongs on stderr and stdout. I'd go any way with that

thezoggy:
its just badly indented per the rest of the document.. we dont use tabs but spaces. generally 4 spaces for each indent.

bfg1981:
I've used 2 spaces per indent, never tabs. I rewrote it to use 4 spaces, and squashed the commit.

PR #788, opened 2014-01-23
More stable startup/shutdown

sestolk:
I could not get Sickbeard to start with the old version so I copied over the sabNZBd rc.d script and changed it so it works with Sickbeard. It now also retrieves the IP and Port from the config.

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -7,44 +7,44 @@
 # Add the following lines to /etc/rc.conf.local or /etc/rc.conf
 # to enable this service:
 #
-# sickbeard_enable (bool):	Set to NO by default.
-#			Set it to YES to enable it.
-# sickbeard_user:  The user account Sick Beard daemon runs as what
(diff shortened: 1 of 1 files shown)


thezoggy:
is this a dupe of https://github.com/midgetspy/Sick-Beard/pull/772 ?

sestolk:
I have no idea, I ran into this problem a few days ago. Haven't looked through 130+ pull requests before I posted this. Ohwell, I might have a look at it tonight.

thezoggy:
no worries, that pull just got pushed.. let us know if it fixes your problems... as i dont think anyone in sb dev runs sb on freebsd

sestolk:
Shutting down actually works fine. I'm still a nooby at linux/freebsd, so I was confused that it worked. Then again, sabnzbd doesn't have an action in the init.d to shutdown either.

PR #793, opened 2014-02-03
Upgrade jquery ui 1.10.3 -> 1.10.4

thezoggy:
Minor bugfixes, mainly for IE9 users.
 -- http://jqueryui.com/changelog/1.10.4/


drewfreyling:
@thezoggy Is there any particular reason we are using a really old school version of jquery? If not, I might submit a pull request to update it.

thezoggy:
Upgrading would require some code rewrites or using migrate plugin. There's no features in the newer jQuery that we are needing at this time... So just needed work for now

-----Original Message-----

drewfreyling:
Ok cool. The lower startup overhead in [1.11.0](http://blog.jquery.com/2014/01/24/jquery-1-11-and-2-1-released/) is nice to have but maybe not if it requires other code rewrites.

thezoggy:
No worries, I do plan to eventually get us up to current version of jquery.. but due to some of the things that was dropped in 1.8.x+ it means I'll have to go update some libs (qtip2) and add libs that they use to deal with the changes. Also it means I'll have to go rewrite some of our own code to deal with the lack of $browser for example. Not a huge deal but also something that can wait until I have some down time :)

PR #794, opened 2014-02-07
XEM Support

echel0n:
Added in XEM Scene Numbering support ported from designgears's Sickbeard branch which is a code port from lad1337's Sickbeard branch.


OverlordQ:
Please god yes. tvdb is utter shit.

echel0n:
Not sure when this will be pulled in but if you can't wait I've added it into my fork already.

PR #799, opened 2014-02-23
Add require/ignore words per show

thezoggy:
Some example cases, note these are PER show:
- Now you could block `webrip` while still getting `web-dl`.
- Now you could seek a specific's groups release like `QRUS` or block `mSD`.
- Now you could set it to grab HD but only if it has `DD5` audio.
- Now you could grab those Top Gear `VFR` (variable frame rate) releases specifically.
- Now you could grab the `Uncut` or `XL` versions of a show (since TVDB does not allow 'duplicate' shows).

Do note that the ignore word logic comes before required words. So be careful on what you ignore. The words are not case sensitive, and if you use multiple words then a release will be rejected if ANY match.

Also the require word logic comes before the proper/repack, this is so that you don't have a proper/repack for (the non uncut / non xl / non vfr) wrong releases that your specifically seeking. If you use multiple require words, ALL of them must be found otherwise it gets rejected.

### DIFF
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -20,6 +20,7 @@ from __future__ import with_statement
 
 import os
 import traceback
+import re
 
 import sickbeard
 
@@ -181,7 +182,7 @@ def searchForNeededEpisodes():
                 if not bestResult or bestResult.quality < curResult.quality:
                     bestResult = curResult
 
-            bestResult = pickBestResult(curFoundResults[curEp])
+            bestResult = pickBestResult(curFoundResults[curEp], curEp.show)
 
             # if all results were rejected move on to the next episode
             if not bestResult:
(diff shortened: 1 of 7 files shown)


sickbeard/search.py
```diff
@@ -456,11 +476,6 @@ def findSeason(show, season):
                 epNum = epObj.episode
                 # if we have results for the episode
                 if epNum in foundResults and len(foundResults[epNum]) > 0:
-                    # but the multi-ep is worse quality, we don't want it
-                    # TODO: wtf is this False for
-                    #if False and multiResult.quality <= pickBestResult(foundResults[epNum]):
-                    #    notNeededEps.append(epNum)
-                    #else:
                     neededEps.append(epNum)
                 else:
```
thezoggy, line 480: note that this else: seems a bit pointless. I was just cleaning up some commented lines.. but this area prob should be reviewed at some time

PR #800, opened 2014-02-23
Huge performance improvements

jvdberg:
When using sickbeard with over 5000+ episodes in the database, it becomes very slow. This pull requests fixes this problem by adding indices and optimizing an sql query.


thezoggy:
your migration schema and the initial schema differ.
tested this out with 2048 shows with 70,000+ episodes. I did not see any speed increase with this.

PR #801, opened 2014-02-26
Fix change metadata

ghost:
Main thing Fix AttributeError exception on stripping actor name when None


thezoggy:
can you also add/take a look at: https://github.com/nexulh/Sick-Beard/commit/a24dbfbb0549afc24199e3732d2d6d1585561e1b

running split on genre when its empty

PR #805, opened 2014-03-04
Download link for the episodes

Ulrar:
- Added a config option to set an URL where the shows can be downloaded
- Using this URL to create a download link for each episode if the option has been set

### DIFF
--- a/data/interfaces/default/config_general.tmpl
+++ b/data/interfaces/default/config_general.tmpl
@@ -154,6 +154,17 @@
                             </div>
                         </div>
 
+                        <div class="field-pair">
+                            <label class="nocheck clearfix">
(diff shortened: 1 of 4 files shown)


thezoggy:
whats the use case for this?

Ulrar:
When sickbeard is running on a remote server, it's a nice way to get the
episodes. I like that better than an apache autoindex.

2014-03-04 19:35 GMT+01:00 thezoggy notifications@github.com:

> whats the use case for this?
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/midgetspy/Sick-Beard/pull/805#issuecomment-36657759
> .

## 

Kévin Lemonnier

PR #807, opened 2014-03-08
Fix FreeBSD init script

Flink:
Sickbeard starts now properly by not invoking `daemon` directly as the main command.
Shutdown is also fixed, the `?pid=${pid}` part in the shutdown URL was missing.

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -17,16 +17,11 @@
 #			Default: /usr/local/sickbeard
 # sickbeard_chdir:  Change to this directory before running Sick Beard.
 #     Default is same as sickbeard_dir.
+# sickbeard_data_dir: Location for storing database, configfile, cache,
+#     logfiles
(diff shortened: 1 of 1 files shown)


Flink:
Ok updated my pull request, the init script should be better :)

Flink:
And an even better script after reading the one used for sabnzbd :)

thezoggy:
should squash and force push to cleanup history

Flink:
Hmm don’t know how to do that, sorry. Seems to involve git rebase -i but there’s nothing to do since my branch is already pushed.

Flink:
Ok tried the rebase and the push --force but the first commit is still there. The rebase does not include it, I don’t understand why… :(

Flink:
Ok found how to do properly :) It’s done!

ghost:
superceded by #908

PR #809, opened 2014-03-14
Update init.freebsd

LukeChannings:
Fix for shutdown issue on FreeBSD.

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -77,7 +77,7 @@ verify_sickbeard_pid() {
 sickbeard_stop() {
     echo "Stopping $name"
     verify_sickbeard_pid
-    ${WGET} -O - -q --user=${sickbeard_web_user} --password=${sickbeard_web_password} "http://${sickbeard_host}:${sickbeard_port}/home/shutdown/" >/dev/null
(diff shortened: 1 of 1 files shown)


thezoggy:
can you check out pull #807 and see if this resolves your issue?

LukeChannings:
The init script references in #807 doesn't hang on shutdown, but it does fail to delete the .pid file if FreeBSD is restarted with SickBeard running, which causes SickBeard to fail to start on boot.

mraspberry:
What version of FreeBSD are you using? I tested #807 on a jail of FreeBSD 10.0-RELEASE and it worked correctly.

josh4trunks:
I recommend you guys stop controlling sickbeard using wget. it responds correctly to the SIG-TERM that the freebsd rc.subr system uses. there is no need to write such complicated functions with unneeded dependencies. also... what freebsd user uses wget over fetch for such a simple task?! =P

you can use my ports init, you need to change the %%VARS%% that are replaced during port make.
https://github.com/josh4trunks/freenas-plugins/blob/master/ports/multimedia/sickbeard/files/sickbeard.in

PR #810, opened 2014-03-22
Fix scene exceptions URL

cdloh:
URL was pointing to old github.com domain. Pointed towards .io
domain.

### DIFF
--- a/sickbeard/scene_exceptions.py
+++ b/sickbeard/scene_exceptions.py
@@ -72,7 +72,7 @@ def retrieve_exceptions():
     query_list = []
 
     # remote exceptions are stored on github pages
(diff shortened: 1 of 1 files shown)


thezoggy:
I can confirm that this should be pulled,

```
>curl "http://midgetspy.github.com/sb_tvdb_scene_exceptions/exceptions.txt" -I
HTTP/1.1 301 Moved Permanently
Server: GitHub.com
Content-Type: text/html
```

vs

```
>curl "http://midgetspy.github.io/sb_tvdb_scene_exceptions/exceptions.txt" -I
HTTP/1.1 200 OK
Server: GitHub.com
Content-Type: text/plain; charset=utf-8
```

PR #811, opened 2014-03-23
WebApp view support for iOS

Jaspur:
Now it also looks more like an app when you add it to the iOS homescreen.


stefantalen:
:+1:

PR #817, opened 2014-04-06
Add 'promo' to default set of IGNORE_WORDS

jbillo:
Recently various NZB indexers have been adding promo/preview ads for shows earlier in the week (eg: Saturday Night Live). SickBeard tends to grab these promotions rather than the actual episode even when searching. Add 'promo' as a default IGNORE_WORDS string to avoid this problem.

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -299,7 +299,7 @@ EXTRA_SCRIPTS = []
 
 GIT_PATH = None
 
-IGNORE_WORDS = "german,french,core2hd,dutch,swedish,480p"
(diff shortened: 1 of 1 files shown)


thezoggy:
promo should not be added to **global** ignore words. people can add it to the show specifically if they want.

jbillo:
I thought it would have been a useful global ignore term (similar to swedish, german, 480p, etc. that are already in this file) since I've seen the issue across a few different shows. If individual show ignore terms are the approach you'd prefer taken, I'll close this pull request.

thezoggy:
The problem is that promo is so vague... I would hate for it to block valid releases if the show name or episode title contained promo.

-----Original Message-----

PR #818, opened 2014-04-06
Set TV processor to ignore the '_unpack' folder

davidrothera:
When downloading with NZBGet it utilises a _unpack folder when
extracting RAR files and therefore these files are not complete.

This diff simply sets the processor to ignore any folders that have
'_unpack' at the start of the folder name.

This was raised off the back of
https://code.google.com/p/sickbeard/issues/detail?id=2453


davidrothera:
@thezoggy - As suggested changed to use startswith rather than contains just to make sure :)

PR #819, opened 2014-04-10
Add xmpp notifications

docmeth02:
This adds support for xmpp notifications to sickbeard. Depends on sleekxmpp.


davidrothera:
Recommend that you squash and rebase your commits.

sickbeard/__init__.py
```diff
@@ -291,6 +291,13 @@
 NMA_API = None
 NMA_PRIORITY = 0
 
+USE_XMPP = False
+XMPP_USERNAME = None
+XMPP_PASSWORD = None
+XMPP_TARGET = None
+XMPP_NOTIFY_DOWNLOAD = 1
```
thezoggy: note that this does not follow our normal naming pattern..  also should default to False
`XMPP_NOTIFY_ONDOWNLOAD = False`

sickbeard/__init__.py
```diff
@@ -1124,6 +1140,14 @@ def save_config():
     new_config['Trakt']['trakt_password'] = TRAKT_PASSWORD
     new_config['Trakt']['trakt_api'] = TRAKT_API
 
+    new_config['Xmpp'] = {}
+    new_config['Xmpp']['use_xmpp'] = int(USE_XMPP)
+    new_config['Xmpp']['xmpp_username'] = XMPP_USERNAME
+    new_config['Xmpp']['xmpp_password'] = XMPP_PASSWORD
+    new_config['Xmpp']['xmpp_target'] = XMPP_TARGET
+    new_config['Xmpp']['xmpp_notify_download'] = int(XMPP_NOTIFY_DOWNLOAD)
+    new_config['Xmpp']['xmpp_notify_snatch'] = int(XMPP_NOTIFY_SNATCH)
```
thezoggy: should be bool() not int() assuming you change to use that instead of an int

thezoggy:
looks like sleekxmpp  does NOT work on python 2.5
 -- https://github.com/fritzy/SleekXMPP

sickbeard/__init__.py
docmeth02: actually it has to be int() here, because otherwise it would write bool to the config file.

docmeth02:
i fixed the bools and naming patterns.
What do you suggest i should do about sleekxmpp? Would it be enough to check the python version and disable xmpp on python < 2.6?

PR #830, opened 2014-05-11
Prepare init.freebsd for FreeBSD port

Sp1l:
- Add datadir option (separate app from data)
- Add webroot option (for proxied setups)
- Remove wget dependency
  - fetch supports RFC1738 Common Internet Scheme Syntax http://user:pass@host:port/path?querystring
  - appending pid=PID works reliably, also with user:pass


ghost:
This is superceded by #908 

There's no need to use fetch or wget to shutdown sickbeard.

PR #831, opened 2014-05-12
Fix ordering for shows with 500+ episodes. Now it works for up to 500k.

rafaelgaspar:
The "Downloads" ordering works by taking the percentage of episodes downloaded vs non downloaded, but it needs to transform this into a integer, previously it was being multiplied by 1000, but that's not enough to sort shows with 500+ episodes, like Simpsons, SNL, etc.

So I changed it to multiply by a million, that way it will work for up to 500k episodes. That should be enough until we evolve into beings that can watch multiple shows at a time while absorbing all of it.

### DIFF
--- a/data/interfaces/default/home.tmpl
+++ b/data/interfaces/default/home.tmpl
@@ -60,7 +60,7 @@
         if (parseInt(nums[0]) === 0)
             return parseInt(nums[1]);
 
-        var finalNum = parseInt((nums[0]/nums[1])*1000)*100;
+        var finalNum = parseInt((nums[0]/nums[1])*1000000)*100;
         if (finalNum > 0)
             finalNum += parseInt(nums[0]);
 


thezoggy:
this was already brought up in an earlier pull but we fixed it internally... it scales depending on the show with largest count..  basically we divide by show count to get relative value to size so its easy to sort.. the precision however at 1000 was not enough as you mentioned. changing it to a really high amount by default just wastes cpu cycles calculating a higher number than needed.. hence our 'fix'. ill get it pushed out soon

rafaelgaspar:
If there is a better implementation done, that's better.

As for wasting CPU cycles it actually doesn't, the number never overflows out of INT32 spectrum so it's the same number of cycles with the modern JS VMs. But of course if there is a better implementation there is no reason not to use it.

Thanks for your time anyway.

thezoggy:
can you update to the latest dev and confirmed this solved your issue?

PR #812, opened 2014-03-25
Adding pushbullet to the list of notification providers

davidrothera:
Added https://www.pushbullet.com to the list of supported Providers.

Checked and tested with all methods (test, snatch, downloaded)

PR #833, opened 2014-05-15
Pushbullet Notifier

pironic:
This is a pretty simple change that adds a new notifier... Pushbullet.com


pironic:
seems i wasn't the first person to implement nearly the exact same methods... 

https://github.com/midgetspy/Sick-Beard/pull/812

PR #834, opened 2014-05-22
pep8 cleanups

ekohl:


thezoggy:
for pep8 we ignore, E501,E201,E202,W0141
we only make safe changes and things that wont cause concerns for others.. also not to do the whole project at once but when were working on existing code.. (otherwise we could just autopep8)
for the menu stuff we do make the exception for E241 (multiple spaces after ,) but all new code we try not to do it.

some of your whitespace changes appear to introduce indent issues...  we use 4 spaces not tabs.

there is a pull request for pep8 to add in: Add checks for `not x in y` and `not x is None`
since technically its not part of pep8 as of yet -- http://bugs.python.org/issue21540

PR #839, opened 2014-05-28
Update boxcar.py to support boxcar2

nbonfire:
Support boxcar2's new API
Replacing the boxcar username field in the settings with the access token should be enough to work.


nbonfire:
i just realized someone else already did the work.

PR #851, opened 2014-07-08
Added support for the new append method from NZBGet 13+

Prinz23:
NZBGet 13 will probably be released in the next few days. This version has a new combined append method that supports both nzb urls and content. Also the return value was changed form boolean to integer (NZBID or error number).

The old methods append and appendurl are now Deprecated, so it's more future proof to add the new method now.

### DIFF
--- a/sickbeard/nzbget.py
+++ b/sickbeard/nzbget.py
@@ -129,12 +129,17 @@ def sendNZB(nzb):
                         nzbcontent64 = standard_b64encode(data)
                         nzbget_result = nzbGetRPC.append(nzb_filename, sickbeard.NZBGET_CATEGORY, add_to_top, nzbcontent64)
 
-        # v12+ pass dupekey + dupescore
-        elif nzbget_version >= 12:
+        # v12 pass dupekey + dupescore
+        elif nzbget_version == 12:
             if nzbcontent64:
                 nzbget_result = nzbGetRPC.append(nzb_filename, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, nzbcontent64, False, dupekey, dupescore, "score")
             else:
                 nzbget_result = nzbGetRPC.appendurl(nzb_filename, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, nzb.url, False, dupekey, dupescore, "score")
+
+        # v13+ has a new combined append method that accepts both (url and content)
+        # also the return value has changed from boolean to integer (Positive number representing NZBID of the queue item. 0 and negative numbers represent error codes.)
+        elif nzbget_version >= 13:
+            nzbget_result = True if nzbGetRPC.append(nzb_filename, nzbcontent64 if nzbcontent64 is not None else nzb.url, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, False, dupekey, dupescore, "score") > 0 else False
 
         # v9+ pass priority, no dupe info
         else:


sickbeard/nzbget.py
```diff
@@ -129,13 +129,18 @@ def sendNZB(nzb):
...
+        elif nzbget_version == 12:
             if nzbcontent64:
                 nzbget_result = nzbGetRPC.append(nzb_filename, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, nzbcontent64, False, dupekey, dupescore, "score")
             else:
                 nzbget_result = nzbGetRPC.appendurl(nzb_filename, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, nzb.url, False, dupekey, dupescore, "score")
 
+        # v13+ has a new combined append method that accepts both (url and content)
+        # also the return value has changed from boolean to integer (Positive number representing NZBID of the queue item. 0 and negative numbers represent error codes.)
+        elif nzbget_version >= 13:
+            nzbget_result = True if nzbGetRPC.append(nzb_filename, nzbcontent64 if nzbcontent64 is not None else nzb.url, sickbeard.NZBGET_CATEGORY, nzbgetprio, False, False, dupekey, dupescore, "score") > 0 else False
```
thezoggy, line 142: can you rewrite this so it isn't on one long line (move the if/else inside the current indent branch)

ghost:
I already did:
https://github.com/itofzo/Sick-Beard/commit/225db89b84a5451a75c9643360c4a6eb1f633464

@Prinz23 Are you ok with it that I kept you as the author?

sickbeard/nzbget.py
Prinz23, line 142: This is a inline "if else" statement, there are no indention in this since writing it in one line it has special meaning in python. This is similar to the

_condition_ **?** _value_if_true_ **:** _value_if_false_ 

conditional operators in other languages like C, only that in python it is written like this:

_value_if_true_ **if** condition **else** _value_if_false_

This is the correct "pythonic way" (fastest and uses less memory) to do it in python 2.5 or higher

To use additional variables uses more memory (if nzb content is given = variable will be copied, which can be many MB) and makes things slower (way slower on resource restricted devices)

thezoggy, line 142: id rather waste one variable for the result for readability (really just segmenting branch logic and re-use variable) and make it similar to the other branching logic..  while we dont stick to the whole 80 char per line, this is just asking for trouble. yes there are pythonic ways to do things, and then there are pythonic ways of doing something but keeping it readable/maintainable to others :)

Prinz23, line 142: So you waste many MB's (Full Season nzb for example) of memory, which are may not even available on resource restricted devices (and this can result in crashes on these devices) like I use....

Prinz23, line 142: referring to: https://github.com/itofzo/Sick-Beard/commit/225db89b84a5451a75c9643360c4a6eb1f633464

nzb_content = nzbcontent64

is coping the entire content of an nzb in another variable.

thezoggy, line 142: gotcha, yeah itofzo can just move the `nzbget_result = nzbGetRPC.append...` line to each branch condition to reduce the need to re-assign the variable (like it is for all the other conditions).

i have to ask, what device are you running sb on that has that limited resources (and whats the specs... i'm just curious)?

Prinz23, line 142: Mipsel PVR with 256 MB RAM (don't forget OS, nzbget, ... has to run on it too) And there are other devices like raspberry pi that also have not much resources. 

I did do it this way because each extra line costs performance on such devices... since python is a interpreter language. 

Of course it can be written the old python 2.4 complex way. I would think since this syntax is standard since the already very old python 2.5 it's not necessary and anyone using/coding python for more just playing around should know this syntax. And in the end you can add a comment line to explain the inline if else for people that don't understand it.

In the end if we don't use the new syntax we can simply don't do it at all, since the old methods (append, appendurl) are still supported by NZBGet for compatibly reasons and simply using the new append method in the old python syntax is a bit pointless to me.

thezoggy, line 142: regardless of your personal viewpoint on how you think things should be done, you have to realize that we have our own preferences especially since we have to maintain the code / support it / deal with the problems if bugs arise. hence why we generally prefer code that is more readable and inline with the rest of the code in the file, im not asking for less efficient and we are more than happy to optimize things where we can (notice that ito has already addressed your concern and force pushed the change). for example we try to use pep8 guidelines (there are still several we dont follow though)

about the whole **ternary** format, I'm quite used to it in javascript world.. it just can get unwieldy quickly.
if you want to get truely code snobbish you can probably rewrite just about every line of python code into something that is shorter/smaller and go off on tangents about using shorter variable names/similar names so it can optimize better.. etc. writing shorter code doesnt always make it better.. 

if i have to spend more than a second looking at your line of code to understand or just read it (because of the length/overly complex/to much branching logic/etc) then there is an issue. ito and myself both felt the same thing as we both brought it up to be addressed (he just happen to do it himself while i mentioned it directly in the pull).

anyways, your variable concern with ito's change has been addressed.

PR #855, opened 2014-07-20
Elandril quality

Elandril:
Additional/modified regex strings for Quality.nameQuality to recognize some additional commonly used release names:
1. ["sdtv", "(pdtv|dsr|tvrip)"] => SDTV
2. ["dvdrip.(wide|ws)?", "sdtv"] => SDDVD
3. webhd == web-dl
4. bd == bluray

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -153,6 +153,8 @@ class Quality:
             return Quality.SDTV
         elif checkName(["(dvdrip|bdrip)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDDVD
+        elif checkName(["dvdrip.(wide|ws)?", "sdtv"], all) and not checkName(["(720|1080)[pi]"], all):
+            return Quality.SDDVD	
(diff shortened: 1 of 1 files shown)


thezoggy:
where are these 'valid'? everywhere i see this stuff its fake/spam stuff...  or foreign things we dont want.

Elandril:
The formats 1 and 2 are found on some older releases (more than 3 years old), when the naming scheme for tv shows hadn't been so established yet. Therefore it is mainly important for backlogs.

The formats 3 and 4 are found even on very current releases of some new shows.

thezoggy:
sdtv/webhd is used by foreign groups (german). which we do not want to support in sickbeard. sorry but we have no plans to include your changes.

Elandril:
Are there really no plans to support any foreign groups? :-(
Is this because of the added complexity or to keep Sick-Beard "english-only"?

PS: If this is an unchangeable no-go, do you know of an multi-language alternative to Sick-Beard?

thezoggy:
why we cant just simply support non-english/foriegn shows:
http://www.sickbeard.com/forums/viewtopic.php?f=8&t=8484

now about forks, there is generally a fork for each major language which have their own providers, their own exceptions, their own changes.

the one you probably want is: https://github.com/cytec/Sick-Beard/

Elandril:
Thanks for the link!
I'll close the pull request.

PR #858, opened 2014-07-24
HTTPS Chain Certificate Support

namidairo:
If one needed a chain certificate for their browser not to complain about their certificate.

As far as I know, sticking it into server.crt does not work?

Default is for a chain.crt in the sickbeard data directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ sickbeard.db*
 autoProcessTV/autoProcessTV.cfg
 server.crt
 server.key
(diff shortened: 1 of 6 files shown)


thezoggy:
this appears to be valid per cherrypy's docs..

realslacker:
I think I duplicated this... looks like you caught a couple things I missed and I got a couple you did.

https://github.com/midgetspy/Sick-Beard/pull/907

PR #861, opened 2014-07-29
Added support for Pushbullet API v2. Dependencies: requests >= v2.0

Jaap-Jan:
Notifications work. Needs requests to work. On Debian and Debian- derivates you can install the package python-requests. It needs to be v2.0 or newer, so if you are on Debian Wheezy, you need the version from wheezy-backports.

It uses a slightly modified version of this library: https://github.com/randomchars/pushbullet.py. The modifications are the removal of the listener and file upload to avoid having dependencies on websocket and magic.


Stanzilla:
cool!

thezoggy:
using `requests` means no python 2.5 support. so until midgetspy gives the go ahead to drop support we cant accept this at this time.

Jaap-Jan:
I've done a new commit which adds support for Python 2.5 via urllib2. Tested with Python 2.5 on Debian Wheezy with Python2.5 from Squeeze. If I use the wrong API key this gets logged (triggers an HTTPError):
2014-08-05 01:45:29 ERROR    CP Server Thread-2 :: Pushbullet: Error sending notification: authentication token is missing or invalid.

If I change https to httph, this triggers an URLError, which is also logged:
2014-08-05 01:48:04 ERROR    CP Server Thread-10 :: Pushbullet: Error sending notification: unknown url type: httph

Also fixed a bug. I used the wrong syntax to access a dictionary. Appeared when testing if the exception handling routines in the class PushBullet25 worked correctly. :blush:

data/interfaces/default/config_notifications.tmpl
```diff
@@ -649,6 +649,53 @@
                     </fieldset>
                 </div><!-- /libnotify component-group //-->
 
+                <div class="component-group clearfix">
+                    <div class="component-group-desc">
+                        <img class="notifier-icon" src="$sbRoot/images/notifiers/pushbullet.png" alt="" title="Pushbullet" />
+                        <h3><a href="http://pushbullet.net/" rel="noreferrer" onclick="window.open('${sickbeard.ANON_REDIRECT}' + this.href, '_blank'); return false;">Pushbullet</a></h3>
```
thezoggy: pushbullet.net is not correct

PR #868, opened 2014-08-25
Return exit code 0 when stopping a non-running process in ubuntu init script

jeroenj:
If any other error occurs in the stop script it will just raise an error as expected and not start sickbeard.

This also allows the restart command to start sickbeard when it's not running yet.


thezoggy:
@midgetspy looks safe to merge

PR #869, opened 2014-08-26
add Transmission support for torrent downloads

skoef:
Inspired by couchpotato's support for adding torrents directly to Transmission, I've added this to sick beard. Big pro versus blackhole: separate machines for sick beard and Transmission and allowing me to specify a directory where to download episodes.


papertigers:
Is this one capable of removing a torrent after a ratio/time is reached like CP?

skoef:
Not yet, but I'm willing to implement it. Also, the current default download directory is Transmission's default, but I think it could as well be the configured download directory for the serie. I'll merge my branch with the master and get back with an update soon.

fz6:
Check the code from this fork for the ratio/seedtime commands:
https://github.com/echel0n/SickRage/blob/ef621a3b170977e7b2e45a5b107327e31b56f768/sickbeard/clients/transmission.py

antongeorgiev:
Not sure if this is helpful but there could be something useful in this fork: https://github.com/bricky/Sick-Beard

skoef:
Apparently, that fork will be gone very soon. I did look into the ratio thing, but couldn't even find the option in couchpotato.
Is that even the most logical place to configure that option, instead of just in transmission itself?

JackDandy:
Hello @skoef 

This project is no longer active, perhaps try [SickGear](https://github.com/SickGear/SickGear) - you can drop in your existing sickbeard.db and continue with more features and search performance improvements.

PR #880, opened 2014-09-30
Add requirements.txt

andersonvom:
Since Sick-Beard depends on Cheetah, having a requirements.txt (a
standard way to declare python project dependencies) makes it easier for
users and developers to get it up and running, while also allowing for
better dependency tracking, since we can describe the specific required
version of Cheetah that is necessary.

Then all users/devs need to do is:

```
$ pip install -r requirements.txt
```

Other requirements can be added in the future as necessary.


thezoggy:
why do you include pysqlite ? sqlite support should be native to python..

andersonvom:
Oh, I included everything I had to install to get it running, but now that you mentioned, I may have ended up using native sqlite instead of pysqlite.  My bad, I'll remove it and submit another PR.

andersonvom:
Updated the PR to include only Cheetah

thezoggy:
you need to rebase+push. still seeing the pysqlite

andersonvom:
Done

chancez:
It would be even nicer to make sickbeard into a proper python package, with a `setup.py` and such. However it could require rewriting a bunch of imports if it wasn't done properly to begin, and would need a few scripts to be setup so it's still easy to run.

PR #890, opened 2014-10-28
Update Plex notification settings and behavior

mmccurdy:
Just a few long-needed tweaks to the Plex client and server notifications.  Allow sending client notifications directly to the modern Plex Home Theater client (instead of having to configure it as an XBMC client), allow authenticated library update requests (important for users who lock down local network access), updated Plex icon & description.


thezoggy:
the 'modern notifications' is just xbmc json api... which is why there is a note that if you want to send notices to PHT you use the xbmc notifier with the correct port.

mmccurdy:
> the 'modern notifications' is just xbmc json api...

Right, but that's pretty confusing, isn't it?  If people want to set up notifications for Plex, why not let them do it all in one place?  There's no need for the old Media Center notifications anymore (that client hasn't been supported for almost two years), and this is a non-breaking change for people who already have it set up and working using the XBMC settings.

thezoggy:
they used to be together (plex was subclassed from xbmc) but when xbmc changed from web api over to json.. it made upgrading xbmc code a pain and in the process broke plex.. thus we split them. so then plex continuing to 'borrow' xbmc's code is now using their api stuff which means once again they agree... but not everyone runs the latest and greatest.. thus we have to support legacy and latest and greatest. anyways.. we do need to just drop legacy xbmc (kodi now) and force people to upgrade? but there are other things to fix...  just low on the priorities as it 'works'

mmccurdy:
Okay, I'm not totally sure I understand why we wouldn't want to fix this, but I backed out the client notification changes since they were actually the least important part of this PR.

What's left:
- Updated logo and description (mainly to remove the "on a Mac connected to your TV" wording which hasn't been a limitation for 4-5 years).
- Re-phrased the client notification field descriptions to make it clearer that they only apply to Plex Media Center (the old version of the client).
- Re-purposed the username and password fields to allow authenticated access to the Media Server for update notifications.  The client doesn't (and has never) allow setting a username and password for the notification HTTP server, so these fields were misleading (at best) as-is.  This change allows SB to send library update notifications to Media Servers on different networks and/or servers that have local access restrictions in place.  In my opinion this is the highest priority piece of this PR.

BrianInAz:
Hi - is there any update on when this might be pulled into SB? Really looking forward to regaining ability to have SB tell Plex Media Server to update the list of available TV after processing a new show (bullet #3 above).

Thanks, love SB!

lumaxis:
I'd love some update on this since especially the authentication part is becoming significantly more important with the upcoming "Plex Home" feature which is gonna require authentication for everyone by default. See also [this](https://support.plex.tv/hc/en-us/articles/203948786-Plex-Home-Security-Changes) knowledge base article.

mmccurdy:
@thezoggy, I'd like to understand any lingering concerns with this PR and address them ASAP.  The recently-released parental controls feature enforces authenticated access to the endpoints required by the updater causing it to fail for lots of folks.  Unfortunately this seems to be driving even more Plex users away from SB (since other competing apps provide authenticated access), but as a SB holdout I'd like to see them stick around.  If you have no intention of merging this, it would also be useful to know that so we can look elsewhere.  Thanks!

imajes:
:+1: 

really need to get this going to make SB even better. @mmccurdy07 - what other alternatives are more Plex friendly, out of interest?

srazer:
This is very much needed now as I have enabled Plex Home and Sick Beard only gets the following when trying to update Plex:

PLEX: Error while trying to contact Plex Media Server: http error : error 401 : Unauthorized

MattZera:
Yeah... something needs to be done as sickbeard can no longer update plex for me

imajes:
@thezoggy - can we get this merged? and/or can we get some more maintainers if there's not been enough time to look at this? :+1:

KevinAnthony:
If this is merged in, will sickbeard still be able to communicate with older versions of Plex?

mmccurdy:
> If this is merged in, will sickbeard still be able to communicate with older versions of Plex?

Yes.

Your question made me realize that in some (presumably very rare) cases, if people were using the old non-functional user/password fields with invalid plex.tv credentials, that could prevent updates from happening.  That has now been fixed.

chancez:
Got tired of waiting for this, and when testing it I got the following error:

http://localhost:8081/home/updatePLEX/

```
500 Internal Server Error

The server encountered an unexpected condition which prevented it from fulfilling the request.

  File "/Users/chance/Sick-Beard/cherrypy/_cprequest.py", line 660, in respond
  File "/Users/chance/Sick-Beard/cherrypy/lib/encoding.py", line 193, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "/Users/chance/Sick-Beard/cherrypy/_cpdispatch.py", line 25, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/Users/chance/Sick-Beard/sickbeard/webserve.py", line 2483, in updatePLEX
    if notifiers.plex_notifier.update_library():
  File "/Users/chance/Sick-Beard/sickbeard/notifiers/plex.py", line 192, in update_library
    auth_tree = etree.parse(response)
UnboundLocalError: local variable 'response' referenced before assignment
```

EDIT:

Seems that I was using the wrong username/password which caused this I think. Although, ideally this exception would be caught in that case.

chancez:
Also, the `Test Plex Notifications` button does not work if you do not have the PMC box filled out. IMO this should work without PMC because all I'm using this feature for is plex media server updates, as my clients are _not_ PMC.

mmccurdy:
> Seems that I was using the wrong username/password which caused this I think. Although, ideally this exception would be caught in that case.

Dumb bug that I just introduced trying to make things safer (oops), fixed now.

> Also, the Test Plex Notifications button does not work if you do not have the PMC box filled out.

I tend to agree, but this is not a regression from the current behavior (it has only ever tested PMC notifications) and it's a slightly more impactful change so I'm inclined to keep it out of this PR in the interest of getting it merged more quickly.

chancez:
Understandable. Maybe could we have a second button for testing only the server, and not the PMC clients? This could be added later, just a thought.

weezywee:
@Developers - Is this going to be merged?
@Anyone - How do I merge this with my local copy until it is officially merged?

midgetspy:
I don't have plex installed atm to test this but I've pulled it as it seems there are a couple of you who have had success with this pull. Thanks.

Ockingshay:
i can confirm this is working...thanks for fix

CyberMew:
+1 for master

benmccallum:
Let's get this on master and into a release!

PR #893, opened 2014-11-11
Properly redirect when behind a mod_rewrite proxy

Lx:
Sick Beard uses redirects after various operations, e.g. adding a show
and saving configuration changes.

When behind an Apache reverse proxy configured using .htaccess rules
only (i.e. where modification to the global Apache configuration is not
preferable or possible), CherryPy by default builds redirect URLs based
off the internal host instead of the external one, breaking sessions
for users accessing Sick Beard by external URLs.

This could be regarded as a shortcoming of the CherryPy implementation
that Sick Beard is using rather than a Sick Beard problem directly.

In any case, the changes in this commit allow Sick Beard to respond to
both requests made directly and those made through a reverse proxy,
with redirects working as expected in both cases.

Thanks to Joel Rivera from the CherryPy Users Google Group for his
support in preparing these changes:

https://groups.google.com/d/msg/cherrypy-users/P87eOftSb_E/BDU6_1sDjtoJ

### DIFF
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -112,6 +112,7 @@ def initWebServer(options={}):
         'tools.gzip.mime_types': mime_gzip,
         'error_page.401': http_error_401_hander,
         'error_page.404': http_error_404_hander,
+        'tools.autoproxy.on': True,
     }
 
     if enable_https:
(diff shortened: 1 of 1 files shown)


Lx:
I've explicitly tested changing the port and switching between HTTP and HTTPS.  There were no issues.

Lx:
To be clear, I've tested these changes both behind a reverse proxy and with no proxy.  In both cases, everything works as expected and redirect URLs have the correct host, port etc.

Lx:
@midgetspy, do you have any outstanding concerns regarding this pull request or should it be okay to merge?

midgetspy:
Sorry for the delay - looks great, thanks.

PR #895, opened 2014-11-13
newznab - tweaks

thezoggy:
- If there are more than 100 results for a query, fetch up to an additional 300 results (4 hits total).
- Handle when newznab sites take their API offline more gracefully.
- Add logic to fall back to query by string if searching by tvrage id returns 0 results
  
  > Right now we search using tvr if matched, if we have 0 results then sb gives up. Which may be bad if they only have newznab/one provider. Since not all newznab sites may have the tvr associated correctly or even have tvr lookups enabled. This now makes it fall back to searching by string (as if we had no tvr) in that scenario, which can result in a few more hits but at least then we can get results.. so in the end it could mean less hits as were not trying forever with no chance of getting anything.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -21,6 +21,7 @@ import email.utils
 import datetime
 import re
 import os
+import time
 
 try:
     import xml.etree.cElementTree as etree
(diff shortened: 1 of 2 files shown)


thezoggy:
Once this is pulled, you can close https://github.com/midgetspy/Sick-Beard/pull/674 and https://github.com/midgetspy/Sick-Beard/pull/838

PR #898, opened 2014-11-20
Create cpasbien

cyril51:


thezoggy:
this is a french torrent provider. not something we want to pull as we do not want foreign items coming in. It may be welcomed in the french or multi-lang forks such as https://github.com/cytec/Sick-Beard/

PR #902, opened 2014-12-02
issue:occasional incomplete copy; comparison confirms valid file

ghost:
added the comparison code a week ago and it has resolved recent file quality issues

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -31,6 +31,7 @@ import traceback
 import urllib
 import urllib2
 import zlib
+import filecmp
 
 from httplib import BadStatusLine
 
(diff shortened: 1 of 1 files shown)


thezoggy:
can you provide a use case / your issue (workflow)?

ghost:
```
Hi there, yes - over the past few weeks sickbeard has started having
files in the destination location as being incomplete or a little
bit messed up when watched. This is since the 2nd last time I
updated sickbeard via the "you're x commits behind" button/link - I
also updated it last week again too hoping it would resolve the
issue but no luck.
During this time I checked out my hardware, thinking a disk might be
going bad, but the checks all came back good. Also I've not had any
other corruption in other applications or files either. As
mentioned, the problem started after updating sickbeard.
Performing a manual copy would result in a perfect copy, no issues
when viewing.

So, I found the filecmp module, added it to my copy and everything
has been perfect since then. Thought I'd share what I'd done in case
others might have been experiencing the same issue (I did a quick
search but didn't find people mentioning anything like my
situation).

Maybe it's just me.

On 03/12/14 12:09, thezoggy wrote:


  can you provide a use case / your issue (workflow)?
```

midgetspy:
This sounds like a local issue but I'm glad this works for you. If filecmp ever has a permanent failure you'll end up in an endless copy loop though ;-P

PR #905, opened 2014-12-10
Added support for IPTorrents

MattZera:
I've added support for IPTorrents


MattZera:
the user number and torrent pass can be retrieved when the user generates an rss feed - I'm not sure where to put that note

when the user number or torrent pass are incorrect iptorrents returns a 404 error. There isn't a way to distinguish if its an authentication error or if the rss just isn't working at the moment so I let the 404 get caught where it would normally and put in the error log that the authentication may be incorrect. that seemed like the easiest and best way to do it.

ebarrere:
:+1: please merge

PR #906, opened 2014-12-14
handle new newznab XML namespace for NZBgeek

claudebossy:
For some reason, sometimes (that's weird...) the newznab namespace in the XML is not the one defined in "sickbeard/providers/newznab.py". I made it work with this quick fix but I don't get why the issue happened (could it be related to my config?).

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -408,6 +408,8 @@ class NewznabProvider(generic.NZBProvider):
                 if parsedXML.tag == 'rss':
                     items = parsedXML.findall('.//item')
                     response = parsedXML.find('.//{http://www.newznab.com/DTD/2010/feeds/attributes/}response')
+                    if not response and "nzbgeek" in self.url:
+                        response = parsedXML.find('.//{http://nzbgeek.info/feeds/attributes/}response')
 
                 else:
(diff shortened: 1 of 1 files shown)


fauxfiction:
This is because of a certain provider changing their DTD to a URI they control because the newznab one has apparently been dead for a few years. I guess they mean that there is no namespace definition at the URI.

Instead of special casing a provider specific DTD, it's probably preferable to either select the node agnostic of the DTD URI or select a node that looks like what we want. Since DTD agnostic checking doesn't seem easy (without some nasty pre-parsing hacks to namespace defs), I'd suggest that L410 simply be changed to something like:

```
response = parsedXML.find('.//*[@offset]')
```

This way, we're looking for an element with the offset attribute which is what we really want. It's not guaranteed to be the correct `newznab:response` tag though so we could also try some things like:
- `nodes = parsedXML.findall(...)` and `assert len(nodes) == 1`
- verify the local tag name like `assert response.tag.endswith("response")`

That's still not a solid guarantee but without either enforcing a DTD that seems dead (AFAIK) or swapping to using lxml's etree (with better XPath support so we can use `*[local-name="foo"]`), using `.//*[@offset]` seems to be a reasonable fix to me.

midgetspy:
This should be fixed in #909

PR #908, opened 2014-12-29
rewrite freebsd init script

ghost:
Greatly simplify FreeBSD init script. 

The use of wget and other hacks to shutdown sickbeard was incorrect; sickbeard captures signals and shuts down cleanly.

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -16,8 +16,6 @@
 #			as root.
 # sickbeard_dir:	Directory where Sick Beard lives.
 #			Default: /usr/local/sickbeard
-# pidfile:		The name of the pidfile to create.
-#     Default is sickbeard.pid in sickbeard_dir.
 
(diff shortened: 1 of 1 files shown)


mraspberry:
Nice. Like your solution much better. Hopefully this will get merged soon

mraspberry:
One thing about this, SickBeard.py can't start because /usr/local/bin isn't in the PATH so "/usr/bin/env python" doesn't work to find python. Maybe the init script should call python via %%LOCALBASE%%/bin/python directly?

ghost:
If sickbeard existed in and was installed by FreeBSD ports then we would use the "shebangfix" feature to automatically replace #!/usr/bin/env python with the correct path. It doesn't exist in ports currently so this will have to suffice. Manually installing sickbeard on FreeBSD does come with a couple rough edges, but that's to be expected.

PR #909, opened 2014-12-30
Locate NawzNab XML response node agnostic of namespace URI

fauxfiction:
Instead of requiring that NewzNab providers use a fixed DTD URI for the
newznab XML namespace, search for the nodes with an 'offset' attribute
and then use the one with a namespaced tag ending in 'response'.

This is an alternative to midgetspy/Sick-Beard#906 but it could feasibly
end up doing more work per API query since it queries the etree twice
and might end up with lots of nodes to filter out if the API data is not like
we expect it to be.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -407,7 +407,22 @@ class NewznabProvider(generic.NZBProvider):
 
                 if parsedXML.tag == 'rss':
                     items = parsedXML.findall('.//item')
-                    response = parsedXML.find('.//{http://www.newznab.com/DTD/2010/feeds/attributes/}response')
+                    # Find all nodes with offset and total attributes
+                    offset_nodes = parsedXML.findall('.//*[@offset]')
+                    total_nodes = parsedXML.findall('.//*[@total]')
+                    # Our set of candidates for the response tag is the union
+                    # of the nodes with offset and total attributes since we
+                    # expect both to be present
+                    response_candidates = set(offset_nodes) & set(total_nodes)
+                    # Filter out any nodes which are not namespaced 'response'
+                    # tags
+                    response_nodes = [node for node in response_candidates if node.tag.endswith('}response')]
+                    # Verify that one and only one node matches and use it,
+                    # return otherwise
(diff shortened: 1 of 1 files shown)


thezoggy:
would this actually work on python 2.6? i seem to recall its xml parsing quite limited..

fauxfiction:
Hmm, you're right. Attribute selection was only added in ElementTree1.3 (ie. Python 2.7). I'll double check with a Python 2.6 but I think this will raise an exception. Using lxml is an alternative but changing to a non-stdlib xml parser just for this seems like overkill (and it only supports down to 2.6, no 2.5).

We could walk the ETree manually with a tree iterator with the `rss` tag as the root - maybe we could try an XPath attribute search first and fall back to that behaviour as well.

Any other suggestions to avoid the hardcoded URI or do we have to live with it?

_Edit_: Yep, exception on 2.6.9 and presumably anything prior to that:

```
>>> r.find(".//*[@offset]")
...
  File "/tmp/Python-2.6.9/Lib/xml/etree/ElementPath.py", line 93, in __init__
    "expected path separator (%s)" % (op or tag)
SyntaxError: expected path separator ([)
```

fauxfiction:
So here's an alternative where we can add extra attributes to the etree to keep track of the local tags. We potentially lose context but that's kind of the point of not wanting to deal with the URI. This code snippet could mutate the tree after the auth check or we could use the same idea to just collect nodes for later processing (item nodes and response nodes) leaving the tree unchanged.

```
import xml.etree.ElementTree as ET

fname = "./api_results.xml"

item_nodes = []
tree = ET.fromstring(open(fname).read())

for el in tree.getiterator():
    el.local_tag = el.tag.split("}", 1)[-1]
```

This change is really intended to cater to changes that providers make to their standard newznab rollout though. Is it appropriate to allow this at all? Is it possible and appropriate to instead subclass for the different behaviour? It'd be pretty easy to extract the DTD URI to a class variable and override that in a provider specific subclass but then the UI would need to be able to list subclasses instead of just having a section for adding NewzNab providers.

PR #910, opened 2015-01-13
config.clean_url may add trailing /'s in unsuitable circumstances.

anthonyryan1:
Also adds some simple tests so we can avoid future regressions.


midgetspy:
Thanks for the pull request, pulled.

PR #912, opened 2015-01-16
modified URL openers so they do not check SSL certificates (fixes #2551)

arucard21:
This restores the sickbeard behaviour from before https://www.python.org/dev/peps/pep-0476/ (which enabled certificate verification by default in python)

Because of the default certificate verification introduced in PEP-0476, sickbeard could no longer connect with sabnzbd, which uses a self-signed certificate, and sabnzbd could no longer connect with sickbeard through the autoProcessTV.py post-processing script because sickbeard also uses a self-signed certificate. 

This was reported in this issue https://code.google.com/p/sickbeard/issues/detail?id=2551

Reproduction in sabnzbd script:
- configure sickbeard to use https
- configure autoProcessTV.cfg with ssl=1
- run sabToSickbeard.py (you can use a fake, non-existing filename as argument, e.g. test.mkv, but you do need an argument)
- you get an error "certificate verify failed" (after fix is applied, you see "Opening URL: https://..." to indicate it succeeded)

Reproduction in sickbeard
- configure sabnzbdplus to use https
- configure sabnzbdplus as nzb method an test the connection
- you get an error "Unable to connect to SAB: socket error" (after fix is applied, this connects successfully)

Note: 2 newlines were changed but I could not change them back, hope it's not a problem.


thezoggy:
note that this would mean that you assume the user base is using python 2.6+, sb still maintains support for python 2.5.. so it'd be on midgetspy to drop 2.5 in support of this change.

arucard21:
I could put every change in a `if sys.version_info >= (2, 7, 9):` statement, similar to what's done in autoProcessTV.py. That should keep it functional for earlier python versions.

thezoggy:
`if sys.version_info < (2, 6):` would suffice

thezoggy:
yes, that would work..

arucard21:
Just added the version checks to the fix as well as the ssl module import, since the ssl module seems to be new since 2.6 and the context part only available since 2.7.9 as well.

bassebaba:
I guess this will fix this problem?

PUSHOVER: Notification failed.error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)

Using python 2.7.9

thezoggy:
Why use SSL encryption if its not trusted... A proper fix would be to contact site owners to fix their cert / get proper URL to use so the cert is happy. Also need to verify if this is only a sickbeard or python issue.

-----Original Message-----

bassebaba:
Zoggy, I have no idea, I just wanted to log this error.
It's not my doing, its SickBeard.

I just installed a new machine yesterday with python 2.7.9 and thats the error-log I get when using the built in pushover functionality.

https://github.com/midgetspy/Sick-Beard/blob/90a7622506cb2e9074c23d8adc79c23e8878d04b/sickbeard/notifiers/pushover.py

It seems that it uses "https://api.pushover.net/" and that site has a valid ssl certificate as far as I can see?

I'm sorry if I'm hijacking this thread, should I open a new one?

cdavin:
Just to add to the discussion and backup bassebaba. I have contacted both sites that I have had sickbeard SSL issues with and both have not only denied any issue with their SSL cert they have provided proof via SSL verification sites. Also, this issue does not affect any of my other application that use python and connect to https APIs (i.e Couchpotato).

arucard21:
@bassebaba This should indeed fix the "certificate verify failed" problem. 

@thezoggy This PR doesn't actually change anything in sickbeard, it reverts its behaviour back to the way it was. The problem here is that python's default behaviour wrt certificate validation has changed, which in turn changed the default behaviour in sickbeard.

As for a proper fix, in the case of @bassebaba your suggestion would be the preferred solution, but as @cdavin mentioned, the failed verification doesn't seem to be caused by invalid certificates on their side. So i think that's a separate problem that should be investigated and addressed on its own.

The reason why I think unvalidated certificates on an SSL connection are still useful, is that self-signed certificates are often still used on self-hosted applications, like sabnzbd and sickbeard itself. Since they are self-hosted, it's easier to trust them and making sure that self-signed certificates are trusted is usually a bit more difficult. Maybe this can be revisited when the Let's Encrypt service (https://www.letsencrypt.org/) is up and running and can be integrated.

As an improvement, I'm also thinking of changing all calls to open a url and make them use the SickbeardURLOpener or AuthURLOpener from classes.py (and changing those two openers so they use the more up-to-date urllib2, but that is secondary). This would allow us to easily change how URL's are opened in the future and provides more consistent behaviour within sickbeard.

bassebaba:
Thans for all your feedback. I "fixed" pushover-notification by using the code from @arucard21
- import ssl
- add "context=ssl._create_unverified_context()" to all urllib2.urlopen(...)

This is obviously a workaorund but I don't have the time to read up on python+urllib2 to understand why it doesnt work (or how to get it working).

midgetspy:
Yes if you could please make a generic function which does the ssl import and the URL open I will pull this. Preferably we only have a single place where the version check is done and then we can call that from all these places. Thanks.

thezoggy:
so apparently i missed the fact that 2.7.9 came out which is causing the ssl nightmare within sb.

https://www.python.org/downloads/release/python-279/
- HTTPS certificate validation using the system's certificate store is now enabled by default.

so yeah something deff needs to be done to handle this, revert 2.7.9's changes.
you can just update getURL under helpers.py [https://github.com/midgetspy/Sick-Beard/blob/development/sickbeard/helpers.py#L170] which covers a lot of the url requests...  the notifiers however cant use getURL due to the way its imported (itozfo mentioned this before in dev chat).. so you'd have to go handle those separately (just the https calls)

arucard21:
I've now updated helpers.py so it can be used to retrieve content from URLs (this was a great tip btw). I've also removed the 2 openers from classes.py since they didn't seem to be used at all. I did not change anything in the lib and cherrypy directory since I wasn't sure if they should be using the sickbeard method in the first place.

The new getURL() is (mostly) backwards compatible so (almost) none of the calls needed to be changed (it no longer accepts a data object as argument, but does accept a Request object which can contain that data object). 

The getURLFileLike() method returns the file-like object that is returned by open() for low-level access to it. It also allows you to let it re-throw the exceptions it catches. This is mostly used by the notifiers that use the exceptions to figure out goes wrong for each specific notifier.

I've also added a readURLFileLike() method to read the file-like object and return it as a string. This also decompresses the object, if needed. This is mostly just used by getURL() but it seemed like a good idea to make it available as a separate method anyway.

I've tested the new getURL method and it seems to work fine, but now I seem to have some problems with my python configuration. It won't import the helpers module from the notifiers and I'm not that familiar with python imports. I'm pretty sure that it's a problem in my own configuration because some of the code that I haven't touched at all would also give me some problems.I get the following when I try to start sickbeard:

```
File "<path to sickbeard dir>/sickbeard/notifiers/xbmc.py", line 27, in <module>
    from sickbeard import common, helpers
ImportError: cannot import name helpers
```

If someone can help me fix this, I can test it again, or if someone else can check if this works as expected. I can't test all the notifiers and providers myself anyway, so I'm not sure what the best way would be to test these. I checked the getURL() method by filling in my sabnzbd in the config and testing the connection. This seemed to work fine.

toddwasson:
arucard21, I'm not positive but I think the problem is because of a circular dependency.  SickBeard.py imports sickbeard, which eventually imports metadata/generic, which imports helpers, which imports notifiers, which imports xbmc, which in turn now imports helpers.  Hence helpers imports helpers and creates a circular dependency.  You might be able to fix this by moving your new functions to a package other than helpers and importing that from the notifiers instead.

arucard21:
@toddwasson Thanks, you were absolutely correct. I managed to fix it by calling the helper method from the sickbeard module (so sickbeard.helpers.getURL() instead of helpers.getURL() ).

I've now managed to test this and it seems to work correctly (after some small fixes). I've tested the following:
- the sabnzbd connection using the "test" button
- the xbmc notifier using the "test" button
- searching for an episode (manually) with the Sickbeard Index provider and a custom Newznab provider (running locally) enabled 
- snatching an episode found through the Sickbeard Index provider and uploading the nzb to sabnzbd

So the fix should work correctly in most cases. I couldn't test all the different notifiers though, but I tried to change as little as possible there so they should all work correctly.

toddwasson:
@arucard21 For what it's worth, it's working great for me now.  Thanks for making the edits, and I'm happy I could help!

thezoggy:
Don't forget about SB's API. Probably needs updating 

-----Original Message-----

arucard21:
@thezoggy I'm not sure what you mean with "SB's API". Did I miss something that also accesses URL's in sickbeard?

thezoggy:
@arucard21 i was just thinking that there was some url calls in the webapi.. there is but none of them are https. so all good.

autoProcessTV/autoProcessTV.py
```diff
@@ -131,7 +134,10 @@ def processEpisode(dir_to_process, org_NZB_name=None):
         password_mgr = urllib2.HTTPPasswordMgrWithDefaultRealm()
         password_mgr.add_password(None, url, username, password)
         handler = HTTPBasicAuthHandler(password_mgr)
-        opener = urllib2.build_opener(handler)
+        if sys.version_info >= (2, 7, 9):
```
midgetspy, line 137: If the version check can't be shared with the getURLFileLike can you at least make a common pythonVersionVerifiesHttpsCertificates or something similar please?

lib/growl/gntp_bridge.py
```diff
@@ -55,7 +57,10 @@ def get_resource(self,key):
 			return self.resources.get(resource[1])['Data']
 		elif resource.startswith('http'):
 			resource = resource.replace(' ', '%20')
-			icon = urllib.urlopen(resource)
+			if sys.version_info >= (2, 7, 9):
```
midgetspy, line 60: Same request here as in autoProcessTV

midgetspy:
Aside from the comment above I think this looks great. Will merge once those last few version checks are cleaned up, thanks very much!

arucard21:
I deliberately left those 2 places as they were for the following reasons:
- with gntp_bridge.py, I wasn't sure if it was desirable to import the sickbeard helper method since they might be third-party libs and not actually part of sickbeard itself. So I should be able to import the sickbeard helper method, if that's acceptable (which now seems like it is). 
- with autoProcessTV.py, it needs to run as a standalone script so it shouldn't be able to use any helpers from the sickbeard codebase. I can abstract the version check away into a separate method within autoProcessTV.py itself, but I don't really see the value in that. So  could you explain what exactly you meant by "make a common pythonVersionVerifiesHttpsCertificates"?

arucard21:
I just verified this in the code and lib/pythontwitter and lib/tvdb_api also use urllib to open URL's. These might also need to be changed although they connect to public URL's which should actually have valid SSL certificates so default behaviour of validation might be correct there. None of these libs include anything from the sickbeard modules so I'm not sure if I should be the first to do that. Ideally, the fixes should be applied upstream for these libs or at least made in such a way that they _can_ be applied upstream. 

I'll admit though that I'm not exactly clear on what you meant by "make a common pythonVersionVerifiesHttpsCertificates" so maybe I'm misunderstanding you entirely.

midgetspy:
Yeah you're right, I hadn't realized that it was in autoProcessTV.py. Thanks very much for the submission!

arucard21:
Thanks for merging. 

I just wanted to mention one more thing. I've included a "validate" parameter in the getURL helper methods which can be used to enable SSL certificate validation on python 2.7.9+ anyway. This can be used by others to enable validation where this is actually useful, like connecting to public SSL sites. 

For example, the pushover notifier can be set to be validated by someone who knows how it works and can test it. But since this could have unintended side-effects, as already reported by @bassebaba and @cdavin, I think it would be best if those parts were improved by someone who actually knows how they work. So the option to validate SSL certificates is still there, for anyone that might want to use it.

kz-fox:
Looks like this change broke the Twitter notification plugin. I get:

```
2015-03-06 11:28:25 SEARCHQUEUE-RSS-SEARCH :: TWITTER: Error Sending Tweet: error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)
```

arucard21:
@austinholbrook This change is not responsible for that error. The problem you have with the Twitter notification plugin is the exact same one that is being fixed by this PR. However, I did not apply the fix to the twitter notifier (sickbeard/notifiers/tweet.py) because it uses an OAuth client to connect to twitter. I am not familliar enough with this to apply the fix to that as well and even if I did try to fix it, I couldn't test if it works since I don't use Twitter. So perhaps somebody else can fix this, either in sickbeard/notifiers/tweet.py or in lib/oauth2 (perhaps there's a new/different version of the oauth2 library that already includes this fix).

It should be noted though, that the twitter notifier connects with the twitter URL which should (and does indeed seem to ) have a valid SSL certificate. This is similar to the problem @bassebaba mentioned earlier, which could indeed be circumvented with this change (which unfortunately isn't available to the twitter notifier). Perhaps it would be better if someone could find out why the valid SSL certificate seems to fail validation. So this would seem like an entirely different bug.

thezoggy:
This was fixed years ago by using requests instead of httplib/oauth/etc. Requests requires python 2.6+ however. Midget just never wanted to drop python 2.5. The oauth2 we use is quite old.... Because the newer version requires 2.6+. 

If you look at the forums people have been complaining of this issue for several  notifiers..  plex, twitter, etc. 

-----Original Message-----

kz-fox:
Before this pull request was merged, even providers with valid SSL certificates were failing. I could modify the Twitter notifier to skip the SSL verification but that seems to defeat the purpose of SSL.

arucard21:
@austinholbrook  Actually, that's exactly what this change is about. It uses a common method to connect to the URL with its default behaviour configured to skip the SSL certificate verification.

The common method allows you to configure and modify the default behaviour for sickbeard, but it also allows you to enable verification of the SSL certificate anyway (using a supplied parameter). For notifiers that connect to public URL's which should have valid SSL certificates, like Twitter and Pushbullet, it would be best to enable SSL certificate verification when using the common method and fix whatever is causing them to fail the validation. 

But that's not what this PR was about, this was about making the _default_ behaviour in sickbeard skip the SSL certificate verification (as was the default before this was changed in python). Obviously, there's still many things that can be improved when it comes to SSL connections, this being one of them.

kz-fox:
Right, my understanding was that a function got made to establish the connection without SSL cert verification. However if someone figures out why python is rejecting these valid certs it should be fine to reenable the verification for providers as well. Perhaps adding an option in the settings to allow self-signed certs or something similar.

I'll look into what I can and open a new bug report.

arucard21:
That sounds like a good idea. 

Just to let you know, I did a basic test with the common getURL() methods that I created for this PR and they seem to be able to verify the SSL certificates correctly. I tried several https sites, including the pushbullet site mentioned earlier, just accessing the base URL without any authentication or query string. So the problem with the notifiers might be specific to those notifiers or perhaps related to their authentication method.

If anyone wants to test this for a specific notifier, you can enable validation of the SSL certificate by adding "validate=True" as parameter to the getURL() or getURLFileLike() method used in that notifier. Note that this will only work on python 2.7.9+.

aptalca:
arucard21, not sure if you saw, but this pull somehow caused issues with updating xbmc, plex and the like. See comments here for more info: https://github.com/midgetspy/Sick-Beard/commit/79cb8f207bef8613db9e47eb392666af059df3d0

Thanks

arucard21:
@aptalca Sorry about that, I couldn't test with many different notifiers but I tried to stay as close to the original behaviour as possible. Strangely enough, I did test with XBMC and it worked just fine, but maybe there's a difference in configuration. If there's any help needed, let me know, but it seems like #923 already fixes the problem.

PR #913, opened 2015-01-19
Filename sanitization fix

webdestroya:
Fixed this issue:

`Season.23/How.Its.Made.S23E28.Wood.Garage.Doors;.Salt.Spreaders;.Animatronic.Dinosaurs.mp4`

which should be:

`Season.23/How.Its.Made.S23E28.Wood.Garage.Doors.Salt.Spreaders.Animatronic.Dinosaurs.mp4`

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -159,7 +159,7 @@ def sanitizeFileName(name):
 
     # remove bad chars from the filename
     name = re.sub(r'[\\/\*]', '-', name)
-    name = re.sub(r'[:"<>|?]', '', name)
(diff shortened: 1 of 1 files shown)


midgetspy:
Hmm, the purpose of that function is simply to make it into a valid non-hidden filename, not to make it more human-readable. I'll have to put a little more thought into whether we want that function to also do other things...

thezoggy:
personally i like the `;` as its used to separate the ep segments. if you trip it... well it loses its meaning. although it is quite cosmetic in the end..

webdestroya:
I guess my preference was to prevent anything that requires escaping when accessing files in shell, which included `;`, `$`, etc

PR #917, opened 2015-02-13
Retain only X episodes (per show setting, optionally delete old episodes automatically)

rootsuid:
Database Changes:
- creating new field on tv_shows : retain_episode_count (integer), default 0. Added to TVShow form. Default (0) ignores this setting entirely.

" Note: A value of 0, will retain all episodes (default behavior)."
" The maximum number of episodes to retain on disk. When new episodes are downloaded,"
" old episodes are deleted, keeping the most recent episodes (determined by air date)."

Implementation:
- At the end of the post processing, and renaming, a method call to TVShow will check retention limits.
- If the show has more than the specified number of episodes on disk (marked as downloaded), the older (by air date) episodes are deleted.
- As each episode is deleted, the episode status is reset to SKIPPED (related files deleted as well, nfo, srt, etc).

Use case:
- You want to have the most recent N episodes available for casual watching, but are uninterested in archiving or long term storage.
- In particular, news shows or time sensitive media. (ie, a daily show)

Error Cases:
- If permissions on the filesystem disallow deletion of the base episode file, it is left as is (still marked "downloaded"). It will be attempted again after the next episode download.
- If the base episode file is deleted, but the related files could NOT be removed, the episode is still considered deleted, and status is updated accordingly.

### DIFF
--- a/data/interfaces/default/editShow.tmpl
+++ b/data/interfaces/default/editShow.tmpl
@@ -91,6 +91,13 @@ Results <b>without one</b> of these words in the title will be filtered out<br /
 Separate words with a comma, e.g. "word1,word2,word3"<br />
 <br />
 
+<span class="showLegend">Retain Episode Count (this show):</span><input type="text" name="retain_episode_count" id="retain_episode_count" value="$show.retain_episode_count" size="5" /><br />
+Note: A value of 0, will retain all episodes (default behavior).<br />
+The maximum number of episodes to retain on disk. When new episodes are downloaded,<br />
+old episodes are deleted, keeping the most recent episodes (determined by air date).<br />
+<br />
+
+
 <input class="btn" type="submit" value="Submit" />
 </form>
 
--- a/sickbeard/databases/mainDB.py
+++ b/sickbeard/databases/mainDB.py
@@ -27,7 +27,7 @@ from sickbeard import encodingKludge as ek
 from sickbeard.name_parser.parser import NameParser, InvalidNameException
 
 MIN_DB_VERSION = 9  # oldest db version we support migrating from
-MAX_DB_VERSION = 18
+MAX_DB_VERSION = 19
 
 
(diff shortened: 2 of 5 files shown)


thezoggy:
this logic could introduce a huge problem when episodes are wanted and only a season pack is available.. this logic runs.. deletes ep1 of a season.. users default status is wanted.. thus it gets marked as wanted.. search is ran... season pack is found.. rinse and repeat.

rootsuid:
After an episode is deleted, the status is set to skipped (not user default, even if 'wanted'). If it was set to user/show default - it would keep refetching even without packs. 

When I say "delete" - I mean it deletes the downloaded files, and status is set to SKIPPED, the episode record itself is never deleted (or that would also cause fetch loops, on top of the fact that it's nice to have all the metadata in the database).

If the pack is extracted, and each episode is processed individually - then this would be ok (it would extract the whole pack, and still only retain the last N episodes). I haven't seen the pack process though, so I'm not sure there yet.

rootsuid:
After reading through the pack processing, it does appear that it processes each episode individually.
If a whole season was wanted (12 episodes), but retention was set to 4 episodes for some reason - it could download a pack, then delete the oldest 8. Once removed, it would mark those 8 skipped. 

If only 1 episode was wanted, but it was only found via a pack, it does look like it would download them all - and process each.  After that download finished, it would delete the oldest 8 anyway - marking them skipped.  (line 505 of search.py if anyone wants to review).

Either case is wasteful on bandwidth; but if you're using this feature to follow live air dates (only keep most recent X), there likely won't be repacks anyway. If there are, you'd end up with what you expected (only recent 4), but would have wasted bandwidth getting it.

Thanks for catching the case, thezoggy. I hadn't seen any of the repack/proper area of SB.

xortim:
What's the hold up here?

PR #923, opened 2015-03-06
Use getURL correctly when gzip encoding is a valid response

ppslim:
Plex notifications were failing due to use of getURLFileLike returning
a file like object that had not been gzip decoded in line with addheader
request.

### DIFF
--- a/sickbeard/notifiers/nmj.py
+++ b/sickbeard/notifiers/nmj.py
@@ -127,8 +127,7 @@ class NMJNotifier:
         try:
             req = urllib2.Request(updateUrl)
             logger.log(u"NMJ: Sending NMJ scan update command via url: %s" % (updateUrl), logger.DEBUG)
-            handle = sickbeard.helpers.getURLFileLike(req)
-            response = handle.read()
+            response = sickbeard.helpers.getURL(req)
         except IOError, e:
             if hasattr(e, 'reason'):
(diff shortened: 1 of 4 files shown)


c360e5f1:
Thank you for this! I will test this later this weekend and report back on if it works for me.

ppslim:
Thanks, Please do report back.

I have been able to near full testing on Plex, but I don't use the other services, which may have been subject to the problem (encoding dependant).

It would be nice to get somebody using Plex more than I am on this.

ppslim:
Just a note. I spotted and corrected an XML parsing error I introduced in plex. Etree is no longer passed a file like object, so now needs to parse a string

DarkHelmet433:
There's an error in here.  An exception is thrown with a getroot attribute error at line 194-ish of plex.py. According to: https://docs.python.org/2/library/xml.etree.elementtree.html#parsing-xml
when switching from .parse() to .fromstring(), the getroot isn't required.

ie:
               xml_tree = etree.parse(sickbeard.helpers.getURLFileLike(url))
               media_container = xml_tree.getroot()
should be:
                media_container = etree.fromstring(sickbeard.helpers.getURL(url))

Your version still has the getroot and that doesn't work.  It throws an exception if you click "Update Plex" from the nav bar, and records the attribute error in the log instead of updating after a download.

PR #929, opened 2015-03-18
Losslessly compress images.

droob:
ImageOptim reports that losslessly compressing the images in the project saves
60K, an average of 24.6% per file.


thezoggy:
most of the images arent even used... can prob just get nuked. and the space saving we got from things like the flags were negligible.

droob:
Fair enough!

PR #930, opened 2015-03-18
Fix Plex Update library

depassp:
The new URL opener helper by default requests gzip content encoding.  Might as
well also use the decompression helper.

This fixes issue 2571 https://code.google.com/p/sickbeard/issues/detail?id=2571


arucard21:
Already fixed in #923 along with some other notifiers

PR #935, opened 2015-03-24
Renamed XBMC to Kodi

roelofr:
This only includes the visible mentions on screens, these include:
- configuration screen
- View series "Update in Kodi" link (on the navbar)
- Log output (All mentions of XBMC are renamed to Kodi)
- Icons

No classes or any other required code is changed.


thezoggy:
kodi doesnt have a http api (legacy), only the json rpc

PR #936, opened 2015-03-29
Added InteliJ IDEA project files to .gitignore list

spidercensus:
If you're using pycharm or InteliJ, this is needed to keep your stuff from being tracked in Git.


mfeltscher:
Use `.git/info/exclude` for these purposes.

PR #939, opened 2015-04-07
Fixed "coercing to Unicode" TypeError when triggering exception

arucard21:
When some exceptions are triggered, you get the following error:
`TypeError: coercing to Unicode: need string or buffer, instance found`

It was also reported in https://code.google.com/p/sickbeard/issues/detail?id=2571

This should fix that.


levibuzolic:
I think the issue here is that `url` can actually be a request object and not just a url string. With this commit in place I still get the same error. It's not until I remove `url` from the log statement that I get any further into debugging the underlying issue.

arucard21:
You're right, I added a check that makes sure we always have the string value of the URL available for logging. It should work correctly now, even if you still include the url in the log statement.

Mantene:
Please implement this!

eamonnfaherty:
+1

arucard21:
Closed because stale

PR #943, opened 2015-05-06
Pathing issueIssue 2571, 2585, 2591

Mad-Piggy:
Patch for issues:
- Issue 2571:   Plex (and possibly pushbullet) notifications no longer work
- Issue 2585:   Patch for: TypeError: coercing to Unicode: need string or buffer, instance found
- Issue 2591:   coercing to Unicode: need string or buffer, instance found

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -209,7 +209,7 @@ def getURLFileLike(url, validate=False, cookies = cookielib.CookieJar(), passwor
         # Before python 2.7.9, there was no built-in way to validate SSL certificates
         # Since our default is not to validate, it is of low priority to make it available here
         if validate and sys.version_info < (2, 7, 9):
-            logger.log(u"The SSL certificate will not be validated for " +  url.get_full_url() + "(python 2.7.9+ required)", logger.MESSAGE)
+            logger.log(u"The SSL certificate will not be validated for " +  req_to_str(url) + "(python 2.7.9+ required)", logger.MESSAGE)
 
         opener = urllib2.build_opener(urllib2.HTTPCookieProcessor(cookies),
                                       MultipartPostHandler.MultipartPostHandler,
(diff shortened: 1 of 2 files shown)


sickbeard/helpers.py
```diff
@@ -224,42 +224,42 @@ def getURLFileLike(url, validate=False, cookies = cookielib.CookieJar(), passwor
...
-        logger.log(u"BadStatusLine error while loading URL " + url, logger.WARNING)
+        logger.log(u"BadStatusLine error while loading URL " +  req_to_str(url), logger.WARNING)
         if throw_exc:
             raise 
         else:
             return None
 
     except socket.timeout:
-        logger.log(u"Timed out while loading URL " + url, logger.WARNING)
+        logger.log(u"Timed out while loading URL " +  req_to_str(url),, logger.WARNING)
```
taylorludwig: tried this pr, got syntax error here. (Double commas)

Mad-Piggy: Fixed. Thx for letting me know.

dimfeld:
Thanks for the fix! I'm happy to have my Plex notifications working again.

jakswa:
Thanks from me as well. Plex updates don't error anymore for me

CyberMew:
Is this on the Dev or master branch yet?

dimfeld:
I don't think so. I just manually patched it in to my local copy from the PR.

KevinAnthony:
Is there a timeline on when this will be merged in?

Mad-Piggy:
It is indeed not yet implemented in dev or master. But this is something we have to w8 for midgetspy to do. If you realy want to pull this patch, you can from my repository: https://github.com/Mad-Piggy/Sick-Beard

arucard21:
Just wanted to mention that the fix for the coercion errors is also provided in #939. That problem is actually bigger than the Plex notifier but shouldn't have any impact on the functionality. It's the logging of an error that goes wrong.

I think the problem with Plex was that the password manager wasn't supplied with the getURL() call, so it failed. But due to the unrelated problem with the logging of that error, it wasn't clear what exactly was going wrong. So perhaps it would be easier to have this merged if both PR's focus on fixing only 1 thing, with #939 for the logging fix and this PR for the Plex fix. I think including too many things in 1 PR might make it harder to determine if it can (and should) be merged (but I could be wrong about that).

Note that with only the plex fix your branch would no longer cause an error when connecting to plex and the error would no longer need to be logged so the coercion problem wouldn't occur anymore either. Just to say that it would still remain entirely functional.

jordansaints:
Just wanted to say thanks for the fixes! Saved them in a stable branch and am never updating SB again.

jpanski:
Any eta on when this pull request will be merged into development or master?

dimfeld:
I think development on SickBeard has ceased... Personally I've switched to SickRage which is somewhat based on this one but still under active development.

jpanski:
Wow, that's really sad if true.  I hope that's not the case and he's just on a sabbatical.

dimfeld:
Yeah, hard to tell. He hasn't said anything on the forums at sickbeard.com either, so nobody really knows for sure...

Amadeus-:
By the way, if folks want to apply these patches (and/or any others while the primary developer is missing), there are good instructions here:  http://stackoverflow.com/a/28729484.      Basically, you can create pull requests between the requests here and your own forked repository.

PR #946, opened 2015-05-18
Page does not load with saved Pushover priority

chrisdambrosio:
This fixes an issue where the notifications configuration page does not
show the saved pushover priority in the select form control. The template
was comparing the html string value to integers which was always false,
hence never returning the selected attribute.


chrisdambrosio:
Is there anything I can do to help get this merged?

PR #846, opened 2014-06-21
Added kickass (katproxy) torrent support.

fabiosoft:
Added support to the biggest torrent provider. 99% of episodes now will be found!

PR #948, opened 2015-06-01
Add a new torrent provider : kickasstorrent

loic-moriame:
EZRSS provider is broken.
I just add one of my favourite provider : kickass torrent

I only take code from an other PR (https://github.com/midgetspy/Sick-Beard/pull/846) but @fabiosoft dont let is Github's fork.
So I do mine (and now I can clone my repo)


chilicheech:
I haven't looked at the code but this is something I want to. That one is also my favorite provider and it would be very useful to add it.

:+1:

PR #949, opened 2015-06-12
Added Kickass provider (more up-to-date)

arucard21:
I've noticed that there are already 2 PR's for adding Kickass as provider in sickbeard, namely #846 and #948 (which seems to be based off of the same code as #846). These seem to be using some older code and based on the URL used in kickass.py they may not work anymore since kickass moved to a different domain.

In the first of these 2 PR's it's recommended to look at the junalmeida branch's kickass provider in order to implement this and this PR is based on the code from the junalmeida branch (which I've updated a while ago to make it more stable). I've been running this code in my own sickbeard with the junalmeida branch for months now and have switched to the midgetspy branch a few days ago where this is also working without any problems. So it's been tested on the midgetspy branch as well and seems to be working correctly. (Note to anyone wanting to switch from junalmeida branch to midgetspy, you need to manually set your db_version to 14 in sickbeard.db before starting the midgetspy branch in order to migrate properly).

This provider code uses the standard getURL() helper methods to retrieve the results from the kickass website, which should make the retrieval more stable.

I've also added an option to provide an alternative URL so users can override the default URL pointing to the kickass website, in case they change domain again or if you just wanna use one of the proxies. This was done in the junalmeida branch and it seemed like a good feature to have. This is done through the normal interface for torrent provider options that is already available in sickbeard. Note that the input is parsed as a URL so it can be tolerant of variations on the URL's representation (e.g. `kat.cr`, `https://kat.cr`, `https://kat.cr/` or `kat.cr/` are all be valid input for this option)


arucard21:
The generic provider was logging any response of None as an error, but that is not always the case. So I changed it to no longer log it as an error by default anymore. This has no impact on the functionality, it only changes the logging of 1 error. It would probably be better to just remove the getURL() method from the generic provider and use the helpers.getURL() method directly, but that should be part of a different PR.

VeNoMouS:
It didn't originate from @junalmeida... i pushed to @junalmeida, I wrote that code.. which you stripped my name off... at least give credit where credit is due... 

additionally.. this code is out of date now due to kickass website changes..

arucard21:
I didn't mean to offend anyone, I removed the author information because the history of the branch showed that you, me and junalmeida all contributed to the code so I thought the author information was out-of-date. But I'll add it back with the updated usearch-based code you wrote. I intended to give credit where it's due, I credited the code from the junalmeida branch, not junalmeida himself. This PR was just about porting that code to this branch.

VeNoMouS:
I don't mind you using my code :) just felt it was rude it was stripped.. 

also re the code, if you notice now on kickass when you hit "page limit" via json, it redirects you to an HTML page.. so what i'm saying is... if you excedded the pages that it actually has (not that we know how many to begin with) it just redirects out of the json to a HTML search  page version of the first page of results... 

if you check the latest commit, you will see I moved away from json search and over to usearch using "rss" for XML...

Also, it was denying user-agents with "python" in them... as is torcache.

arucard21:
I updated the PR so it uses your usearch-based code now and added back your author info. I did change the code so it retrieves the torrent URL instead of the magnet link since magnet links don't always work (they don't work for me anyway). I also added the actual search URL that is used to the log (at debug level). I think this is useful for users that want to check if the URL is actually correct so they can verify if any problems that occur are from sickbeard or kickass. Might be useful for you as well. 

The user-agent string I'm using is defined by sickbeard in sickbeard/common.py. It doesn't seem to include "python" in it, so this should still be good.

arucard21:
Closed because no longer relevant

PR #947, opened 2015-06-01
Update pynma.py

Bakkoda:
The current NMA domain has expired and is not responding. Updating the API_SERVER variable to a working domain fixes this issue.

PR #950, opened 2015-06-15
NotifyMyAndroid Endpoint Fix

GrahamBlanshard:
See Notify My Android's [author comments in Google+](https://plus.google.com/110805020263741864090/posts/3mfTq6cYwTf) regarding the issue.

Summary:
`nma.usk.bz` is no longer supported as an endpoint for NotifyMyAndroid. Instead, `www.notifymyandroid.com` is recommended.

### DIFF
--- a/data/interfaces/default/config_notifications.tmpl
+++ b/data/interfaces/default/config_notifications.tmpl
@@ -855,7 +855,7 @@
                 <div class="component-group clearfix">
                     <div class="component-group-desc">
                         <img class="notifier-icon" src="$sbRoot/images/notifiers/nma.png" alt="" title="NMA" />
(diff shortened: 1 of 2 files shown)


mweber26:
I was just about to submit a pull request for the same fix :+1:

arucard21:
Just FYI, this is the same as #947 though this PR also updates the site address in the user interface.

pingers:
This is a library... https://github.com/uskr/pynma

I updated to the latest version of that lib and it works fine now. I.e. 1.0 which uses www.notifymyandroid.com.

I suggest pulling in the latest version of the pynma library in place of this PR.

PR #952, opened 2015-06-27
Remove umask changing

jeroenh:
The changed umask is very annoying, it causes all new show directories to be automatically created with a 077 umask. This means that other process wanting to read from that directory have to be run as the same user.

In this pull request I've removed the umask changing from SickBeard, should someone want to run with a more restricted umask, they can set it themselves in the init file (where it belongs IMHO).

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -94,10 +94,6 @@ def daemonize():
 
     os.setsid()  # @UndefinedVariable - only available in UNIX
 
-    # Make sure I can read my own files and shut out others
-    prev = os.umask(0)
-    os.umask(prev and int('077', 8))
-
     # Make the child a session-leader by detaching from the terminal
     try:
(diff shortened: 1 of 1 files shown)


thezoggy:
this is for sb's config files? not your show data... your show data it uses the parent folder to know what to set. so if sb is setting your new shows as 077 its because you have your top level root dir set to 077. fix your problem.. dont cause new problems for others with this code change

jeroenh:
The (deleted) code sets the umask for the SickBeard program to 077.
This means that all new files and directories created by SickBeard are thus created with 077. Existing files keep their original umask, unless you delete them completely and create them again.

If you want configuration files to have a more specific mask, then you should either explicitly create them with that umask at the time that you create them, or you should manually change them once they are created. I'll make another pull request with a more specific fix so that configuration files are created with a 077 umask, and others with the regular one.

PR #956, opened 2015-09-27
remove image resizing and add caching

giantspook:
resizing looks bad on retina displays


thezoggy:
if you dont want resizing just dont install PIL, and if you have to have PIL and dont want resizing use the noresize option..

PR #968, opened 2016-01-12
Add support for Slack as a notifier

mallen86:


medjai:
yay! Finally slack notifications. Was just about to supply this myself!

medjai:
You are missing the option though to send message as the authenticated user token. This way you don't have to enter an icon for your bot it will just take it off of the bot's profile

mallen86:
Ah ha, I was not aware that was a feature. I'll add it soon.

PR #972, opened 2016-03-07
uWSGI Support

rmackinnon:
Greetings,

My server at our home runs uWSGI, and I really wanted to have the app run and be managed uniformly with the rest of our apps.  So I created a new daemon to run with uWSG with the added feature of taking flags from the environment. Tear down and restart are a little not so good, but I believe that is due to additional threading being managed by the Sickbeard daemon thread. The time it takes to tear down and start up a new version just hits a timeout and gets destroyed.  All in all, running on my own uWSGI emperor managed server, config and save all seem to work.


sickbeard/webserveInit.py
```diff
@@ -201,4 +204,4 @@ def autoproxy(
     )
 
     cherrypy.server.start()
-    cherrypy.server.wait()
+    #cherrypy.server.wait()
```
midgetspy, line 207: What is the purpose of these changes?

rmackinnon, line 207: I found that the .wait() method doesn't take any real effect beyond delaying the time the service comes up to an available state.  Is there a defined reason to delay here?

PR #977, opened 2016-05-19
Refining FreeBSD Init Script

tehgooch:
Fixed bad behaviour when shutting down the system. Sickbeard would fail to start after a restart/shutdown. The reason is that rc.shutdown issues the faststop command which doesn't delete the pidfile. As a result sickbeard would not start up complaining of pidfile existing on start. You can reproduce on a FreeBSD system by shutting down and starting up. Sickbeard will not be running. If you try to start it manually it will complain about the pidfile. You can also reproduce by running `service sickbeard faststop` then `service sickbeard start`. The same behaviour will be noticed. To resolve you must manually delete the pidfile and start sickbeard again. My change overrides the faststop function replacing it with the normal stop function.

I also cleaned up some miscellaneous things that weren't ideal like command_args and procname being used which are discouraged in [FreeBSD documentation](https://www.freebsd.org/doc/en/articles/rc-scripting/).

### DIFF
--- a/init.freebsd
+++ b/init.freebsd
@@ -29,8 +29,8 @@ load_rc_config ${name}
 : ${sickbeard_dir=/usr/local/sickbeard}
 
 pidfile=${sickbeard_dir}/sickbeard.pid
-procname="*python"
 command=${sickbeard_dir}/SickBeard.py
+command_interpreter="/usr/local/bin/python2.7"
 sickbeard_flags="--datadir=${sickbeard_dir} --daemon --pidfile=${pidfile} ${sickbeard_flags}"
 faststop_cmd="${name}_stop"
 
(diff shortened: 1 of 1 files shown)


tehgooch:
Argh! I didn't do enough testing. Looks like this doesn't actually fix it during a shutdown/restart. I only tested `service sickbeard faststop`. I just actually tested a reboot and it did not start. Back to the drawing board. The other commits are probably good ideas though.

tehgooch:
So it turns out if you put the pid file in /var/run it will get cleaned up on shutdown, restart or faststop. Changing the pidfile directory to /var/run/sickbeard works as long as it has proper permissions. I might just make a port for sickbeard then this whole pull request is moot.

PR #987, opened 2017-11-21
Best Fit Quality

ChanceLyon:
After finding several episodes in the logs that appeared to be valid search results (to me) but did not match any of the quality filters I added my own "Best Fit Quality" setting to the general search settings.  It is a fallback if a perfect quality match cannot be found.  The setting defaults to Off, but in practice has improved my search results and I'd prefer finding a "most likely correct" quality to an Unknown and therefore usually discarded quality.

I believe this is an improvement compared to choosing the "Any" quality setting for the show.  I would rather categorize filenames that are pretty close to ideal and download those rather than settle for SDTV or lower quality.

Also: Relatively new to Git and not sure how to undo or ignore the random whitespace changes that appeared.  Git doesn't allow committing whitespace-only changes.  Thanks Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,4 +41,3 @@ ehthumbs.db
 Thumbs.db
 .directory
 *~
-*.nzb
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ ehthumbs.db
 Thumbs.db
 .directory
 *~
+*.nzb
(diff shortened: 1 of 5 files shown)


BrentOnRails:
Your whitespace changes are correct, if you want to view the Pull Request without seeing the whitespace changes (fairly common use case when doing reviews on github), you can add ` ?w=1 ` to the end of the pull request url view.

https://github.com/midgetspy/Sick-Beard/pull/987/files?w=1

source: https://github.com/blog/967-github-secrets

JackDandy:
@BrentOnRails 
This project is no longer active, perhaps try [SickGear](https://github.com/SickGear/SickGear) - you can drop in your existing sickbeard.db and continue with more features and search performance improvements.

PR #990, opened 2021-03-02
Switch Cheetah Version check to allow Cheetah 3

thisisthetechie:
Enable the use of Python-Cheetah v3


JackDandy:
@thisisthetechie This project is no longer active, perhaps try [SickGear](https://github.com/SickGear/SickGear) - you can drop in your existing sickbeard.db and continue with more features and search performance improvements. Cheetah3 was added there years ago ;)

